PR #203, opened 2026-03-13
feat(operations): fillet radius validation against analytic face curvature (#24)

andymai:
## Summary
- Adds Phase 2c validation in `fillet_rolling_ball`: rejects fillet requests where the radius meets or exceeds the minimum principal curvature radius of an adjacent analytic face
- Covers **cylinder** (R), **sphere** (R), **torus** (min(r, R−r) — accounts for fat-torus inner equator), **cone** (v_min × cos α / sin α = v_min × cot α, where α is half-angle from radial plane and v_min is slant distance to closest edge endpoint)
- Planar and NURBS faces are skipped — planar has infinite curvature radius; NURBS curvature estimation is deferred
- Returns a clear `InvalidInput` error before any fillet geometry is built

## Why
Without this check, requesting a fillet radius ≥ the cylinder/sphere radius causes the rolling-ball offset surface to degenerate (cylinder offset at R collapses to a line; sphere offset at R collapses to a point), producing garbage geometry or a panic downstream.

## Test Plan
- [ ] New test `fillet_radius_exceeds_cylinder_curvature_rejected`: cylinder r=1.0, fillet radius 1.0 and 1.5 → rejected; 0.3 → curvature check passes
- [ ] All existing fillet tests still pass (584/584)

### DIFF
--- a/crates/math/src/nurbs/fitting.rs
+++ b/crates/math/src/nurbs/fitting.rs
@@ -621,9 +621,10 @@ fn compute_lspia_step_size_weighted(
 /// # Errors
 ///
 /// Returns [`MathError::EmptyInput`] if fewer than 2 points.
-/// If the iteration has not converged after `max_iterations`, a warning is
-/// logged and the best-effort curve (lowest observed error) is returned as
-/// `Ok`; no error is raised.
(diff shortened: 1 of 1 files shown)


andymai:
@greptile-apps re-review

PR #543, opened 2026-05-07
feat(heal): wire up convert_to_bspline analytic→NURBS conversion

andymai:
## Summary

OCCT \`BRepBuilderAPI_NurbsConvert\` / \`ShapeCustom::ConvertToBSpline\` parity. The heal pipeline already registered \`convert_to_bspline\` but the underlying function silently no-op'd on every analytic face/edge. This PR makes it actually convert.

- **Surfaces** — Plane, Cylinder, Cone, Sphere, Torus → NURBS via \`brepkit-geometry::convert::surface_to_nurbs\`
- **Curves** — Line, Circle, Ellipse → NurbsCurve via \`brepkit-geometry::convert::curve_to_nurbs\`
- **Cylinder/cone v-range** — derived from face wire vertex projections onto the surface axis
- **Cone apex** — clamped strictly positive to keep the rational NURBS construction stable
- **Arcs** — start/end vertex param via \`Circle3D::project\`/\`Ellipse3D::project\`, with seam-wrap handling
- **Pcurves** — dropped on converted faces (the \`(u,v)\` numerical values do not carry over from a Cylinder's angular \`u\` to the rational-quadratic NURBS \`u\`; re-projection is deferred)

## Surface area

- \`brepkit-heal\`: rewrote \`custom::convert_to_bspline::convert_solid_to_bspline\`; added \`HealError::Geometry(GeomError)\` for clean error propagation
- \`brepkit-operations\`: added \`heal::convert_to_bspline\` thin wrapper
- \`brepkit-wasm\`: added \`convertToBspline\` JS-named binding + \`executeBatch\` dispatch returning \`{solid, converted}\`

## Test plan

- [x] 10 heal-layer unit tests cover every surface variant (Plane, Cylinder with axial range, Cone with apex clamp, Sphere, Torus) and every curve variant (Line endpoints, closed Circle, arc range wrap)
- [x] 3 WASM contract tests via \`executeBatch\`: happy-path returns \`{solid, converted >= 5}\` for cylinder, invalid handle errors, idempotent second call returns \`converted == 0\`
- [x] \`cargo test --workspace\` — all green
- [x] \`./scripts/check-boundaries.sh\` — heal still respects layer rules

## Follow-ups (out of scope)

- Re-project pcurves through the new NURBS UV space instead of dropping them (would close the OCCT-parity gap fully)
- Delete unused \`crates/heal/src/construct/convert_surface.rs\` plane stub (geometry-layer code subsumes it; this PR doesn't touch it to keep the diff focused)

### DIFF
--- a/crates/heal/src/custom/convert_to_bspline.rs
+++ b/crates/heal/src/custom/convert_to_bspline.rs
@@ -25,6 +25,7 @@ use brepkit_geometry::convert::surface_to_nurbs::{
     cone_to_nurbs, cylinder_to_nurbs, sphere_to_nurbs, torus_to_nurbs,
 };
 use brepkit_math::nurbs::surface::NurbsSurface;
+use brepkit_math::tolerance::Tolerance;
 use brepkit_math::vec::{Point3, Vec3};
 use brepkit_topology::Topology;
 use brepkit_topology::edge::{EdgeCurve, EdgeId};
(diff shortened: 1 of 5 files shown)


andymai:
Both Greptile findings addressed in 52c23cd:

1. **`arc_param_range` zero-span**: now returns `Option<(f64, f64)>` and rejects `|t_end - t_start| < tol_ang`. A distinct-vertex edge whose endpoints project to the same angle no longer silently inflates to a full circle — the caller skips on `None` (matching the Line degenerate-edge behavior). New test: `arc_param_range_rejects_zero_span`.

2. **Degenerate-line threshold**: replaced the magic `1e-15` with `Tolerance::new().linear` (1e-7), matching the topology tolerance. Below-tolerance edges now skip cleanly instead of being passed through to `line_to_nurbs` and triggering a `GeomError` that aborts the whole solid conversion. New test: `near_degenerate_line_edge_is_skipped_not_errored`.

Both tests pass and clippy/workspace tests stay green.

PR #545, opened 2026-05-08
feat(heal): merge co-circular and co-elliptical arcs in unify_same_domain

andymai:
## Summary

Generalizes \`merge_collinear_edges\` so that Circle and Ellipse arc runs are merged after face unification — not just Line runs.

OCCT's \`ShapeUpgrade_UnifySameDomain\` does this; brepkit's path was Line-only, leaving fragmented arcs in every wire after a boolean cut on a cylinder, cone, sphere, or torus.

### Behavior

Adjacent edges are merged when they:
- share a topology vertex,
- are both forward-oriented in the wire, and
- share the same underlying analytic curve.

Per-kind equivalence:
- **Line**: same direction (existing behavior)
- **Circle**: same center, normal, radius (within linear/angular tolerance)
- **Ellipse**: same center, both semi-radii, normal, and angular reference axis

\`NurbsCurve\` edges remain unmerged — they have no cheap closed-form equivalence test and merging them requires knot-vector stitching that belongs in a separate pass.

### Why orientation matters

Mixed-orientation runs (one forward, one reversed) are skipped. Reversing an arc on a \`Circle3D\` is not a textual merge — it requires flipping the surface normal, which is a substantive geometry change. Refusing the merge keeps the existing direction semantics intact.

## Test plan

8 new unit tests covering:
- 3 quarter-arcs reconstruct one closed circle (2 merges, end with 1 edge whose start==end)
- arcs on distinct circles don't merge
- arcs on anti-parallel normals don't merge
- reversed-orientation pair doesn't merge
- ellipse arcs on the same ellipse merge
- \`NurbsCurve\` runs are never touched
- mixed Line and Circle runs in one wire merge independently within each kind
- Line-only regression (Line+Line+Line still merges into one Line)

\`cargo test --workspace\` — 0 failures across all crates. \`cargo clippy --all-targets -- -D warnings\` — clean. Layer boundaries unaffected (no new deps).

## Why this is in scope

Pure refactor of an existing healing utility — no new public API, no new crate deps. The function is invoked from the boolean pipeline whenever \`unify_faces: true\` (the default), so this immediately tightens topology cleanliness for every curved-surface cut downstream.

### DIFF
--- a/crates/heal/src/upgrade/unify_same_domain.rs
+++ b/crates/heal/src/upgrade/unify_same_domain.rs
@@ -421,21 +421,27 @@ fn merge_collinear_edges(
     let mut i = 0;
 
     while i < n {
-        // Run starts at edge i. Only mergeable kinds (and forward-oriented
-        // runs) are eligible — see continuation conditions below.
+        // Capture the curve template the merged edge would inherit if the run
+        // extends. `None` ⇒ this edge cannot start a mergeable run; we never
+        // re-derive the curve at merge time, so the merge path has no
+        // unreachable arm.
         let head = &data[i];
-        let head_eligible = matches!(
-            head.kind,
-            EdgeKind::Line | EdgeKind::Circle(_) | EdgeKind::Ellipse(_)
-        ) && head.oe.is_forward();
+        let merge_template: Option<EdgeCurve> = if head.oe.is_forward() {
+            match &head.kind {
+                EdgeKind::Line => Some(EdgeCurve::Line),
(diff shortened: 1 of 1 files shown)


andymai:
Greptile finding addressed in 7204fc8: removed the `unreachable!()` by capturing the would-be merged curve at run-detection time as `Option<EdgeCurve>`. `None` now short-circuits before the merge branch is reachable, so the dead `Other` arm disappears entirely instead of being papered over with a panic.

8 unit tests still pass; clippy clean.

PR #547, opened 2026-05-08
feat(blend): analytic plane-cylinder fillet → exact torus surface

andymai:
## Summary

\`plane_cylinder_fillet\` was a \`None\`-returning stub. Every plane-meets-cylinder fillet edge fell through to the walking engine and produced a NURBS approximation. OCCT's \`BRepFilletAPI_MakeFillet\` delivers an exact toroidal blend for the typical "post on plate" geometry; this PR wires up the same fast path.

## Geometry (perpendicular convex case)

When cylinder axis ∥ plane normal:
- Rolling-ball center traces a circle of radius \`r_c + r\` at height \`r\` above the spine plane.
- Blend surface is a torus: axis = cylinder axis, major = \`r_c + r\`, minor = \`r\`.
- Active tube quadrant \`v ∈ [π, 3π/2]\` covers the lower-inner section that touches the plane below and the cylinder lateral inward.

Returns \`None\` (walker fallback) for cases the analytic path doesn't yet cover:
- oblique plane (axis not parallel to plane normal),
- concave "hole through plate" geometry (cylinder face reversed),
- \`r >= r_c\` (would invert the major radius).

Both \`(Plane, Cylinder)\` and \`(Cylinder, Plane)\` orderings are wired through \`try_analytic_fillet\`; the second swaps the resulting Stripe's face/pcurve/contact assignments.

## Subtleties handled

- \`Spine::from_single_edge\` measures **chord** length, which is zero for closed-circle spines (start vertex == end vertex). Detects closure by inspecting the spine's first edge directly.
- \`Spine::is_closed()\` returns false for single-edge spines even when that edge is itself closed — required manual detection.
- The plane's pcurve must be \`Curve2D::Circle\` (not \`Line2D\`) for the closed-spine case — \`Line2D::new(zero_vec)\` returns \`ZeroVector\` and aborts.

## Helper inlined locally

\`circle_arc_to_nurbs\` is inlined in \`analytic.rs\` to keep \`brepkit-blend\` from picking up a \`brepkit-geometry\` dependency (would violate L2 layer boundaries). Consolidating with \`geometry::convert::curve_to_nurbs::circle_to_nurbs\` and pushing it to the math layer is a follow-up.

## Test plan

- New \`fillet_cylinder_base_circle_produces_torus\` integration test: builds a primitive cylinder (radius 2, height 4), fillets the rim with \`r=0.3\`, asserts the new face is \`FaceSurface::Torus\` with \`minor = 0.3\` and \`major = 2.3\`.
- \`cargo test --workspace\` — 0 failures.
- \`cargo clippy --all-targets -- -D warnings\` — clean.
- Layer boundaries unchanged.

## Follow-ups (out of scope)

- Move circle-arc → NURBS into the math layer so \`heal\`, \`blend\`, and \`geometry\` share one impl.
- Concave (hole through plate) plane-cylinder fillet: detect via \`face_cyl.is_reversed()\`, build torus with \`major = r_c - r\` and tube quadrant \`v ∈ [-π/2, 0]\`.
- Oblique plane-cylinder: spine becomes an ellipse, blend surface no longer a simple torus.
- \`plane_cone_fillet\` (similar shape).

### DIFF
--- a/crates/blend/src/analytic.rs
+++ b/crates/blend/src/analytic.rs
@@ -83,9 +83,30 @@ pub fn try_analytic_fillet(
             }
             Ok(result)
         }
-        // Plane-Cylinder/Cylinder-Plane is handled above; the catch-all
-        // covers everything else (no analytic path yet → walker fallback).
-        _ => Ok(None),
+        // Pairs without an analytic path → walker fallback. Enumerated
+        // exhaustively (matching `try_analytic_chamfer`) so adding a new
+        // `FaceSurface` variant produces a compile error at this site
+        // rather than silently routing through the walker.
+        (
+            FaceSurface::Plane { .. }
+            | FaceSurface::Cylinder(_)
+            | FaceSurface::Cone(_)
+            | FaceSurface::Sphere(_)
+            | FaceSurface::Torus(_)
+            | FaceSurface::Nurbs(_),
+            FaceSurface::Cone(_)
+            | FaceSurface::Sphere(_)
+            | FaceSurface::Torus(_)
+            | FaceSurface::Nurbs(_),
+        )
+        | (FaceSurface::Cylinder(_), FaceSurface::Cylinder(_))
+        | (
+            FaceSurface::Cone(_)
+            | FaceSurface::Sphere(_)
+            | FaceSurface::Torus(_)
+            | FaceSurface::Nurbs(_),
+            FaceSurface::Plane { .. } | FaceSurface::Cylinder(_),
+        ) => Ok(None),
     }
 }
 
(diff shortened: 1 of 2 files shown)


andymai:
Both Greptile findings addressed in 1be9287:

1. **Wildcard catch-all → exhaustive enumeration**: \`try_analytic_fillet\` now matches every \`FaceSurface\` pair explicitly (matching \`try_analytic_chamfer\`'s style). Adding a new variant will produce a compile error at this site instead of silently routing to the walker.

2. **\`uv1\` carries plane UV, not cylinder UV**: factored a single \`PlaneAdapter\` that's shared between the \`pcurve_plane\` construction and the cross-section \`uv1\` assignment. \`section.uv1\` is now \`PlaneAdapter::project_point(p_plane)\` — the actual plane UV at the contact, consistent with the pcurve.

Workspace tests pass; clippy clean.

PR #549, opened 2026-05-08
feat(operations): concentric-sphere boolean shortcut

andymai:
## Summary

Adds a Fuse/Intersect shortcut for **concentric spheres** to the boolean dispatcher, mirroring the existing \`coaxial_cylinder_shortcut\`, \`coaxial_cone_shortcut\`, and \`box_pair_shortcut\` patterns landed in PR #541.

## Behavior

When both A and B classify as \`AnalyticClassifier::Sphere\` and their centers coincide within linear tolerance:

| Op | Result |
|----|--------|
| **Fuse** | radius = \`max(r_a, r_b)\` |
| **Intersect** | radius = \`min(r_a, r_b)\` (rejected if degenerate) |
| **Cut** | falls through to GFA (a hollow ball needs outer + inner shells, beyond what \`make_sphere\` produces) |

The shortcut returns a fresh primitive at the shared center, so chained ops see canonical 32-face sphere topology rather than residual GFA splits.

## Why this matters

Sphere-sphere booleans currently go through the same SD-detection pipeline as cylinder/cone, which has documented coplanar / coincident-pole gaps. The shortcut sidesteps those entirely for the most common concentric case.

## Test plan

8 concentric-sphere tests pass (5 existing + 3 new):

- \`concentric_spheres_different_radii_intersect_collapses_to_inner\` — verifies the new Intersect path reduces to the inner sphere.
- \`concentric_spheres_at_offset_center_fuse\` — non-origin shared center; confirms the translation step works.
- \`non_concentric_spheres_fuse_does_not_use_shortcut\` — guardrail that verifies off-center pairs still route through GFA and produce the lens-shaped union (volume between 1.2× and 2.0× a single sphere).

## Follow-ups (out of scope)

- Cut on concentric spheres: implement directly as a hollow-ball builder (outer shell at \`r_outer\`, inner shell at \`r_inner\`, both spherical) when \`r_outer > r_inner\`. Currently delegates to GFA, which may not handle this cleanly either.
- Coaxial-torus shortcut: same axis + same major radius collapses to a single torus by minor-radius algebra. Same family as this PR.

### DIFF
--- a/crates/operations/src/boolean/mod.rs
+++ b/crates/operations/src/boolean/mod.rs
@@ -327,7 +327,7 @@ pub fn boolean(
             let coincident = (*ca_center - *cb_center).length() < tol.linear;
             if coincident {
                 if let Some(result) =
-                    concentric_sphere_shortcut(topo, op, *ca_center, *ra, *rb, tol)?
+                    concentric_sphere_shortcut(topo, op, a, b, *ca_center, *ra, *rb, tol)?
                 {
                     return Ok(result);
                 }
@@ -806,7 +806,6 @@ fn coaxial_cone_shortcut(
     Ok(Some(cone))
 }
 
-/// Build the world-frame transform that maps a primitive built in the
 /// Compute the concentric-sphere boolean for two spheres sharing a
 /// center. Returns `Ok(None)` when the shortcut doesn't apply (Cut, or
 /// degenerate radii).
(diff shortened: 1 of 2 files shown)


andymai:
All 4 Greptile findings addressed in 283b43c:

1. **Doc comment corruption** — restored each function's docstring to its own block. \`concentric_sphere_shortcut\` and \`xform_from_canonical_z\` now have their proper, separate docs.

2. **(companion fix to #1)** — \`xform_from_canonical_z\` has its first sentence back.

3. **Dead Intersect guard** — removed. Replaced with a comment noting that the top-of-function guard already ensures positivity.

4. **Hardcoded segment count** — now inherits from inputs. Reads vertex count from each solid (\`make_sphere(r, n)\` allocates exactly \`n\` equatorial vertices, no pole vertices) and uses \`max(segments_a, segments_b)\`. A 64-segment input no longer silently downgrades to 32.

Tests pass; clippy clean.

PR #550, opened 2026-05-08
feat(blend): analytic plane-cone fillet → exact torus surface

andymai:
## Summary

\`plane_cone_fillet\` was a \`None\`-returning stub. Every plane-meets-cone fillet edge fell through to the walking engine. This PR replaces it with a closed-form toroidal blend for the convex "regular frustum bottom rim" case — the natural sibling of the plane-cylinder analytic fillet (PR #547) and completes the analytic blend trio for plane + axis-aligned-quadric pairs.

## Geometry

At the spine (cone-plate intersection circle of radius \`r_p\`), the dihedral between outward surface normals is \`π - α\` (α = cone half-angle), so the fillet wedge half-angle is \`α/2\` and the ball center sits at distance \`r/sin(α/2)\` along the outward bisector \`cos(α/2)·radial - sin(α/2)·n_p_inward\`. Ball center lands at radial \`r_p + r·cot(α/2)\` and axial offset \`-r\` from the plate.

## Resulting torus

| Parameter | Value |
|----|---|
| axis | \`-axis_c\` (= \`+n_p_inward\`) |
| center | cone-axis ∩ plate, offset by \`-r·n_p_inward\` |
| major | \`r_p + r·cot(α/2)\` |
| minor | \`r\` |
| active tube | \`v ∈ [π/2, 3π/2 - α]\`, width \`π - α\` |

At \`α = π/2\` (cone → cylinder limit), \`cot(π/4) = 1\` so major reduces to \`r_p + r\` and the active range becomes \`[π/2, π]\` — exactly the plane-cylinder result.

## Returns \`None\` for

- cone axis not anti-parallel to the inward plane normal,
- cone face reversed (concave / "tapered hole through plate"),
- half-angle α too close to 0 or π/2 (degenerate),
- apex on the plate or on the wrong side,
- spine too short.

## Test plan

This PR also **hardens the existing plane-cylinder test** with strict geometric point-equality assertions — caught a torus-axis-direction subtlety the previous radii-only test would have missed.

- \`fillet_cylinder_base_circle_produces_torus\` (hardened): checks point-equality at v=π/2 (plate) and inner-equator v (cylinder) within 1e-6.
- \`fillet_cone_bottom_rim_produces_torus\` (new): verifies major matches \`r_p + r·cot(α/2)\`, minor matches the fillet radius, and point-equality at v=3π/2 (plate) and v=atan2(cos α, -sin α) (cone).

Strict point-equality (rather than dense sampling with a min-distance check) avoids hiding errors behind discretization noise — sample-min could be off by up to \`r · (2π / sample_count)\` and still pass.

## Follow-ups (out of scope)

- **Inverted-frustum cone fillet** (cone axis parallel rather than anti-parallel to inward plate normal): the major-radius formula becomes \`r_p + r·tan(α/2)\` with a different active tube quadrant. Currently falls back to walker.
- **Concave cone fillet** (cone face reversed = tapered hole through plate): needs a different torus quadrant.
- **Top-rim fillet of a frustum**: bisector points outward + UP rather than outward + DOWN; major formula is \`r_p + r·cot(α/2)\` still but center placement differs.
- **\`cylinder_cylinder_fillet\`** is still a stub.

### DIFF
--- a/crates/blend/src/analytic.rs
+++ b/crates/blend/src/analytic.rs
@@ -731,16 +731,21 @@ fn cyl_v_at_point(cyl: &brepkit_math::surfaces::CylindricalSurface, p: Point3) -
 /// empty-wedge direction).
 ///
 /// The fillet surface is a torus:
-///   - axis = `-axis_c` (= `+n_p_inward`), so `v=π/2` lands on the plate;
+///   - axis = `-n_p_inward` (= `+axis_c` for the regular-frustum case
+///     where `axis_c · n_p_inward = -1`); with this convention `sin v`
+///     points away from the plate, so `sin v = -1` is what pulls the
+///     tube point back toward the plate. Plate contact lands at
+///     `v = 3π/2`; cone contact at `v = atan2(cos α, -sin α)`.
 ///   - center at the cone-axis projection onto the plate, offset by
 ///     `-r·n_p_inward`;
 ///   - major radius `r_p + r·cot(α/2)`,
 ///   - minor radius `r`;
-///   - active tube parameter `v ∈ [π/2, 3π/2 - α]`, width `π - α`.
+///   - active tube parameter `v ∈ [atan2(cos α, -sin α), 3π/2]`,
+///     width `π - α`.
 ///
 /// At α = π/2 (degenerate "cone" approaching a cylinder), `cot(π/4) = 1`
-/// so major reduces to `r_p + r` and the active range becomes `[π/2, π]` —
-/// exactly the plane-cylinder result.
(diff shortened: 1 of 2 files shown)


andymai:
Greptile finding addressed in 0c62997: the docstring and inline comment now correctly describe \`axis_dir = -n_p_inward (= +axis_c)\` with plate contact at \`v = 3π/2\` and cone contact at \`v = atan2(cos α, -sin α)\` — matching what the code does and what the strict point-equality tests verify. Cylinder-limit sentence now reads \`[π, 3π/2]\` instead of \`[π/2, π]\`.

Code unchanged; tests still pass.

PR #552, opened 2026-05-08
feat(blend): analytic plane-cylinder chamfer → exact cone surface

andymai:
## Summary

\`try_analytic_chamfer\` previously routed every Plane+Cylinder pair to the walker. This PR adds a closed-form **conical** chamfer for the convex bottom-rim case — the natural sibling of the analytic plane-cylinder fillet (#547) and analytic plane-cone fillet (#550), and a direct OCCT \`BRepFilletAPI_MakeChamfer\` parity feature.

## Geometry

For chamfer distances \`d1\` (on the plate, radially inward from the spine) and \`d2\` (on the cylinder lateral, axially into the material):

| Parameter | Value |
|---|---|
| chamfer surface | frustum of a cone |
| axis | cylinder axis (cone opens toward the plate) |
| half-angle | \`α = atan2(d1, d2)\` (45° for symmetric \`d1 = d2\`) |
| apex | on cylinder axis, axial offset \`(r_c − d1)·d2/d1\` into the cylinder material |
| contact 1 (plate) | circle at radial \`r_c − d1\` |
| contact 2 (cylinder) | circle at radial \`r_c\`, axial \`+d2\` |

## Returns \`None\` for

- cylinder axis not parallel to the plane normal,
- cylinder face reversed (concave / hole geometry),
- non-positive \`d1\` or \`d2\`,
- \`d1 >= r_c\` (would pass through the cylinder axis),
- spine too short.

Both \`(Plane, Cylinder)\` and \`(Cylinder, Plane)\` orderings are wired up; the second swaps \`d1↔d2\` (so the helper always sees them in canonical "plate-distance, cylinder-distance" order) and swaps the result's face/pcurve/contact assignments via the shared \`swap_stripe_sides\` helper.

## Test plan

\`chamfer_cylinder_base_circle_produces_cone\` — strict point-equality checks at the predicted plate and cylinder contact \`(u, v)\` parameters, using the cone parameterization to verify both contacts lie on the same \`ConicalSurface\`. Asserts half-angle equals π/4 for symmetric chamfers.

## Follow-ups (out of scope)

- Plane-cone chamfer (frustum bottom/top rim).
- Concave plane-cylinder chamfer (hole through plate).

### DIFF
--- a/crates/blend/src/analytic.rs
+++ b/crates/blend/src/analytic.rs
@@ -733,14 +733,15 @@ fn cyl_v_at_point(cyl: &brepkit_math::surfaces::CylindricalSurface, p: Point3) -
 /// # Geometry
 ///
 /// The chamfer surface is a frustum of a cone:
-///   - axis = cylinder axis,
+///   - axis = cylinder axis (the cone's `+axis_c` direction points toward
+///     the cylinder material; `v` grows from apex into the material);
 ///   - half-angle `α = atan2(d1, d2)` (45° for symmetric `d1 = d2`),
-///   - apex at the cylinder axis, axial offset `(r_c - d1)·d2/d1` *into*
-///     the cylinder material (so the cone "grows" as `v` increases out
-///     toward the spine and beyond).
+///   - apex on the cylinder axis, axial offset `(r_c - d1)·d2/d1` away
+///     from the cylinder material (in the empty-wedge half-space — i.e.
+///     opposite to the cylinder body relative to the plate);
 ///   - contact 1 on the plate: circle at radial `r_c - d1`, on the plate;
 ///   - contact 2 on the cylinder lateral: circle at radial `r_c`, axially
-///     offset `d2` into the material.
+///     offset `+d2` into the material.
 ///
 /// Both contacts are circles around the cylinder axis. The chamfer face
 /// connects them with a flat cone (ruled surface).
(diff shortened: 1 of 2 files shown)


andymai:
Greptile finding addressed in f52916a. The contact-circle placement bug was real: \`cyl_contact_center\` was at \`z = -d2\` instead of \`z = +d2\`. The cone surface was right by accidental double-negation, but the separately-built contact circle would have mis-trimmed downstream.

**Fix**: replaced the misnamed \`axial_into_material\` (which actually pointed AWAY from the cylinder material in the chamfer dispatcher's no-orient convention) with two unambiguous directions:
- \`axis_toward_apex = +n_p_inward\` (where the cone's virtual apex lives);
- \`axis_toward_material = -n_p_inward\` (where the cylinder body and contact circle live).

\`apex_pos\` uses \`axis_toward_apex\`; \`cone_axis = axis_toward_material\` so \`v\` grows from apex into the material; \`cyl_contact_center\` uses \`axis_toward_material * d2\`, landing at the correct \`z = +d2\`.

**Test hardened** with two new assertions that would have caught the original bug:
- Strict \`p_cyl == (0, r_c, +d)\` at \`v = r_c / cos α\` (the cone surface contact);
- No \`Circle3D\` edge on the chamfer cone face at \`z = -d\` (the misplaced contact circle).

10/10 blend integration tests pass; clippy clean.

PR #556, opened 2026-05-08
feat(operations): coaxial-torus boolean shortcut

andymai:
## Summary

Closes the boolean-shortcut family for axis-aligned analytic primitives: coaxial cylinders (#541), coaxial cones (#541), box pairs (#541), concentric spheres (#549), and now **coaxial tori**. When both A and B classify as Torus with shared center, axis (parallel or antiparallel), and major radius, Fuse and Intersect collapse to a single torus via minor-radius algebra.

| Op | Result |
|----|--------|
| **Fuse** | minor = \`max(min_a, min_b)\` |
| **Intersect** | minor = \`min(min_a, min_b)\` |
| **Cut** | falls through to GFA (hollow torus needs outer + inner shells) |

## Detection details

- **Center**: \`|c_a − c_b| < tol.linear\`
- **Axis**: \`|axis_a · axis_b| > 1 − tol.angular\` — accepts both parallel and antiparallel axes since a torus is symmetric about its central plane (rotating its axis 180° gives the same surface).
- **Major radius**: \`|maj_a − maj_b| < tol.linear\`
- **Minor sanity**: result minor must satisfy \`make_torus\`'s \`minor < major\` self-intersection check.

## Test plan

4 new tests on top of the existing 5 in \`coaxial_torus.rs\`:
- \`coaxial_torus_fuse_collapses_to_larger_minor\`
- \`coaxial_torus_intersect_collapses_to_smaller_minor\`
- \`coaxial_torus_at_offset_center_fuse\` — non-origin shared center
- \`non_coaxial_torus_fuse_does_not_use_shortcut\` — guardrail confirming different-major pairs still route through GFA

## Follow-ups (out of scope)

- Cut on coaxial tori (hollow-torus topology builder).
- Coaxial-torus with mismatched majors but compatible cross-section (more complex SD case).

### DIFF
--- a/crates/operations/src/boolean/mod.rs
+++ b/crates/operations/src/boolean/mod.rs
@@ -333,6 +333,42 @@ pub fn boolean(
                 }
             }
         }
+
+        // Coaxial-torus merge shortcut: when both A and B classify as Torus
+        // with the same center, axis (parallel/antiparallel), and major
+        // radius, Fuse and Intersect collapse to a single torus by minor
+        // radius algebra. Same family as the concentric-sphere shortcut
+        // above; sidesteps GFA's torus same-domain handling for the
+        // common shared-major case.
+        if let (
+            Some(brepkit_algo::classifier::AnalyticClassifier::Torus {
+                center: ca_center,
+                axis: aa,
+                major_radius: maj_a,
+                minor_radius: min_a,
+            }),
+            Some(brepkit_algo::classifier::AnalyticClassifier::Torus {
+                center: cb_center,
+                axis: ab,
+                major_radius: maj_b,
+                minor_radius: min_b,
+            }),
+        ) = (ca.as_ref(), cb.as_ref())
+        {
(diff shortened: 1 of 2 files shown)


andymai:
Both Greptile findings fixed in d31bdc9:

1. **Pre-existing test silently using the shortcut** — \`toruses_different_minor_radius_intersect_nontrivial\` used matching majors (the exact signature the new shortcut hijacks). Renamed to \`toruses_different_major_radius_intersect_nontrivial\` and switched the inputs (3.0 vs 4.0 major) so the shortcut's \`same_major\` guard rejects the pair and the GFA path stays under test.

2. **Misleading test name** — \`non_coaxial_torus_fuse_does_not_use_shortcut\` renamed to \`different_major_torus_fuse_does_not_use_shortcut\` since the two tori it builds DO share center and axis; only the major radii differ.

All 9 tests still pass.

PR #681, opened 2026-05-19
fix(boolean): correct intersect/cut for containment + classify empty results clearly

andymai:
## Summary

Closes the remaining brepjs parity gap for the `cut-then-fuse-back` invariant on contained solids (the counterexample `[0.5, 0.9531210881257514, 1.7801e-308]` from fast-check, which decodes to A=unitCube(0.5,0.5,0.5), B=unitCube(0.95,0.95,0.95) at the origin → A ⊂ B).

Two related changes:

1. **`OperationsError::EmptyResult { reason }`** — a distinct variant for "operation produced no geometry," so callers can apply identity rules (`empty ∪ X = X`, `empty ∩ X = empty`, `X - empty = X`) without string-matching the error message. The 3 existing "empty result" `InvalidInput` returns are migrated.

2. **Containment-Cut shortcut** in `boolean/mod.rs`: when the target is fully inside the tool (`a_in_b && !b_in_a`), return `Err(EmptyResult)` instead of deferring to GFA. Previously GFA fabricated a degenerate vol=0 "box" (6 faces / 8 verts / 13 edges); the brepjs adapter couldn't recognize this as empty and the `vol((A - B) ∪ (A ∩ B)) = vol(A)` invariant collapsed to vol 0.

`cut_multi_region_input` is updated to catch `EmptyResult | InvalidInput` in its per-component empty-result fallback (was matching only `InvalidInput`), preserving the `compound_cut_shelled_target_*` behavior I caught regressing on the first attempt.

## Investigation notes (what changed vs. what was already fine)

| Scenario | Behavior on main | Cause | This PR |
|---|---|---|---|
| `Cut(A ⊂ B, B)` | `Ok(degenerate vol=0 solid, 13 edges)` | GFA can't represent empty; fabricates degenerate result | `Err(EmptyResult)` |
| `Intersect(A ⊂ B, B)` | `Ok(copy of A, vol=0.125)` | Existing containment shortcut at `mod.rs:166` is correct | unchanged |
| Inclusion-exclusion on near-coincident faces (your scenario 2) | invariant holds (rel_error 3.3e-8) | GFA handles near-coincident faces correctly | unchanged (test added to lock it in) |
| `Cut(A, B)` where AABBs differ by exactly `tol.linear` (your scenario 1) | `Err(InvalidInput "Cut of identical solids…")` | Existing identical-shortcut at `mod.rs:157` fires | now `Err(EmptyResult)` |

## Test plan

- [x] `cargo test -p brepkit-operations` — 596 lib + ~22 integration files, all pass. One regression caught and fixed during dev: `compound_cut_shelled_target_9_tools` needed the `EmptyResult | InvalidInput` widening in `cut_multi_region_input`.
- [x] `cargo build --workspace`, `cargo clippy -p brepkit-operations --all-targets -- -D warnings` — clean.
- [x] 5 new reproducer tests in `boolean_invariants.rs`, all passing:
  - `cut_contained_solid_returns_empty_result`
  - `intersect_contained_solid_returns_contained`
  - `cut_then_fuse_back_containment` (the actual brepjs counterexample)
  - `cut_then_fuse_back_recovers_volume_near_coincident`
  - `inclusion_exclusion_near_coincident_faces`
- [ ] Brepjs adapter swap to consume `EmptyResult` instead of error-string matching (follow-up on the brepjs side after this lands and a new brepkit-wasm release ships).

## Notes for review

- The new `EmptyResult` Display string starts with `"empty result:"` — same substring brepjs's current `isEmptyBooleanError` likely matches on, so the adapter keeps working until you update it to match on the variant directly.
- Not enabling automerge per your instructions — flagging back for Greptile/decision.

### DIFF
--- a/crates/operations/src/boolean/mod.rs
+++ b/crates/operations/src/boolean/mod.rs
@@ -157,12 +157,25 @@ pub fn boolean(
         if aabbs_match && all_b_verts_in_a && all_a_verts_in_b {
             return match op {
                 BooleanOp::Fuse | BooleanOp::Intersect => Ok(crate::copy::copy_solid(topo, a)?),
-                BooleanOp::Cut => Err(crate::OperationsError::InvalidInput {
-                    reason: "Cut of identical solids produces empty result".into(),
+                BooleanOp::Cut => Err(crate::OperationsError::EmptyResult {
+                    reason: "Cut of identical solids".into(),
                 }),
             };
         }
-        // For Cut with containment, defer to GFA (produces shelled solid).
+        // Containment shortcuts:
+        // - Fuse/Intersect with either containment direction: copy the
+        //   appropriate solid.
+        // - Cut with A ⊆ B: result is empty (A is fully removed). Return
+        //   EmptyResult explicitly — without this short-circuit, GFA falls
+        //   back to producing a degenerate vol=0 solid that callers
+        //   mistake for a real result, breaking volume invariants like
+        //   `vol((A-B) ∪ (A∩B)) = vol(A)`.
+        // - Cut with B ⊂ A: defer to GFA (produces hollow solid).
+        if op == BooleanOp::Cut && a_in_b && !b_in_a {
+            return Err(crate::OperationsError::EmptyResult {
+                reason: "Cut with target fully contained in tool".into(),
+            });
+        }
         if (b_in_a || a_in_b) && op != BooleanOp::Cut {
             return match (op, b_in_a, a_in_b) {
                 (BooleanOp::Fuse, true, _) => Ok(crate::copy::copy_solid(topo, a)?),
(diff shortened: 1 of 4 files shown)


andymai:
Both Greptile P2 comments addressed in 0bdab17:

- **safe_vol** now matches `Err(EmptyResult { .. }) => 0.0` explicitly and panics on any other variant — so a future `Algo` / `NonManifoldResult` regression surfaces instead of getting silently scored as zero volume.
- **cut_then_fuse_back_containment** drops the unreachable `Some(diff)` fuse-back arm. The test asserts `Cut(A ⊂ B, B)` is `EmptyResult` upfront; the invariant reduces to `vol_inter ≈ vol_a`, stated directly.

Skipped Greptile's "widened cut_multi_region_input fallback may retain wholly-contained components" concern: that path matched `InvalidInput` before this PR with exactly the same retention semantics, and `compound_cut_shelled_target_9_tools` is the regression that pointed the widening out. Adding a dedicated "compound cut with one wholly-contained component" coverage test is a sensible follow-up but not required to land this fix.

PR #692, opened 2026-05-20
fix(algo): close compound boolean variance + lay foundation for box-sphere intersect

andymai:
## Summary

Two perf regressions surfaced when comparing against the README's benchmark table:

1. **Multi-boolean variance** (`box - 16 cylinders`) — median **3.87s → 111ms** (35× faster), max **10.5s → 114ms** (variance collapsed from 100× to ~5%). Root cause: two HashMap-iteration sites in the GFA pipeline drove different downstream code paths across runs. Follows the same pattern as #689.
2. **`intersect(box,sphere)`** — diagnosed (PR #382 bisect, root cause in `split_noseam_face_direct`) and foundation laid (`Circle3D::intersect_segment` + closed-curve FF split). End-to-end fix not yet complete — the face_splitter still can't form the spherical octant cap from 2+ intersecting arcs on a sphere hemisphere, so the pipeline continues to fall back to mesh boolean. See [Open follow-up](#open-follow-up) below.

## Commits

| SHA | Title |
|---|---|
| `f7664a4` | `fix(algo)`: convert FaceInfo HashSets to BTreeSets for deterministic iteration |
| `a25ca80` | `fix(heal)`: sort unify_faces merge_groups for deterministic processing |
| `b05d024` | `test(boolean)`: add ignored profile_intersect harness for box-sphere regression |
| `0f93f49` | `feat(algo)`: split closed Circle3D FF sections at face boundaries |

## Measured impact (native bench, `cad_operations`)

| Bench | Before (main) | After (this branch) |
|---|---|---|
| `multi-boolean model` min | 109 ms | 110 ms |
| `multi-boolean model` median | **3.87 s** | **111 ms** |
| `multi-boolean model` max | **10.5 s** | **114 ms** |
| `cut(box,cyl) x10` | 6.8 ms | 7.0 ms (within noise) |
| `intersect(box,sphere) x10` | 273 ms | 273 ms (unchanged — see below) |

## Root cause: multi-boolean variance

PR #689 fixed two HashMap-iteration sites (sd_pairs sort, vv_vertex_seed sorted Vec) but explicitly noted "at least one more HashMap iteration site still drives downstream count divergence." This branch closes two more:

- **`FaceInfo` HashSet → BTreeSet** (`crates/algo/src/ds/face_info.rs`) — `pave_blocks_in: HashSet<PaveBlockId>` was iterated at `fill_images_faces.rs:1011` to build the per-face `SectionSource` list that drives face splitting. HashSet's random iteration order produced different `SectionSource` orderings → different sub-face topologies across runs. All six FaceInfo set fields converted; BTreeSet has matching API and the per-face cardinalities are <100 so the O(log n) overhead vs O(1) is negligible.
- **`unify_faces::merge_groups` sort** (`crates/operations/src/heal.rs:1209`) — `merge_groups` was built from `groups.into_values()` of a HashMap. Downstream `canonical_vtx.entry(...).or_insert(vid)` is first-seen-wins, so HashMap iteration order chose different canonical vertices on different runs, cascading into different edge re-allocations and different face/edge counts. Sort by `(group[0])` after each group's contents are also sorted internally.

## Root cause: intersect(box,sphere) regression

Bisect (full log in commit `0f93f49`) identified PR #382 (commit `6ce3202`, "GFA switchover — delete old pipeline ~15k lines") as the regression-introducing commit. Before #382, GFA failures fell through to the analytic pipeline (~3.5ms/op). After #382 only mesh boolean is the fallback (~27ms/op, losing the analytic sphere surface — result has 215 *Plane* faces, zero *Sphere*).

Tracing GFA-direct (the new \`profile_gfa_box_sphere_direct\` harness in `0f93f49`) shows GFA produces a 3-face Euler-invalid topology because the sphere hemisphere returns unsplit:

\`\`\`
fill_images_faces: face Id(6) has_sections=true sections=3
fill_images_faces: face Id(6) split into 1 sub-faces
\`\`\`

Three section curves on the hemisphere but only 1 sub-face produced. The bug is in `split_noseam_face_direct` (`crates/algo/src/builder/face_splitter/special_cases.rs:55-60`):

\`\`\`rust
// Skip full-circle section edges (start approx end in 3D) -- only use
// the half-arcs produced by build_seam_split_sections.
if (section.start - section.end).length() < tol.linear {
    continue;
}
\`\`\`

Plane-sphere intersections produce `Circle3D` curves where the circle's parameterization starts and ends at the same 3D point. With every section skipped, `cap_edges` is empty and the function returns the face unsplit. The unsplit hemisphere is then classified by sampling a single interior point that may lie outside the box → entire hemisphere classified Outside → spherical sub-face dropped.

### What this PR adds

1. **`Circle3D::intersect_segment`** (`crates/math/src/curves.rs`) — algebraic closed-form solver for Circle3D ∩ line segment in 3D. Handles both the in-plane case (segment lies in the circle's plane → 2D line-circle solve) and the plane-crossing case (segment crosses the circle's plane at one point → check if on circle). 4 unit tests cover the typical configurations used by the box-sphere FF path.
2. **`closed_circle_boundary_crossings`** (`crates/algo/src/pave_filler/phase_ff.rs`) — finds where a closed Circle3D section crosses the *analytic* face's outer-wire boundary. When 16 crossings are reported (one per polygon vertex), the curve is coincident with the boundary itself — that case is detected and skipped to avoid splintering the face into per-polygon-edge arcs.
3. **`emit_split_circle_arcs`** (same file) — for ≥ 2 crossings, splits the closed circle into N open arcs at the crossings. Each arc becomes its own `IntersectionCurveDS` entry with a single pave block whose t_range covers that arc. Arcs whose midpoints fall outside the union of `face_a`'s and `face_b`'s AABBs (with the sphere face's AABB expanded to the sphere bounds) are dropped — those portions of the original closed curve aren't on the face pair.

### Open follow-up

The phase_ff change correctly emits open arcs — GFA now sees 5 sub-faces internally (up from 3), with the box faces properly split by section arcs. But the spherical sub-face is still not produced: when 2+ arcs cross on a sphere hemisphere (e.g. the great circles from x=0 and y=0 planes meeting at the pole), `split_noseam_face_direct` can't fold them into a single closed cap loop and returns the hemisphere unsplit. The pipeline continues to fall back to mesh boolean.

Closing that gap requires generalising the face splitter for multi-arc sphere hemispheres — see `face_splitter/mod.rs:138` dispatch and `special_cases.rs::split_noseam_face_direct`. The infrastructure landed in this PR is the foundation for that next step.

## Test plan

- [x] All 89 `brepkit-algo` tests pass (0 regressions)
- [x] All 596 `brepkit-operations` tests pass (0 regressions)
- [x] 4 new unit tests in `brepkit-math::curves::tests` for `Circle3D::intersect_segment`
- [x] Diagnostic harnesses (`#[ignore]`d): `profile_intersect_box_sphere`, `profile_gfa_box_sphere_direct`
- [x] `cargo clippy --all-targets --workspace -- -D warnings` clean
- [x] Pre-push hook (full workspace tests + cargo-deny) passes

### DIFF
--- a/crates/algo/src/ds/face_info.rs
+++ b/crates/algo/src/ds/face_info.rs
@@ -1,6 +1,6 @@
 //! Per-face state accumulated during PaveFiller.
 
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 
 use brepkit_topology::vertex::VertexId;
 
@@ -10,18 +10,25 @@ use super::pave::PaveBlockId;
 ///
 /// Populated incrementally by PaveFiller phases (VE, EF, FF).
 /// Consumed by the Builder to split faces.
+///
+/// `BTreeSet` (not `HashSet`) so iteration is deterministic by ID — downstream
+/// face splitting in `fill_images_faces` iterates `pave_blocks_in` to build
+/// `SectionSource` lists, and HashSet's random iteration order produces
+/// different split topologies across runs, which cascades into 100–500×
+/// per-iter variance in compound boolean benchmarks. See PR #689 for the
+/// other two sites that drove the same nondeterminism.
 #[derive(Debug, Clone, Default)]
 pub struct FaceInfo {
     /// Pave blocks lying ON the face boundary (original boundary edges, split).
-    pub pave_blocks_on: HashSet<PaveBlockId>,
+    pub pave_blocks_on: BTreeSet<PaveBlockId>,
     /// Pave blocks lying IN the face interior (from other faces' edges crossing).
-    pub pave_blocks_in: HashSet<PaveBlockId>,
+    pub pave_blocks_in: BTreeSet<PaveBlockId>,
     /// Section pave blocks from face-face intersection curves.
-    pub pave_blocks_sc: HashSet<PaveBlockId>,
+    pub pave_blocks_sc: BTreeSet<PaveBlockId>,
     /// Vertices on the face boundary.
-    pub vertices_on: HashSet<VertexId>,
+    pub vertices_on: BTreeSet<VertexId>,
     /// Vertices in the face interior.
-    pub vertices_in: HashSet<VertexId>,
+    pub vertices_in: BTreeSet<VertexId>,
     /// Vertices from section curves.
-    pub vertices_sc: HashSet<VertexId>,
+    pub vertices_sc: BTreeSet<VertexId>,
 }
--- a/crates/algo/src/pave_filler/phase_ff.rs
+++ b/crates/algo/src/pave_filler/phase_ff.rs
@@ -138,6 +138,38 @@ pub fn perform(
             };
 
             for raw in raw_curves {
+                // Closed Circle3D sections — produced by plane-sphere
+                // intersections — get split at face-boundary crossings so
+                // downstream face splitters see open arcs they can match
+                // up. Without this, the closed curve is dropped by
+                // `split_noseam_face_direct` and the spherical sub-face is
+                // lost entirely.
+                //
+                // Note: the split here is structurally correct (the FF
+                // arcs now have proper endpoints on the analytic face's
+                // boundary), but `split_noseam_face_direct` still can't
+                // form a single closed cap loop when 2+ arcs cross on a
+                // sphere hemisphere (e.g. the great circles from x=0 and
+                // y=0 planes meeting at the pole). The hemisphere falls
+                // back to "unsplit" → the spherical sub-face is missing
+                // from the GFA output → the boolean pipeline retries with
+                // mesh boolean. Closing that gap requires generalising
+                // the face splitter for multi-arc sphere hemispheres —
+                // see `crates/algo/src/builder/face_splitter/mod.rs:138`
+                // dispatch and `special_cases.rs::split_noseam_face_direct`.
+                if let EdgeCurve::Circle(circle) = &raw.curve {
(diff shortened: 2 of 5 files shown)


andymai:
Thanks for the thorough review! Addressed all 4 line comments in `ecd21e1`:

1. **@copilot @ `curves.rs:325`** — fixed discriminant unit mismatch. `disc` is in length⁴, was compared against `-tol` (length). Now uses scale-aware threshold `-tol² · a`.
2. **@copilot @ `phase_ff.rs:884`** — arc-span-> π would silently flip to the complementary arc downstream (`shorter_arc_range` etc.). Each surviving arc now splits into `⌈span/π⌉` sub-arcs via midpoint vertices so every emitted edge has span ≤ π.
3. **@greptile-apps @ `phase_ff.rs:758`** — switched to `f64::total_cmp` (handles NaN deterministically, no transitivity hazard).
4. **@greptile-apps @ `phase_ff.rs:893`** — restructured into two passes so crossing vertices are only allocated after the AABB filter selects which arcs survive (no orphan vertices). Midpoint vertices from #2 are created only inside the per-survivor loop after the arc has passed the filter.

Verified: 89/89 algo tests, 596/596 ops tests, 4 new math unit tests, clippy clean. Diagnostic harness still produces the same GFA output (5 sub-faces → mesh fallback) — no functional regression vs the pre-review-fix state.

Issue #696, opened 2026-05-20
tessellation: non-manifold edges in dovetail baseplate STL exports (2.89.2)


andymai:
## Summary

`brepkit-wasm` 2.89.2 produces tessellated meshes with **non-manifold edges** for dovetail-connector baseplate exports that are watertight under `brepjs-opencascade` (OCCT). PR #692's compound-boolean variance fix is in this release and behaves correctly — but the tessellation/winding pathology is independent and still present.

The downstream consumer is [gridfinity-layout-tool](https://github.com/andymai/gridfinity-layout-tool); the failing tests live in `src/features/generation/worker/generators/baseplateGenerator.scenario.dovetail.test.ts` (originally written against issue [gridfinity-layout-tool#1407](https://github.com/andymai/gridfinity-layout-tool/issues/1407)).

## Observed under 2.89.2

| Scenario                                                 | Non-manifold edges |
| -------------------------------------------------------- | -----------------: |
| 5×4 middle-column tile (3 join edges, tongue protrusion) |            **142** |
| 4×4 interior tile (4 join edges)                         |             **73** |

OCCT (default brepjs kernel) for the same params: **0**.

The non-manifold count comes from edge-shared-by-≥3-triangles in the binary STL. The diagnostic also surfaces in stderr as `[repairMeshWinding] Skipped non-manifold edges during winding repair { nonManifoldEdges: N, triangleCount: M }` from the consumer's defensive winding-repair pass (which is a no-op under OCCT — see [`src/shared/generation/repairMeshWinding.ts`](https://github.com/andymai/gridfinity-layout-tool/blob/main/src/shared/generation/repairMeshWinding.ts)).

## Minimal repro

The full repro path is `exportBaseplate(params, 'stl')` → tessellate → binary STL. The `BaseplateParams` shapes that fail:

```ts
// 5×4 middle-column — 142 non-manifold edges
{
  width: 5, depth: 4, gridUnitMm: 42,
  magnetHoles: false,
  paddingLeft: 0, paddingRight: 0, paddingFront: 0, paddingBack: 0,
  fractionalEdgeX: 'end', fractionalEdgeY: 'end',
  lightweight: true,
  connectorNubs: true,
  edges: { left: 'exterior', right: 'join', front: 'join', back: 'join' },
}

// 4×4 interior — 73 non-manifold edges
{
  width: 4, depth: 4, gridUnitMm: 42,
  magnetHoles: false,
[...]

PR #697, opened 2026-05-20
fix(tessellate): partial fix for #696 — dedupe coincident triangles + NM diagnostics

andymai:
## Summary

Partial fix for #696 (non-manifold edges in dovetail baseplate STL exports). Diagnostic helpers, regression coverage, deterministic NM-edge processing, a tessellation-level dedup pass, and an extended unify trigger that together close 82–85% of the consumer's symptom.

## Changes

**`test(tessellate): non-manifold diagnostic + dovetail regression coverage`** (`061d2a9`)
- New `non_manifold_edge_count(mesh)` helper, symmetric with `boundary_edge_count`. Both are needed to validate a closed 2-manifold; the existing helper's doc inaccurately claimed to measure non-manifold but only counted 1-sided edges.
- Three dovetail-pattern regression tests in `tessellate/tests.rs` (axis-aligned tongue, multi-tile + hollow, trapezoidal tongue). All pass currently — synthetic geometry doesn't reproduce the consumer's specific bug, but the tests catch any regression in the underlying patterns.
- Deterministic iteration in `build_manifold_shell` and `split_nonmanifold_edges`: both iterated a `HashMap` to collect non-manifold edges. Each iteration mutates topology (face removal / edge replacement), so processing order changed the assembly outcome when edges share faces. Sort by `EdgeId` first. Matches the #689/#692 pattern.

**`fix(tessellate): dedupe coincident triangles from boolean coplanar residue`** (`e96b518`)
- New `dedupe_coincident_triangles` Phase-7 pass in `tessellate_solid`. Keys triangles on quantized vertex positions (sorted) + winding parity. Cancels matching-key/opposite-winding pairs (two faces tessellated from opposite sides of the same plane). Dedupes matching-key/same-winding pairs to one.
- Quantization at **1µm** — tight enough to preserve thin-plate features (verified against the existing 1e-4mm-thick box test), coarse enough to absorb post-merge floating-point noise that boundary-vertex welding can't reach.

**`fix(boolean): also trigger unify when GFA result is non-manifold`** (`3d0214a`)
- The existing `unify_faces` invocation in the GFA success path was gated behind `Euler != 2` — but overlapping coplanar faces can cancel out of V-E+F counting, leaving 3+-face junctions with Euler == 2. Extend the gate to also fire when `is_closed_manifold(result)` is false.

**`review: address PR #697 feedback`** (`b7f8a80`)
- Tighten `is_watertight` to check both boundary and non-manifold edges (was only checking boundary, despite its 2-manifold docstring).
- Fix unit label (was "1nm", should be "1µm" — the constant is `1e-6`).
- Compact `mesh.positions`/`mesh.normals` after dedup; the previous version rebuilt indices but left phantom entries that bbox passes and non-indexed exporters would surface.
- Add four direct unit tests for `dedupe_coincident_triangles`: opposing-winding cancel, same-winding collapse, position-coincidence with distinct IDs, thin-plate preservation.

## Impact

Measured end-to-end against `baseplateGenerator.scenario.dovetail.test.ts` from gridfinity-layout-tool:

| Scenario | Before | After |
|---|---:|---:|
| 5×4 middle-column tile | 142 | **21** |
| 4×4 interior tile | 73 | **12** |
| 5×4 inverted middle | — | **6** |

82–85% reduction in non-manifold mesh edges. The tessellation dedup carries most of the load; the conditional unify is small but additive.

## Why not a full fix

The residual NM edges (6–21) come from **3 truly distinct triangles per edge** in the boolean's topology output — real branching where 3 faces with the same surface meet at one edge. Same root-cause class as the existing `#[ignore]`d `d4_shelled_box_fuse_lip` test.

A reference CAD-kernel design study suggests the architecturally-correct fix: in the assembler's same-domain detection, generalize beyond edge-set matching to use **point-in-face geometric verification**, and allow **within-rank pairing** (faces from the same input solid that overlap after splitting). Brepkit's `detect_same_domain` currently gates on identical edge sets AND opposing ranks — the dovetail residue trips both gates.

Tried and reverted on this branch:
- **Strict NM rejection in `validate_boolean_result`** — broke 15 currently-working ops tests that have 1–97 minor NM edges.
- **Mesh-level "branching resolver"** that drops triangles at 3+-incidence edges — broke `sweep_miter_l_shaped_volume_correct` (volume 10 → 4.8); miter joints legitimately produce 3-face junctions.
- **Relaxing `unify_same_domain`'s `len()==2` gate to `len()>=2`** — passes all tests but doesn't move the dovetail numbers; the merge phase's `edge_count==1` boundary filter then drops 3-shared edges as gaps, so the merge silently aborts.
- **Always invoking `unify_faces`** — broke `unify_boolean_box_reduces_faces` which tests behavior against pre-unify state.

The architectural fix (within-rank SD with point-in-face) is the right next step — too invasive for this PR.

## Verification

- [x] `cargo test -p brepkit-operations` — 603 pass (4 new dedup unit tests)
- [x] `cargo test --workspace --lib` — all pass
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean
- [x] `./scripts/check-boundaries.sh` — clean
- [x] End-to-end via gridfinity-layout-tool dovetail test — confirmed 82–85% NM-edge reduction across 3 failing scenarios

### DIFF
--- a/crates/operations/src/boolean/assembly.rs
+++ b/crates/operations/src/boolean/assembly.rs
@@ -478,10 +478,16 @@ fn build_manifold_shell(
     }
 
     // Find non-manifold edges (shared by 3+ faces).
-    let nonmanifold: Vec<(EdgeId, Vec<(usize, bool)>)> = edge_faces
+    // Sort by EdgeId so subsequent face-removal decisions are reproducible
+    // across runs — the same class of HashMap-iteration variance fixed in
+    // #689/#692, propagated to this pass. Order matters because each NM
+    // edge's "remove the opposing-normal face" decision interacts with
+    // others when faces are shared between multiple NM edges.
+    let mut nonmanifold: Vec<(EdgeId, Vec<(usize, bool)>)> = edge_faces
         .into_iter()
         .filter(|(_, faces)| faces.len() > 2)
         .collect();
+    nonmanifold.sort_by_key(|(eid, _)| eid.index());
 
     if nonmanifold.is_empty() {
         return Ok(face_ids.to_vec());
@@ -1369,10 +1375,15 @@ pub(super) fn split_nonmanifold_edges(
     }
 
     // Find non-manifold edges (shared by > 2 faces).
-    let nonmanifold: Vec<(usize, Vec<(usize, bool)>)> = edge_faces
+    // Sort by edge index for deterministic processing order — each NM-edge
+    // split mutates the topology (via edge_replacements), so order changes
+    // the assembly outcome when edges share faces. Same pattern fixed in
+    // #689/#692 for earlier GFA iteration sites.
+    let mut nonmanifold: Vec<(usize, Vec<(usize, bool)>)> = edge_faces
         .into_iter()
         .filter(|(_, faces)| faces.len() > 2)
         .collect();
+    nonmanifold.sort_by_key(|(eid, _)| *eid);
 
     if nonmanifold.is_empty() {
         return Ok(());
--- a/crates/operations/src/boolean/mod.rs
+++ b/crates/operations/src/boolean/mod.rs
@@ -468,7 +468,15 @@ pub fn boolean(
                 #[allow(clippy::cast_possible_wrap)]
                 let euler_pre2 = (v2 as i64) - (e2 as i64) + (f2 as i64);
 
-                if euler_pre2 != 2 {
+                // Run unify_faces if Euler is off (existing condition) OR if
+                // the topology already has 3+-face junctions, which can occur
+                // with Euler==2 when overlapping coplanar faces cancel in V-E+F
+                // counting. The same-domain detection in the assembler only
+                // pairs faces across opposing ranks with identical edge sets,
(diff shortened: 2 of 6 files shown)


andymai:
Addressed in `b7f8a80`:

- **Copilot (is_watertight doc)**: tightened the implementation to actually be a closed-2-manifold check — calls both `boundary_edge_count == 0` and `non_manifold_edge_count == 0`. Matches the existing docstring.
- **Greptile (1nm → 1µm)**: fixed in code comment + PR description. Constant is `1e-6`, which is 1µm.
- **Greptile (unreferenced positions)**: added a compact pass after the index rebuild — drops positions/normals not referenced by any surviving triangle.
- **Greptile (dedup uncovered by tests)**: four new direct unit tests in `tests.rs` — opposing-winding cancel, same-winding collapse, position-coincidence with distinct IDs, thin-plate preservation (1e-4mm-apart triangles must NOT collapse).

PR #699, opened 2026-05-20
fix(algo): within-rank same-domain detection with point-in-face containment (#696)

andymai:
## Summary

Architectural follow-up to PR #697 for issue #696. The assembler's same-domain detection only paired faces across opposing ranks with identical edge sets — boolean residue from sequential operations slips through both gates. This PR generalizes the SD model to also detect within-rank duplicates, including the common case of one planar face fully contained inside another's outer boundary.

## Why this matters

Each \`boolean()\` call's input A inherits faces from prior operations. When the consumer's \`booleanPipeline\` does many fuses + cuts in sequence, stale coincident faces accumulate in A. The previous \`detect_same_domain\` skipped them with \`if sub_faces[j].rank == rank_i { continue; }\`. Within-rank semantics are actually simpler than across-rank — no operation-specific selection logic needed, just "keep the representative, drop the duplicates."

## Changes

- **Generalize \`SameDomainResult\`**: returns both cross-rank \`pairs\` (unchanged semantics) and a new \`within_rank_dups: Vec<WithinRankDuplicate>\`.
- **Remove the rank skip** in the edge-set union pass so members from any rank combination can union when edge sets match and \`surfaces_same_domain\` agrees.
- **Add a planar containment pass** (\`planar_faces_overlap\`) for residue with different outer boundaries: for each pair of planar sub-faces sharing the same plane, test whether all of one face's outer-wire vertices (plus its interior sample) lie inside the other face's polygon. The conservative full-containment criterion is required — see "Tried and reverted" below.
- **Plumb \`within_rank_dups\` through \`bop::select_faces\`**: duplicates are dropped before the truth-table selection runs. Only the duplicate is forced to "On" classification; the representative goes through normal IN/OUT — otherwise it would silently fall out of selection.

## Impact

Cumulative end-to-end against \`baseplateGenerator.scenario.dovetail.test.ts\`, starting from main pre-#697:

| Scenario | main (pre-#697) | post-#697 | **This PR** |
|---|---:|---:|---:|
| 5×4 middle-column tile | 142 | 21 | **20** |
| 4×4 interior tile | 73 | 12 | **10** |
| 5×4 inverted middle | — | 6 | **6** |

The within-rank infrastructure catches the planar-full-containment subset of residue. Remaining NM edges follow patterns this PR doesn't address: non-planar surfaces (cylindrical nub side faces), and 3-face junctions with mixed surfaces. Those need different detection strategies and are left for follow-up.

## Tried and reverted

- **Interior-only overlap test** (\`A.interior in B.polygon OR B.interior in A.polygon\`) — broke \`fuse_ring_overlapping_shelled_box_height\` (volume 1972 → 1175). Adjacent coplanar faces with concave outlines can have an interior point that happens to land inside a neighbor's polygon without the faces actually overlapping. Switched to whole-wire-vertex containment which is the strict subset that catches residue without false positives.
- **Add both \`representative\` and \`duplicate\` to \`sd_indices\` in classify_sub_faces** — broke \`cut_box_by_translated_sphere\` (volume 0 → 1000). Forcing the representative to "On" classification removes it from normal selection but it has no matching \`SameDomainPair\`, so \`apply_sd_selection\` doesn't pick it up. Fixed by only adding the duplicate.

## Verification

- [x] \`cargo test --workspace\` — 603 ops tests pass, 0 regressions
- [x] \`cargo fmt --all --check\` — clean
- [x] \`./scripts/check-boundaries.sh\` — clean
- [x] End-to-end via gridfinity-layout-tool dovetail test — measurements above

## Test plan

- [ ] CI passes
- [ ] Existing analytic/GFA boolean tests show stable face counts and volumes
- [ ] Consumer integration confirms the dovetail NM-edge reductions

### DIFF
--- a/crates/algo/src/bop.rs
+++ b/crates/algo/src/bop.rs
@@ -324,4 +324,46 @@ mod tests {
         let selected = select_faces(&sub_faces, BooleanOp::Intersect, &[], &[]);
         assert_eq!(selected.len(), 2, "Intersect keeps only Inside faces");
     }
+
+    #[test]
+    fn within_rank_dup_drops_duplicate_keeps_representative() {
+        // idx 0 (representative) + idx 1 (duplicate): both rank A, Outside —
+        // would normally both be selected for Fuse. The dup record should
+        // remove idx 1, keep idx 0.
+        let mut topo = Topology::new();
+        let sub_faces = vec![
+            make_sub_face(&mut topo, Rank::A, FaceClass::Outside),
+            make_sub_face(&mut topo, Rank::A, FaceClass::Outside),
+            make_sub_face(&mut topo, Rank::B, FaceClass::Outside),
+        ];
+        let dups = vec![WithinRankDuplicate {
+            representative: 0,
+            duplicate: 1,
+        }];
+        let selected = select_faces(&sub_faces, BooleanOp::Fuse, &[], &dups);
+        assert_eq!(
+            selected.len(),
+            2,
+            "Fuse should keep representative + B-Outside, drop within-rank duplicate"
+        );
+    }
+
+    #[test]
+    fn within_rank_dup_out_of_bounds_skips_gracefully() {
+        let mut topo = Topology::new();
+        let sub_faces = vec![
+            make_sub_face(&mut topo, Rank::A, FaceClass::Outside),
+            make_sub_face(&mut topo, Rank::B, FaceClass::Outside),
+        ];
+        let dups = vec![WithinRankDuplicate {
+            representative: 5,
+            duplicate: 10,
+        }];
+        let selected = select_faces(&sub_faces, BooleanOp::Fuse, &[], &dups);
+        assert_eq!(
+            selected.len(),
+            2,
+            "Out-of-bounds within-rank-dup record should be ignored, not panic"
+        );
+    }
 }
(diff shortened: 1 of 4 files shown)


andymai:
Addressed in `fba1a14`:

- **Greptile P1** (cross-rank containment flag): track which UF groups came from the geometric pass via `geometric_overlap_groups: HashSet<root>`, and set `b_contained_in_a: true` for cross-rank pairs in those groups. Edge-set-matched groups still get `false` (touching, not overlapping). Pinned by the new `cross_rank_geometric_containment_marks_overlapping` test.

- **Greptile P2** (misleading comment): rewrote — the implementation uses strict `point_in_polygon_2d` with no tolerance; the comment now matches and warns callers that vertices exactly on the containing boundary may be missed.

- **Copilot** (`bop.rs` tests): added `within_rank_dup_drops_duplicate_keeps_representative` and `within_rank_dup_out_of_bounds_skips_gracefully`.

- **Copilot** (`same_domain.rs` tests): added `detects_within_rank_planar_containment`, `adjacent_coplanar_faces_not_duplicates` (regression guard against the interior-only test that broke ring+shell during development), and `cross_rank_geometric_containment_marks_overlapping` (pins the P1 fix).

5 new tests pass, 0 regressions.

PR #701, opened 2026-05-20
feat(wasm): bridge Rust log calls to JS console (#696 diagnostics)

andymai:
## Summary

Adds a `setLogLevel(level)` WASM export that routes brepkit's Rust `log::*` calls to JS `console.{log, warn, error}`. Without this every `log::warn!` / `log::info!` / `log::debug!` in the engine is silently dropped under wasm-pack — useful boolean and validation diagnostics never reach the consumer's stderr.

Issue #696 investigation hit a wall here: I could measure the symptom downstream (non-manifold edges in the tessellated STL) but couldn't see why brepkit produced the broken topology. The bridge unblocks that diagnostic loop.

## Changes

New `crates/wasm/src/logging.rs`:

- `ConsoleLogger` implements `log::Log`, routing by level to JS `console.log` / `console.warn` / `console.error` via wasm-bindgen extern decls. Messages are tagged `[brepkit:<module>]` so the source is visible.
- `setLogLevel(level)` wasm-bindgen export accepting `"off"`, `"error"`, `"warn"`, `"info"`, `"debug"`, `"trace"` (case-insensitive). Idempotent; call again to change the filter. Unknown values default to `"off"` so a typo doesn't produce a flood.
- Logger installs lazily on first `setLogLevel` call; default is `"off"` so the bridge has zero overhead when unused.
- Runtime level held in an `AtomicUsize` so the filter is hot-swappable after install — `log::set_logger` succeeds at most once globally.

Works under both `--target nodejs` and `--target web` (both expose `console.*` as a JS global).

## Verification

**Direct Node test** — `setLogLevel("trace")` followed by a box-box fuse prints e.g.

    [brepkit:brepkit_algo::pave_filler::phase_ef] EF: edge Id(0) crosses face Id(10) at t=0.250000
    [brepkit:brepkit_algo::pave_filler::phase_ff] FF: faces Id(1) and Id(8) intersect ...

**End-to-end against the gridfinity dovetail failure** with `setLogLevel("warn")` immediately revealed that **every** boolean op in the consumer's connector pipeline produces invalid GFA output and falls back to mesh boolean:

    [brepkit:brepkit_operations::boolean] GFA result failed validation in 0.0ms (faces=29), falling back
    [brepkit:brepkit_operations::boolean::assembly] boolean result has 5 validation error(s):
    Euler characteristic V-E+F = -11 is invalid ...; edge 1051 is shared by 3 faces; ...
    [brepkit:brepkit_algo::builder::face_splitter] face_splitter: hole with 8 edges could not be assigned to any sub-face

This overturns my earlier assumption (PR #699) that within-rank SD residue from successful GFA was the dominant pattern. The bridge paid for itself in one test run — the residual NM-edge fix is now a properly informed follow-up.

## Usage from JS

```js
import { setLogLevel } from 'brepkit-wasm';
setLogLevel('warn'); // recommended starting point for diagnostics
```

## Test plan

- [x] `cargo test --workspace` — all pass (no behavior change; bridge is inert until `setLogLevel` is called)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean
- [x] `./scripts/check-boundaries.sh` — clean
- [x] Manual end-to-end Node test confirms log output flows
- [x] Manual end-to-end vitest run confirms log output reaches stderr

### DIFF
--- a/crates/wasm/src/lib.rs
+++ b/crates/wasm/src/lib.rs
@@ -12,6 +12,7 @@ pub mod error;
 mod handles;
 mod helpers;
 pub mod kernel;
+mod logging;
 pub mod shapes;
 mod state;
 mod types;
--- a/crates/wasm/src/logging.rs
+++ b/crates/wasm/src/logging.rs
@@ -1,0 +1,105 @@
+//! Bridge brepkit's `log` crate calls to JS `console.{log, warn, error}`.
+//!
+//! Without this, all `log::warn!` / `log::info!` etc. throughout the Rust
+//! code are silently dropped under wasm-pack — useful diagnostic output
+//! never reaches the consumer's stderr. The bridge sets up a logger that
+//! routes by level:
+//!
+//! - `error` → `console.error`
+//! - `warn`  → `console.warn`
+//! - `info` / `debug` / `trace` → `console.log` (with level prefix)
+//!
+//! Initialize once via `BrepKernel.setLogLevel(level)` from JS. Calling
+//! again with a different level updates the runtime filter without
+//! reinstalling the logger. Default level (if `setLogLevel` is never
+//! called) is **off** — zero overhead, no console noise.
+//!
+//! Works under both `wasm-pack --target nodejs` and `--target web` since
+//! `console.*` is a universal JS global.
+
+use core::sync::atomic::{AtomicUsize, Ordering};
+
+use log::{Level, LevelFilter, Log, Metadata, Record};
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = console, js_name = log)]
+    fn console_log(s: &str);
+    #[wasm_bindgen(js_namespace = console, js_name = warn)]
+    fn console_warn(s: &str);
+    #[wasm_bindgen(js_namespace = console, js_name = error)]
+    fn console_error(s: &str);
+}
+
+/// Runtime level filter. Defaults to `off` (= 0). `set_log_level` writes here
+/// rather than mutating `log::max_level` so the level can change after the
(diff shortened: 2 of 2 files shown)


andymai:
Addressed in \`cfc6eec\`:

- **Greptile P2 + Copilot (LEVEL redundancy)**: dropped the \`AtomicUsize LEVEL\`. \`log::set_max_level\` already provides the runtime-mutable fast-path filter the \`log\` framework checks before dispatch — \`ConsoleLogger::enabled\` now reads \`log::max_level()\` directly. The stated rationale didn't hold.
- **Greptile P2 + Copilot (module doc)**: corrected "Initialize once via \`BrepKernel.setLogLevel\`" — the binding is a standalone export, JS call-site is \`setLogLevel('warn')\`.
- **Copilot ("with level prefix")**: removed the inaccurate claim. Messages are \`[brepkit:<target>] ...\` without an embedded level (console.log/warn/error already distinguishes severity).
- **Greptile P2 (silent invalid-level)**: \`parse_level\` returns \`Option<LevelFilter>\`; \`setLogLevel\` returns \`Result<(), JsError>\`. Unknown levels throw a JS Error listing the accepted values rather than silently defaulting to \`off\`. Verified end-to-end:

  \`\`\`
  wasm.setLogLevel("warn");   // ok
  wasm.setLogLevel("INFO");   // case-insensitive
  wasm.setLogLevel("warnn");  // throws: setLogLevel: unknown level "warnn"; expected one of "off", ...
  \`\`\`
- **Copilot (no unit tests)**: added a \`#[cfg(test)]\` module with three tests — known values, case-insensitivity, unknown-string rejection.

3 new tests pass, 0 regressions.

PR #703, opened 2026-05-20
fix(algo): preserve dropped holes in face splitter (#696 diagnostic)

andymai:
## Summary

Fix for the \`face_splitter: hole with 8 edges could not be assigned to any sub-face\` warning surfaced by PR #701's diagnostic logging. The face splitter was silently dropping holes whose first vertex landed on a sub-face boundary after splitting — geometry was lost without an error.

## Bug

The "original holes" distribution loop in \`split_face_2d\` tested only the hole's first vertex against each sub-face's outer polygon. When the splitter cuts the face along a line that passes through a hole vertex (common at split intersections), that vertex sits exactly on a sub-face boundary and \`point_in_polygon_2d\`'s strict ray-cast returns false for every sub-face. The hole was logged and dropped — the sub-face ended up without its cutout, producing wrong topology downstream.

The "new holes" loop just above had a fallback (\`sub_faces.first_mut().inner_wires.push(hole)\`); the "original holes" loop forgot it.

## Fix

Two changes:

1. **Test the hole's interior sample** (\`sample_interior_point\` of the hole's UV polygon) instead of its first vertex. A point inside the hole's enclosed area is robust against split lines passing through the hole's boundary vertices.

2. **Fall back to the largest-area sub-face** when even the interior sample doesn't match (rare: hole straddling multiple sub-faces, numerical edge case). A warning still fires for visibility, but the hole is preserved instead of dropped.

## Impact

End-to-end against the dovetail consumer test:

- Warning changes from \`could not be assigned to any sub-face\` to \`did not contain-test in any sub-face; attaching to largest sub-face as fallback\` — the hole is preserved.
- NM-edge counts shift slightly: 20→21 (5×4 middle), 10→12 (4×4 interior), 6→6 (5×4 inverted). This is the expected cost of converting silent data loss into visible (and topologically faithful) output — the previously-dropped hole now contributes geometry that downstream passes have to deal with.

The proper next step is to **split the straddling hole at the cut line** so it can be distributed across sub-faces correctly, but that's a deeper refactor; this PR is the immediate "stop silently dropping geometry" fix.

## Verification

- [x] 603 ops tests pass, full workspace clean (0 regressions)
- [x] \`cargo fmt --all --check\` — clean
- [x] \`./scripts/check-boundaries.sh\` — clean
- [x] End-to-end consumer test confirms the fallback fires once (matches the single original warning)

## Test plan

- [ ] CI passes
- [ ] No regression in volume / face-count assertions in other ops tests

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -486,58 +486,65 @@ pub fn split_face_2d(
     }
 
     // Distribute original inner wires (holes from the source face) to sub-faces.
-    // Each hole is assigned to the sub-face whose outer wire contains it.
-    //
-    // Test with the hole's interior sample (a point inside the hole's enclosed
-    // area), not its first vertex — when the splitter cuts the face along a
-    // line that passes through a hole vertex, that vertex sits exactly on a
-    // sub-face boundary and `point_in_polygon_2d`'s strict ray-cast returns
-    // false for every sub-face. The earlier code logged a warning and
-    // silently dropped the hole, leaving the sub-face missing its cutout.
-    // Falling back to the largest-area sub-face when even the interior
-    // sample doesn't match keeps the geometry intact for the rare residual
-    // cases (e.g., hole straddling multiple sub-faces); a warning still fires
-    // so the case is visible.
+    // Each hole is assigned to the sub-face whose outer wire contains its
+    // interior sample point (a point inside the hole's enclosed area, not
+    // its first vertex — that vertex often sits exactly on a sub-face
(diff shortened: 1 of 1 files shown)


andymai:
Addressed in \`128794a\`:

**Greptile P1 (silent drop on degenerate sample)**: routed the \`hole_pts.len() < 3\` branch through the same largest-area fallback the no-containment-match case uses. A warning fires listing the reason (\`degenerate UV sample (<3 pts)\` vs \`did not contain-test in any sub-face\`). Refactored to share the "find largest sub-face" lookup between the two fallback paths.

The \`continue;\` regression is gone — no path now drops a hole silently. Tests + clippy + fmt clean.

PR #705, opened 2026-05-20
test(boolean): n-iteration repro tests for #696 + minimal box-cut-pocket

andymai:
## Summary

Two \`#[ignore]\`d diagnostic tests that scope issue #696 (dovetail baseplate non-manifold STL) to a much smaller failure surface than the gridfinity-layout-tool consumer chain. **No production code changes** — these are documentation + investigation handles.

### \`n_iteration_repro_dovetail_pipeline_issue_696\`

Mimics the consumer's slab → 16 pockets → connector nubs pipeline at 168×168×8mm and prints topology metrics after each step:

\`\`\`
step                         topology counts          issues
0. slab                      F=  6 E= 12 V=  8 Euler=  2 ✓ NM=0 bd=  0
1. pocket cut #1             F= 11 E= 30 V= 20 Euler=  1 ✗ NM=0 bd=  4
2. pocket cut #2             F= 17 E= 54 V= 36 Euler= -1 ✗ NM=0 bd= 11
4. pocket cut #4             F= 34 E=124 V= 80 Euler=-10 ✗ NM=0 bd= 41
16. pocket cut #16           F= 79 E=262 V=188 Euler=  5 ✗ NM=0 bd=147
20. nub row 1 (×4 sides)     F= 99 E=334 V=231 Euler= -4 ✗ NM=2 bd=180
24. nub row 2 (×4 sides)     F= 99 E=334 V=231 Euler= -4 ✗ NM=2 bd=180
28. nub row 3 (×4 sides)     F= 99 E=334 V=231 Euler= -4 ✗ NM=2 bd=180
\`\`\`

Two findings:
1. The **first pocket cut already breaks Euler** (1 vs expected 2) and introduces 4 boundary edges.
2. After the first nub fuse, subsequent fuses are **silent no-ops** — F/E/V are identical across steps 20/24/28. The boolean returns unchanged geometry because the input is already broken.

### \`minimal_box_cut_pocket_should_be_manifold\`

The minimal repro — single box cut by a single pocket box. Currently fails with F=11 E=30 V=20 Euler=1 NM=0 boundary=4.

**Root cause identified**: the 4 extra vertices come from the **diagonals of the pocket's tessellated side faces intersecting the slab top plane**. Each pocket vertical face (a 37×6 rectangle) is triangulated with a diagonal corner-to-corner; \`mesh_boolean\` splits that diagonal at z=0, producing intermediate intersection points like \`(-69.166667, -44.5, 0)\` — exactly \`-81.5 + 37/3\`, i.e., where the diagonal crosses z=0. These intermediates don't exist in BREP geometry; they're tessellation artifacts.

4 such artifacts (one per pocket side face) survive vertex merging because the slab top face's hole inner wire uses them, but the pocket vertical faces below z=0 use 3-edge outlines that don't all share the same edges → 4 unpaired half-edges → boundary edges.

GFA also fails this case (separate bug, surfaces as "4 boundary edges" in the GFA validate logs); the mesh-fallback inherits this pattern from \`mesh_boolean\`'s tessellation-driven topology.

## Use

Invoke explicitly:

\`\`\`
cargo test -p brepkit-operations --lib minimal_box_cut_pocket \\
    -- --ignored --nocapture

cargo test -p brepkit-operations --lib n_iteration_repro \\
    -- --ignored --nocapture
\`\`\`

Neither runs in CI; they're handles for future investigation.

## Where this leaves #696

The dovetail bug is downstream of a **simpler bug** in box-pocket cut. If \`minimal_box_cut_pocket_should_be_manifold\` is made to pass, the rest of the dovetail pipeline should follow because every step depends on each individual boolean being clean.

The likely fix direction is in either:
- **\`mesh_boolean\`**: drop intersection points that don't lie on a true BREP boundary (i.e., dedup along intersection segments, keeping only segment endpoints)
- **\`assemble_solid_mixed\`**: detect when "intermediate" vertices on a coplanar segment are between two other vertices on the same segment and merge them into the segment

Either fixes the symptom at the topology-assembly layer; the cleanest fix is at the mesh_boolean layer.

## Verification

- [x] 603 ops tests pass (24 ignored: existing + 2 new diagnostic)
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --all --check\` clean

## Test plan

- [ ] CI passes
- [ ] Run \`minimal_box_cut_pocket\` locally and confirm the failure output matches the description above
- [ ] Open follow-up tracking issue for the mesh_boolean tessellation-artifact fix

### DIFF
--- a/crates/operations/src/boolean/tests.rs
+++ b/crates/operations/src/boolean/tests.rs
@@ -3200,6 +3200,27 @@ fn coplanar_box_cut_d1a2() {
 }
 
 // ── #696 diagnostic: N-iteration repro ───────────────────────────────
+
+/// Count edges shared by 3+ faces (non-manifold) and edges shared by < 2
+/// faces (boundary) across all wires of a solid's faces. Shared between
+/// the #696 diagnostic tests so they always measure the same thing.
+fn count_nm_and_boundary_edges_696(topo: &Topology, solid: SolidId) -> (usize, usize) {
+    let mut edge_count: std::collections::HashMap<usize, usize> = std::collections::HashMap::new();
+    let faces = brepkit_topology::explorer::solid_faces(topo, solid).unwrap();
+    for fid in &faces {
+        let face = topo.face(*fid).unwrap();
+        for wid in std::iter::once(face.outer_wire()).chain(face.inner_wires().iter().copied()) {
+            let wire = topo.wire(wid).unwrap();
+            for oe in wire.edges() {
+                *edge_count.entry(oe.edge().index()).or_default() += 1;
+            }
+        }
+    }
+    let nm = edge_count.values().filter(|&&c| c > 2).count();
+    let bd = edge_count.values().filter(|&&c| c < 2).count();
+    (nm, bd)
+}
+
 //
 // The gridfinity-layout-tool dovetail tests fail with 6–20 non-manifold
 // mesh edges in the exported STL. Diagnostic logging from #701 showed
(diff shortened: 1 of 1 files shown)


andymai:
Addressed in \`b936d91\`:

- **Greptile P2 (dead \`d * 0.0\`)**: dropped — replaced with plain \`wall_x\` so the base-vs-tip asymmetry is unambiguous.
- **Greptile P2 (duplicated NM/boundary counting)**: extracted to module-level \`count_nm_and_boundary_edges_696\` helper. Both tests call it instead of copy-pasting.
- **Copilot (winding comment)**: rewrote — the point order genuinely flips with \`protrude_dir\` to align the face normal with the +Z extrude direction. The comment now explains that asymmetry.
- **Copilot (unused notes column)**: removed.

Diagnostic output unchanged; 603 ops tests pass, clippy + fmt clean.

PR #706, opened 2026-05-20
fix(operations): planarity-aware tessellation-artifact dedup in mesh_boolean (#696)

andymai:
## Summary

Closes the minimal box-pocket repro from PR #705 (single 168×168×8 slab cut by 37×37×6 pocket → 4 boundary edges before, 0 after) and moves the dovetail end-to-end NM-edge counts further down without regressing any of the existing 603 ops + 48 stress + 184 wasm tests. Builds on the threading direction from #696 issue thread.

## What

The bug: when one mesh's planar face is crossed by two triangles from the other mesh that share a tessellation-diagonal edge, each contributes an intersection segment ending at the diagonal-vs-plane crossing. That crossing is mathematically collinear with the segments' outer endpoints but isn't a real BREP vertex — purely a side-effect of the diagonalization. Keeping it splits the triangle into two collinear sub-edges that don't pair with anything on the other side, leaving boundary edges in the assembled solid.

The prior naive "drop any collinear midpoint" attempt (investigated in #696 issue thread) regressed \`fuse_two_cylinders\`: cylinder-touch-line intermediates ARE load-bearing tessellation vertices, and dropping them created T-junctions on the cylinder sides.

## How

Two safety gates make the drop selective:

1. **Per-triangle planarity flag**, computed from topology by matching each tessellated triangle's centroid + face normal against the input solid's planar face plane equations. Triangles whose source surface is curved (cylinder, sphere, NURBS) don't match any planar face and stay flagged \`false\` — their intermediates pass through.

2. **Coplanar segment guard**: when ANY intersection segment on this triangle came from \`intersect_coplanar_triangles\` (e.g. two cylinder caps overlapping at z=0), skip the drop. Coplanar intersections produce intermediates that represent real cap-on-cap geometry, not tessellation diagonals.

## API additions

- \`mesh_boolean_with_metadata(mesh_a, planar_a, mesh_b, planar_b, op, tolerance)\` accepts optional per-triangle planarity slices. \`mesh_boolean\` (unchanged signature) is now a thin wrapper that calls it with \`None\` for both — no behavior change for direct callers.
- \`TriTriIntersection\` gains a \`from_coplanar: bool\` field so the splitter can identify coplanar-source segments.
- \`infer_planar_triangle_flags(topo, solid, mesh, tol)\` (private helper in \`boolean/mod.rs\`) derives the planarity slice for a tessellated solid via face-equation matching.
- \`mesh_boolean_fallback\` computes per-mesh planarity and threads it through.

## Impact

End-to-end against gridfinity-layout-tool dovetail tests, cumulative from pre-#697:

| Scenario | main | post-#697+#699+#703 | **this PR** |
|---|---:|---:|---:|
| 5×4 middle-column | 142 | 20 | **15** |
| 4×4 interior | 73 | 10 | **4** |
| 5×4 inverted middle | — | 6 | 8 |

4×4 interior dropped significantly (10 → 4); 5×4 middle improved (20 → 15); 5×4 inverted ticked up by 2 (6 → 8) — likely points at a different residue pattern this fix doesn't address.

## Test fixup

\`minimal_box_cut_pocket_should_be_manifold\` (added \`#[ignore]\`d in #705) is un-\`#[ignore]\`d and now passes: \`F=11 E=28 V=20 Euler=3 NM=0 boundary=0\`. Its strict \`Euler == 2\` assertion was loosened to \`NM == 0 && bd == 0\` — Euler equals \`2 + inner_loop_count\` (3 here for the slab top's one inner wire), so the strict check would have failed even for a fully-correct result.

## Verification

- [x] 604 ops tests pass (was 603 + the now-enabled minimal test)
- [x] 48 boolean_stress tests pass — no regression on fuse_two_cylinders, cut_cylinder_from_cylinder, etc.
- [x] Full workspace test passes (940+ tests, zero failures)
- [x] \`cargo clippy --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --all --check\` clean
- [x] End-to-end gridfinity dovetail test confirms the NM reductions above

## Test plan

- [ ] CI passes
- [ ] No regression in volume / face-count assertions across the pre-existing ops tests
- [ ] Consumer integration confirms the dovetail NM reductions

### DIFF
--- a/crates/operations/src/boolean/mod.rs
+++ b/crates/operations/src/boolean/mod.rs
@@ -1697,7 +1697,22 @@ fn mesh_boolean_fallback(
     let mesh_a = crate::tessellate::tessellate_solid(topo, a, deflection)?;
     let mesh_b = crate::tessellate::tessellate_solid(topo, b, deflection)?;
 
-    let mb_result = crate::mesh_boolean::mesh_boolean(&mesh_a, &mesh_b, op, tol.linear)?;
+    // Compute per-triangle "is on a planar face" flags by matching each
+    // triangle's centroid + normal against the input solid's planar
+    // face equations. Used by mesh_boolean to drop tessellation-diagonal
+    // artifacts on planar faces while keeping load-bearing intermediates
+    // on curved surfaces (issue #696).
+    let planar_a = infer_planar_triangle_flags(topo, a, &mesh_a, tol);
+    let planar_b = infer_planar_triangle_flags(topo, b, &mesh_b, tol);
+
+    let mb_result = crate::mesh_boolean::mesh_boolean_with_metadata(
+        &mesh_a,
+        Some(&planar_a),
+        &mesh_b,
+        Some(&planar_b),
+        op,
+        tol.linear,
+    )?;
     let face_specs = mesh_result_to_face_specs(&mb_result);
     if face_specs.is_empty() {
         return Err(crate::OperationsError::EmptyResult {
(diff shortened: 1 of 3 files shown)


andymai:
Addressed in \`2358d93\`:

- **Copilot + Greptile P2 (dead \`other_mesh\` param)**: removed entirely. \`split_mesh_by_intersections\` no longer takes it; both call sites updated. Cross-mesh planarity is a later-PR thing if it's ever needed.

- **Copilot (unnormalized plane equations)**: \`infer_planar_triangle_flags\` now normalizes each topology plane (\`n /= |n|\`, \`d /= |n|\`) before storing. The cosine comparison and point-to-plane distance now use unit normals as intended.

- **Greptile P2 (dimensional mismatch in degenerate guard)**: switched from \`fn_len < lin_tol\` (length² vs length) to \`area_sq < lin_tol²\` (length² vs length²). The effect was already conservative — triangle stayed unflagged either way — but the threshold now correctly means "edges too short to produce a reliable normal".

- **Greptile P2 (O(n³) loop)**: tried the suggested extreme-pair O(n²) variant and reverted. Multi-line case breaks it: when a triangle is crossed by intersection segments along multiple distinct lines (e.g. all four sides of a rectangular pocket hole), the extreme pair captures only one diagonal and misses the per-side midpoints. Pocket test regressed back to \`F=11 E=30 V=20 NM=0 bd=4\`. Kept the O(n³) scan with an updated comment explaining why and noting that n is 2–8 in practice (the rectangular hole hits the high end).

Tests + clippy + fmt clean. 604 ops tests pass; pocket stays manifold; cylinder fuse unbroken.

PR #743, opened 2026-05-29
fix(fillet): variable-radius fillet removes material on convex edges

andymai:
## Problem

The variable-radius (canal-surface) fillet path inflated solid volume on convex
edges: filleting an edge of a unit cube produced a solid with volume greater than
the original. A fillet on a convex edge must *remove* material, so this was
geometrically impossible and produced a malformed shell.

## Root causes

1. **Wrong in-plane contact direction sign.** The cross-section direction toward
   each face's rolling-ball contact was selected by comparing against the
   neighbouring face's outward normal with the wrong sign, so the blend bulged
   *outward* (away from the material) instead of inward.
2. **Arc middle control point placed away from the edge.** The rational-quadratic
   arc's middle control point was offset along the bisector away from the sharp
   edge, so the arc bulged outward rather than toward the material.
3. **Seam vertices not shared.** Trimmed neighbour-face boundary vertices were
   derived independently from the blend rails, producing a non-watertight seam.

## Fix

- Add `geometry::inward_section_dir(tangent, n_self, n_other)` — the in-plane
  direction pointing into the material wedge (opposite the face's own outward
  normal), with the sign fixed unambiguously by the neighbour normal. Used for
  both the per-station canal build and the t=0 fallback.
- Place the arc middle control point at the original sharp-edge point so the
  rational arc bulges toward the material (convex fillet that removes material).
- Pre-compute the rail contact point at each filleted-edge endpoint keyed by
  (vertex, edge, face); trimming and the blend rails reuse these exact positions
  for a watertight seam.
- Set the blend strip's `reversed` flag from the bisector so the strip's outward
  normal points out of the solid.

This also removes the previous workaround note about ~30% volume inflation on the
all-edges path — the corrected direction convention resolves it.

## Tests

- `fillet_variable_linear_law` now asserts the filleted unit cube has volume < 1.
- `fillet_variable_linear_law_removes_material` — box(20,20,20), single-edge
  linear law: result is manifold, has exactly one added blend face, and
  7000 < V < 8000 (material removed).
- `fillet_variable_constant_law_matches_rolling_ball` — a constant-law variable
  fillet matches the rolling-ball fillet volume within 1%.

All `brepkit-operations` fillet tests pass; fmt, clippy (`-D warnings`),
crate-boundary checks, and the full pre-push suite are green.

### DIFF
--- a/crates/operations/src/fillet/geometry.rs
+++ b/crates/operations/src/fillet/geometry.rs
@@ -94,6 +94,19 @@ pub(super) fn edge_v_samples(curve: &EdgeCurve) -> usize {
     }
 }
 
+/// In-plane direction from the edge toward the rolling-ball contact on a face.
+///
+/// Within the section plane (normal `tangent`), this is the face normal `n_self`
+/// projected onto the plane, with the sign flipped so it points into the solid
+/// material (toward the ball center) rather than out along the outward normal.
+/// `n_other` is the outward normal of the neighbouring face; the wedge interior
+/// lies on the side away from `n_other`, which fixes the sign unambiguously.
+pub(super) fn inward_section_dir(tangent: Vec3, n_self: Vec3, n_other: Vec3) -> Option<Vec3> {
+    let c = tangent.cross(n_self);
+    let dir = if c.dot(n_other) < 0.0 { c } else { -c };
+    dir.normalize().ok()
+}
+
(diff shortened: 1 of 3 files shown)


andymai:
Closing as duplicate. This gap (variable-radius fillet inflating volume on convex edges) was already fixed and released in #739 (v2.101.1), which landed an equivalent fix to the same three files. This branch was cut from an older main before #739 merged. No action needed.

PR #790, opened 2026-06-13
fix(algo): keep coincident same-domain cap faces in fuse/intersect

andymai:
## Summary

Un-ignores `boolean_stress.rs::fuse_cylinder_and_box` (a cylinder resting flush on a box floor) and fixes the underlying GFA defect.

## The bug

A flush cap that lands exactly on a planar face gets split into a **coincident same-domain (SD) pair** — here the cylinder's bottom cap and the disk sub-face carved out of the box floor (both at z=0, same outward normal −z).

`classify_sub_faces` exempted **disc sub-faces** (a single closed-curve outer loop) from the usual "force SD faces to `On`" rule and instead ray-cast them, sampling an interior point deliberately offset *into* the solid. For an SD-coincident disc that sample lands inside the opposing solid → classified `Inside` → `apply_sd_selection` discarded **both** faces of the pair, leaving a hole in the box floor.

That hole is a free edge → open shell → the result fails the manifold acceptance gate in `boolean()` → silent fallback to the faceted **mesh boolean** (28 garbage faces, non-manifold). Calling GFA directly shows the clean-but-holed 8-face analytic result.

## The fix

Force **every** SD sub-face (disc or not) to `On`, so `apply_sd_selection` keeps exactly one representative of each cross-rank pair — consistent with how all other SD faces are already handled. A same-oriented coincident pair always lies on the result's exterior (both solids' material is on the same side of the shared plane), so the prior "discard when the representative is `Inside`" branch was never geometrically correct and is removed; genuinely-internal coincident faces have **opposite** orientation and are still cancelled by the else-branch.

Within-rank duplicates are dropped by `select_faces` regardless of classification, so `On` is safe for them too. The pre-computed disc interior point is still used (correctly) by non-SD discs — e.g. the box *top* at z=2, a genuine through-hole — so `face_splitter.rs` is untouched.

Net effect: this **removes** a fragile ray-cast at a coincident boundary rather than adding logic.

## Verification

- `fuse_cylinder_and_box` passes; 10× in fresh processes (no HashMap-iteration nondeterminism).
- `brepkit-algo` 112/112, `brepkit-operations` full suite green (0 regressions — incl. `coaxial_torus`, `regress_hexwall_cuts`, all `coincident_*`/`coaxial_*` boolean suites), `brepkit-io` 156, `brepkit-wasm` 210.
- `./scripts/check-boundaries.sh` clean (change is contained to `brepkit-algo`).

## Note

Also includes `chore(hooks)`: the pre-push hook now delegates the full-suite run to CI instead of duplicating it locally.

⚠️ Touches the GFA builder's SD classification (a core boolean path) — flagging for human review rather than auto-merge.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,16 +1,9 @@
 #!/usr/bin/env bash
 
-# Full test suite + dependency audit before push
-echo "🔍 Running pre-push checks..."
-
-cargo test --workspace || {
-  echo "❌ Tests failed."
-  exit 1
-}
-
-cargo deny check 2>/dev/null || {
-  echo "⚠️  cargo-deny not installed, skipping license/advisory check."
-  echo "   Install with: cargo install cargo-deny"
-}
-
-echo "✅ Pre-push checks passed."
+# Validation is delegated to CI (.github/workflows/ci.yml): the `test` job
+# runs the full `cargo nextest run --workspace` plus doc tests, and `deny`,
+# `fmt`, `clippy`, `boundaries`, `audit`, `docs`, `machete`, and `taplo` run
+# alongside it — all gated by the `ci-pass` job. Re-running the whole suite
+# here just doubled the wait, so the push side relies on CI. Fast local
+# feedback still comes from the pre-commit hook (fmt + clippy + machete).
+echo "ℹ️  Pre-push: validation delegated to CI (.github/workflows/ci.yml)."
--- a/crates/algo/src/bop.rs
+++ b/crates/algo/src/bop.rs
@@ -160,7 +160,7 @@ fn apply_sd_selection(
 
         if same_ori_needed == pair.same_orientation {
             // Orientations match what the operation needs:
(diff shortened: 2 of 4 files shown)


andymai:
Thanks for the reviews — addressed:

- **Copilot — `bop.rs:165` (Cut comment misleading):** Fixed in b7e9531. The Cut branch keys off `is_touching` (`b_contained_in_a`), not `sf_a.classification`; the bullet now reads "keep A only when the faces merely touch on the boundary; discard both when A overlaps B's interior."

- **Greptile P2 — pre-push hook now delegates to CI:** Intentional and requested. The full `cargo test --workspace` was a duplicate of the CI `test` job (`cargo nextest run --workspace` + doc tests), and `ci-pass` gates the PR on test + deny + fmt + clippy + boundaries + audit + docs + machete + taplo. Fast local feedback still comes from the pre-commit hook (fmt + clippy + machete). Waiving — no change.

PR #792, opened 2026-06-13
fix(algo): resolve d4 shelled-box + lip fuse (holed-face & section-arrangement splitting)

andymai:
## Summary

Un-ignores `boolean::tests::d4_shelled_box_fuse_lip` — the long-standing "D4 gridfinity" shelled-box + lip fuse that produced non-manifold, wrong-genus topology. The result is now a genus-0 closed manifold (adjusted Euler == 2), verified 10× deterministic.

Four independent face-splitter fixes, each regression-clean on its own, were needed — the bug was a stack of distinct defects in how a planar face is partitioned by section edges. They generalize well beyond d4 (any boolean where tool walls cross a holed/rim face).

## The four fixes

1. **`c1d9046` — drop pendant section edges** (reference `PerformShapesToAvoid`). A section that doesn't connect two parts of the face boundary dangles into the interior; the angular wire traversal walks out and back along it, spuriously over-splitting the face. (nm 10→8)

2. **`91b55b0` — T-junction plane-face split.** The geometric fallback handled an X-crossing of two sections (4 quadrants) but not a **T-junction** (one section's endpoint mid-way on another → 3 regions). Generalized `try_split_crossing_plane_face` to N regions and made the caller prefer the geometric split when it yields more regions than the wire builder. (nm 8→2)

3. **`d50caee` — split hole boundaries when a holed plane face is cut by sections.** A face carrying an inner-wire hole (the shelled box top with its cavity opening) was split only on its outer boundary; the hole was attached whole. When the lip walls cross the opening, the sub-face ended up as a square carrying the over-sized cavity hole instead of the true L-shaped rim. Now the hole boundaries are woven into the section arrangement (sections trimmed at hole crossings, hole edges split), so the wire builder traces the real material region. (nm 2→0, free 6→0 — closed manifold)

4. **`da0d7dc` — deduplicate coincident section edges.** A face-face interference can be recorded twice for one face (the box outer walls each carried a duplicated y=8/x=8 section). A duplicated *dividing* section makes the wire builder weave a zero-area **slit** instead of splitting — and a slit gives a face the Euler of an annulus (0) rather than a disk (1), so each slit reads as one unit of spurious genus. Two slit walls → apparent genus-1 (Euler 0 vs 2) even on an otherwise-clean closed manifold. Dedup by endpoints + a midpoint sample (so split-circle half-arcs aren't collapsed), and the surviving single section forms a clean T-junction that fix #2 resolves.

## Verification

- `d4_shelled_box_fuse_lip` passes, 10/10 in fresh processes (no HashMap nondeterminism).
- Full suite green after each commit: `brepkit-algo`, `brepkit-operations`, `brepkit-io`, `brepkit-wasm` — 0 regressions.
- `./scripts/check-boundaries.sh` clean (all changes contained to `brepkit-algo`).

## Notes

- The sibling `fuse_shelled_box_with_socket_loft` (curved loft socket) is a separate, deeper case and stays ignored.
- ⚠️ These touch the GFA face splitter (a core boolean path). Flagging for human review rather than auto-merge.

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -40,6 +40,193 @@ use special_cases::{
 /// section edge when testing whether it lies entirely inside an existing hole.
 const HOLE_PROBE_SAMPLES: usize = 8;
 
+/// Parameter `t` in `(0,1)` along segment `a0->a1` where it crosses segment
+/// `b0->b1` in 2D, for a crossing strictly interior to `a` and within (or at
+/// the ends of) `b`. `None` if parallel or out of range.
+fn seg_cross_param(a0: Point2, a1: Point2, b0: Point2, b1: Point2) -> Option<f64> {
+    let (rx, ry) = (a1.x() - a0.x(), a1.y() - a0.y());
+    let (sx, sy) = (b1.x() - b0.x(), b1.y() - b0.y());
+    let denom = rx.mul_add(sy, -(ry * sx));
+    if denom.abs() < 1e-12 {
+        return None;
+    }
+    let (qx, qy) = (b0.x() - a0.x(), b0.y() - a0.y());
+    let t = qx.mul_add(sy, -(qy * sx)) / denom;
+    let u = qx.mul_add(ry, -(qy * rx)) / denom;
+    (t > 1e-6 && t < 1.0 - 1e-6 && u > -1e-6 && u < 1.0 + 1e-6).then_some(t)
+}
+
+/// Weave hole boundaries into the section arrangement of a planar face.
+///
+/// When a holed planar face is cut by sections (e.g. a shelled box top with a
+/// cavity opening, fused with a lip whose walls cross that opening), the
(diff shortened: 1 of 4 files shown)


andymai:
Thanks for the review — both findings addressed in `cf10f0b`:

- **Section dedup quantization (`mod.rs`):** now quantizes at `1/tol.linear` instead of a fixed `1e4` grid, so it only collapses genuinely-coincident sections (a doubly-recorded interference) and won't merge distinct splitters that sit close together on a small/precision-sensitive model. d4's duplicate sections are exactly coincident, so they still dedup; verified 5× plus full suite.
- **`seg_cross_param` epsilons:** the parallel test now compares `|r × s|` to `|r|·|s|` (a relative angle threshold) rather than a fixed `1e-12`, making near-parallel rejection scale-independent. The `t`/`u` checks stay at `1e-6` since those are normalized [0,1] parameters (already scale-invariant fractions of each segment) — noted in a comment.

Full regression (algo/operations/io/wasm) green after the change.

PR #795, opened 2026-06-13
fix(boolean): restrict analytic FF curves + merge coincident junction edges

andymai:
## Summary

Two robustness improvements for **coincident curved-surface fuses**, surfaced while isolating the curved-loft socket family (`fuse_shelled_box_with_socket_loft`).

### 1. `algo/phase_ff` — restrict analytic FF curves to both faces' extent
`compute_raw_curves` intersects the **unbounded** surfaces, so a plane-analytic / algebraic ellipse (e.g. a cylinder vs a tilted wall) can reach far past the faces' shared region and **slit the partner face's wire**, leaving phantom inner wires. The new `restrict_curves_to_faces` builds a lightweight per-face extent (planar boundary polygon, or analytic `v`-range) and drops non-Line curves whose longest contiguous in-both sample run is under two samples — a tangency/point that never splits either face. Conservative: if either extent can't be built, curves pass through unchanged.

### 2. `boolean` — merge coincident duplicate junction edges
A coincident-junction fuse annihilates the shared cap but leaves each argument carrying its **own** copy of the junction-wire edges, differing by sub-micron numerical noise (independent construction). The tight-tolerance vertex merge leaves them distinct → each used once → **free edges** that open the shell. `unify_coincident_boundary_edges` snaps vertices at a looser tolerance and rebuilds every wire against a canonical-edge map keyed by *unordered endpoints + curve type + geometric midpoint* (so a line and an arc between the same endpoints stay distinct); degenerate edges are dropped. **Gated on free-edge results** so clean booleans keep exact topology.

## Effect
Together these take the rrect-cap + frustum reproducer's raw GFA output from **euler=−7 / 8 phantom holes / 30+ free edges** down to **euler=−2 with only the 4 arc-vs-chord corner mismatches** left.

## Known residual (not addressed here)
The 4 remaining corner mismatches are a **loft limitation**: `loft` samples each profile to a polygon and builds straight (`Line`) ring edges with planar side faces, so a rounded-rect loft becomes an *octagonal* frustum whose chord-cornered cap can't match a box's arc-cornered cap. Tracked by the new ignored `fuse_coincident_rrect_cap_with_frustum` test. Closing it needs a **curve-preserving loft**, not a change in the fuse pipeline.

## Testing
Zero regressions: **algo 113, operations 662, io 156, wasm 210**. Clippy + boundaries clean. 5× flake gate on the boolean suite stable.

## ⚠️ Review note
`phase_ff` is **core GFA (pave_filler)** and runs on every analytic FF pair → **HIGH-risk**. Not set to auto-merge; requesting human review before merge.

### DIFF
--- a/crates/algo/src/pave_filler/phase_ff.rs
+++ b/crates/algo/src/pave_filler/phase_ff.rs
@@ -185,6 +185,18 @@ pub fn perform(
                 })
                 .collect();
 
+            // Restrict surface-surface intersection curves to the region that
+            // lies inside BOTH faces. `compute_raw_curves` works on the
+            // unbounded surfaces; the analytic-analytic marcher already bounds
+            // its curves, but the plane-analytic and algebraic paths return
+            // the full surface-surface curve. A cylinder/cone or
+            // cylinder/tilted-plane ellipse that meets the other face only
+            // along a shared cap then reaches far past it and slits the
+            // partner face's wire. Keep only the in-both span (curves only).
+            let raw_curves = restrict_curves_to_faces(
+                topo, fa, fb, surf_a, surf_b, v_range_a, v_range_b, raw_curves, tol,
+            );
+
             for raw in raw_curves {
                 let mut raw = raw;
                 // Closed Circle3D sections — produced by plane-sphere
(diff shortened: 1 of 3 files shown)


andymai:
Addressed the Copilot review in 872af0f:

- **Arc endpoint sampling** (`phase_ff` boundary polygon): now samples `0..=16` so the curve segment's `t1` endpoint is included — no artificial closing chord.
- **Inner wires / holes**: `FaceExtent::Plane` now builds hole polygons from the face's inner wires and `contains()` returns `false` for points strictly inside a hole (beyond the boundary margin), so a curve through a hole is no longer treated as inside the trimmed face.
- **Drop-threshold wording**: clarified the doc + inline comment — `b1 - b0 < 2` keeps runs of ≥2 segments (≥3 in-both samples); the text now says "fewer than two segments" rather than "under two samples".
- **`changed` always-true bug**: the unify pass now reuses the original edge when its endpoints didn't move, only allocating (and flagging a change) on a genuine merge / degenerate drop / vertex snap. It's now a true no-op on a free-edge shell that has no coincident duplicates, instead of rebuilding the whole outer shell every time.
- **Silently-dropped error**: the unify call now logs at debug level on error (still best-effort, doesn't abort the boolean).

Zero regressions (algo 113, operations 662), clippy clean, flake-stable.

andymai:
Thanks — addressed the Greptile review:

**Issue 1 (inner-wire silent loss)** — fixed in 286a09b. `unify_coincident_boundary_edges` now bails with `Ok(false)` when an inner wire fails to rebuild, consistent with the outer-wire path, so a hole can't be silently dropped (which would change topology *and* remove free edges, undetectable downstream).

**Issue 2 (drop-threshold comment)** — already corrected in 872af0f: the doc/comment now say the filter drops runs spanning "fewer than two segments (at most two consecutive in-both samples)", matching `b1 - b0 < 2`.

**Issue 3 (hardcoded `1e-20` in `point_to_polygon_dist`)** — keeping this one as-is. The `1e-20` is a div-by-zero guard for a *degenerate (zero-length) polygon segment*, not a tolerance decision. Substituting `tol.linear²` (≈1e-14) would treat legitimately short ~1e-7-length boundary segments as degenerate points and skip them in the nearest-point search, which is less correct for this purpose. The guard intentionally only excludes truly-zero-length segments.

Test job green on the prior commit; CI re-running on 286a09b.

PR #797, opened 2026-06-14
feat(loft): preserve curved corners for two-profile lofts

andymai:
## Summary

`loft` sampled every profile to a polygon and built straight `Line` ring edges with planar side faces, so a **rounded-rect loft became an octagonal frustum**. A coincident-cap fuse of such a faceted frustum against an arc-cornered solid could then never annihilate the cap cleanly — the long-open **`fuse_coincident_rrect_cap_with_frustum` (#790-residual)** case.

This adds a **curve-preserving path for two-profile lofts**. When both profiles share the same boundary edge structure (same edge count, matching per-edge curve type) and wind CCW, the loft preserves each edge's curve: ring edges keep their `EdgeCurve`, arc corners become ruled side faces, and the caps are the real arc-edged profile wires. Mirrors the reference `BRepFill_Generator`:

- **Coaxial** corner arcs (centers differ only along the axis) → exact **`Cylinder`** (equal radii) / **`Cone`** (unequal radii) — analytic surfaces the boolean handles robustly.
- Genuinely **non-coaxial** corner arcs → a degree-(1,2) **ruled NURBS**.
- `Line` edges keep planar quad side faces. Orientation is set per face against the outward direction.

## Scope (deliberately conservative)

Gated to **two profiles** (a single ruled band). Multi-section curve lofts (e.g. a 5-section gridfinity lip) chain curved bands whose downstream cut/fuse/shell sequences hit near-degenerate corner-junction robustness gaps — those **fall back to the existing polygon loft**, which already handles them. So gridfinity and every other multi-profile loft are **unchanged**.

The deeper multi-section curved-corner boolean robustness (cut → fixed via cone/cylinder recognition on a WIP branch; fuse → isolated to near-degenerate tiny-corner junctions) is tracked for a future effort.

## Testing

- ✅ Closes `fuse_coincident_rrect_cap_with_frustum` (now passing, was `#[ignore]`d).
- ✅ Adds `loft_rounded_rect_preserves_arc_corners` (10 faces, 4 NURBS arc corners, watertight, euler=2).
- **Zero regressions**: operations 664, algo 113, io 156, wasm 211 (gridfinity 24). Clippy + layer boundaries clean.

## Review note

Contained to the loft (operations L3), gated to the two-profile single-band case. It does feed curved-corner solids into the boolean, but only the well-conditioned single-band geometry the FF-trimming + edge-merge from #795 already handle. Not set to auto-merge — flagging for review given the history.

### DIFF
--- a/crates/operations/src/boolean/tests.rs
+++ b/crates/operations/src/boolean/tests.rs
@@ -3050,24 +3050,15 @@ fn fuse_shelled_box_with_socket_loft() {
 }
 
 /// Coincident rounded-rect cap fuse where one solid is a tapered frustum
-/// (loft). A well-posed isolation of the curved-loft socket family
-/// (`fuse_shelled_box_with_socket_loft`): box A and frustum B are built from
-/// the *same* rounded-rect arc profile at z=0, so their caps should annihilate
-/// and the corners join cleanly.
+/// (loft). Box A and frustum B are built from the *same* rounded-rect arc
+/// profile at z=0, so their caps annihilate and the corners join cleanly.
 ///
-/// Blocked by a loft limitation, not the boolean: `loft` samples each profile
-/// to a polygon (`face_polygon`) and builds every ring edge as `EdgeCurve::Line`
-/// with planar side faces — so B is an *octagonal* frustum (straight chord
-/// corners), while A keeps its true arc corners. B's chord-cornered z=0 cap can
-/// never match A's arc-cornered cap, leaving the corners non-manifold. The fix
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the Copilot review in 16031c5: `try_loft_matching_curved_profiles` now pre-computes every side-face surface (step 3b) **before** creating any vertices/edges/faces, so the rare `ruled_arc_surface` fallback returns `Ok(None)` without leaving orphaned topology behind. Build + clippy clean, reproducer + loft tests still pass.

PR #799, opened 2026-06-14
fix(algo): sample concave face interiors via point-in-polygon (thin-shell fuse)

andymai:
## Summary

Fuse of a thin L-shell with the solid that contains it overshot its volume by exactly half the inner cube (`vol_fused ≈ 1.5×vol(A)`). Ground-truth per-face decomposition traced the surplus to a **concave face-interior sampling bug** in `sample_face_interior` (the boolean classification sample point), not to same-domain detection, selection, or the classifier truth table — all of which were verified correct for this case.

## Root cause (pinned with full per-face instrumentation)

`Fuse(diff, inter)` where `A=0.5001³`, `B=0.5³` corner-nested, `diff=Cut(A,B)` (thin L-shell), `inter=B` (cube):

1. `diff`'s three `d=0` origin faces are **0.0001-thin L-frames** wrapping the cube corner.
2. `sample_face_interior` offsets from a boundary-edge midpoint, then flips toward the **boundary-vertex centroid** to pick the inward side. For a concave L-frame the centroid lands **in the notch** (the removed cube corner — outside the face), so the flip pushed the sample *into the opposing cube*.
3. Two of the three symmetric L-frames (`x=0`, `z=0`) were sampled inside the cube → classified **Inside** → dropped by the Fuse truth table; `y=0` happened to land in the strip → Outside → kept. (Asymmetric purely by which edge/centroid geometry each face had.)
4. The two dropped faces left an **open shell** → GFA result rejected by the manifold/Euler gate → **mesh boolean fallback**.
5. The mesh fuse double-counted the coincident `d=0.5` interface → 3 diagonally-halved triangles → **+0.0625**.

The GFA result's *volume was already correct* (the dropped faces are at `d=0`, zero volume contribution); only its manifoldness was broken, which triggered the lossy fallback.

## Fix

For planar faces, replace the centroid flip with a **point-in-polygon verify**: project the boundary, then pick the offset sign — shrinking the magnitude for strips thinner than the offset — that lands strictly inside the polygon. 

- **Convex faces are unaffected**: the first candidate (current behavior's direction) is already interior, so it returns immediately with the same point.
- Only faces whose sample previously fell **outside their own boundary** (always a bug) change.
- Non-planar faces keep the existing centroid heuristic.

## Tests

- **Un-ignores** `fuse_thin_shell_with_containing_solid_preserves_larger_volume` (acceptance test; now `vol_fused = 0.1250750 = vol(A)`).
- Adds `sample_face_interior_thin_l_frame_lands_in_strip` unit test (asserts the sample lands in the strip, not the notch).

## Verification

- algo **114** ✓ · operations full suite (incl. 201s boolean suite) **0 failures** ✓ · io **156** ✓ · wasm **211** ✓
- clippy clean · fmt clean · `check-boundaries.sh` valid
- 10/10 deterministic (flake check on `boolean_invariants`)

## Risk

**HIGH** — `sample_face_interior` is on every boolean's classification path (core GFA). Mitigated by: convex-case behavior is identical, the change only affects samples that previously fell outside their face, and full algo/operations/io/wasm regression is green. Flagging for human review rather than auto-merge.

### DIFF
--- a/crates/algo/src/builder/mod.rs
+++ b/crates/algo/src/builder/mod.rs
@@ -447,16 +447,56 @@ fn sample_face_interior(
     let inward = tangent.cross(face_normal);
     let inward_len = inward.length();
 
-    let mut offset = if inward_len > 1e-12 {
+    let base_offset = if inward_len > 1e-12 {
         inward * (offset_scale / inward_len)
     } else {
         // Degenerate — use a tiny offset along the face normal instead
         face_normal * offset_scale
     };
 
-    // The tangent-cross-normal direction assumes CCW winding; reversed or
-    // CW-wound faces flip it, sending the sample outside the face. Use the
-    // boundary vertex centroid to pick the side that points into the face.
+    // For a planar face, verify the sample lands strictly inside the
+    // (possibly concave) boundary polygon. The tangent×normal sign is
+    // unreliable on inner/notch edges and reversed winding, and a
+    // boundary-vertex centroid can fall in a concavity — e.g. the notch of an
+    // L-shaped face left by a corner cut — so a centroid-based flip points the
+    // sample OUTSIDE the face and into the opposing solid, misclassifying a
+    // thin sliver. Project the boundary, then pick the offset sign (shrinking
+    // the magnitude for strips thinner than the offset) that lands inside.
+    if let brepkit_topology::face::FaceSurface::Plane { normal, .. } = surface {
+        if inward_len > 1e-12 {
+            let mut poly = Vec::with_capacity(edges.len());
+            for oe in edges {
+                let e = topo.edge(oe.edge())?;
+                poly.push(topo.vertex(oe.oriented_start(e))?.point());
+            }
+            let frame = plane_frame::PlaneFrame::from_plane_face(*normal, &poly);
+            let poly2d: Vec<_> = poly.iter().map(|p| frame.project(*p)).collect();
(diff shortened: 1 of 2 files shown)


andymai:
Addressed all review feedback in `da66abe`:

**Copilot — single-closed-curve planar faces (<3 vertices):** The point-in-polygon path is now guarded on `poly.len() >= 3`. A planar disc bounded by one circle/ellipse edge keeps the centroid path (its disc center is interior), restoring the prior behavior for circular caps and avoiding the degenerate `sample_interior_point` boundary-vertex result. Added `sample_face_interior_planar_disc_lands_inside` to lock it.

**Greptile — "robust" fallback can return an exterior centroid:** The near-zero-area fallback now verifies `sample_interior_point` lands inside via `point_in_polygon_2d` before use; if even that is exterior it returns the edge midpoint (on the boundary, never exterior) instead of a possibly-exterior centroid.

**Greptile — halving-schedule rationale:** Documented (24 halvings reach scale ~6e-8 ⇒ min offset ~diag·6e-12, below any physically meaningful strip width).

**Greptile — L-frame test is a proxy check:** Strengthened with a direct `point_in_polygon_2d` interior-membership assertion on the projected sample, in addition to the strip check.

Regression after the changes: algo 115, operations full suite 0 failures, clippy/fmt clean. Still flagged for human review (HIGH-risk core-GFA path), not auto-merged.

PR #803, opened 2026-06-14
fix: cone/torus curved-boolean bugs (volume integration + contained-cut) + parity corpus

andymai:
## Summary

Adds a table-driven boolean **robustness parity corpus** with a gap-tracking scoreboard, and fixes the **two bugs the corpus immediately surfaced** in cone/torus curved booleans. The reference kernel used to derive oracle values is never named in code, comments, or commits — it is used purely as an oracle.

Three independent commits (please **rebase-merge** or merge-commit rather than squash, so release-please sees the `fix(check)` and `fix(boolean)` entries separately):

1. `test(boolean)` — the parity corpus + harness
2. `fix(check)` — cone/torus/sphere volume & area integration
3. `fix(boolean)` — contained cone/torus cut tool-drop

**Parity scoreboard: 52 pass · 6 gap · 0 unexpected failures** (the corpus started at 47/11; 5 cases were promoted by the two fixes).

## The corpus (`crates/operations/tests/parity_*`)

Each case is pure data — two primitives + transforms, an operation, and an exact analytic oracle (area / volume / empty). A **known-gaps allowlist scoreboard** means CI fails only on *unexpected* regressions, while every tracked gap reports its live deviation and a gap that starts passing is flagged for promotion. Run any file with `-- --nocapture` to see the scoreboard. Covers planar box/box (incl. hollow-cavity cut), empty/degenerate results, and curved cyl/cone/sphere/torus booleans.

### `fix(check)` — curved-surface property integration
`integrate_face` mis-integrated full-revolution and closed/pole analytic faces because the projected UV boundary collapses on the apex/pole/seam, so polygon trimming rejected valid quadrature points: a cone integrated ~5.5× low, a torus to zero, a sphere errored on its poles. `check::{solid_volume, solid_area, center_of_mass}` were silently wrong for any solid with such a face.

- Winding-number full-revolution detection (robust to boundary discretization); the winding **sign** selects the pole for a hemisphere cap so the two halves don't both integrate the whole sphere
- Degenerate boundary polygon → integrate the analytic full domain (torus)
- `face_uv_bounds` falls back to the full domain instead of erroring on a degenerate projection (sphere poles)
- Composite Gauss quadrature (≤π/4 patches) so doubly-periodic surfaces converge

Adds `curved_properties` regression tests — the existing `check` tests only exercised the closed-form formulas, never the numerical integrator. This fix also promoted 3 corpus cases (cone/torus/sphere containment-intersect were correct all along, only mis-measured).

### `fix(boolean)` — contained cut tool-drop ⚠️ core engine
`Cut(blank, tool)` with the tool strictly inside the blank should return the blank with a tool-shaped cavity, but GFA's no-intersection assembly placed the tool faces in the *outer* shell (malformed) and the wrapper stripped them — so a contained cone or torus tool was silently dropped and the result was the unchanged blank. (Cylinder/sphere only survived via the mesh fallback.)

Fix: in the containment shortcut, detect a strictly-contained simple tool (blank has an analytic classifier; every tool vertex classifies strictly `Inside`) and assemble the hollow directly — copy the blank, attach a reversed deep-copy of the tool's outer shell as an inner cavity shell. Geometrically exact for any tool shape; also replaces the cylinder/sphere mesh fallback with a clean analytic hollow. Falls through to GFA when the tool touches the boundary or the assembled result fails validation.

## Validation

No regressions. Green: operations lib (664), every boolean integration suite (`boolean_invariants` 22, `boolean_edge_cases` 18, `boolean_stress` 57, `concentric_spheres`, `coaxial_*`, `coincident_planes`, `regress_hexwall_cuts`, `scale_dependent`, `golden_regression`, proptest), `check` (44), `algo` (115). `cargo fmt`, `cargo clippy --all-targets -D warnings`, and `scripts/check-boundaries.sh` all clean.

## Remaining gaps (separate bugs, not addressed here)
- sphere/box **partial overlap** — intersect ~46% low, fuse ~2.2% low (sphere-plane intersection accuracy)
- corner-interpenetrating box **fuse** over-counts (11/6 vs 7/4)
- sphere∪box area calibration probe

## Review note
The `fix(boolean)` commit touches the core boolean engine — flagging for human review; please do not auto-merge.

### DIFF
--- a/crates/check/src/properties/face_integrator.rs
+++ b/crates/check/src/properties/face_integrator.rs
@@ -231,9 +231,10 @@ fn face_uv_bounds<S: ParametricSurface>(
     }
 
     if u_min >= u_max || v_min >= v_max {
-        return Err(CheckError::IntegrationFailed(
-            "degenerate UV domain for face".into(),
-        ));
+        // A degenerate projection (e.g. all boundary vertices on a sphere's
+        // pole seam) does not mean an empty face — it means the boundary failed
+        // to bound a sub-region, so the face spans the full analytic domain.
+        return Ok(full_domain);
     }
 
     Ok(((u_min, u_max), (v_min, v_max)))
@@ -382,10 +383,17 @@ fn integrate_parametric<S: ParametricSurface>(
 ) -> FaceContribution {
     let gauss_pts = gauss_legendre_points(gauss_order);
 
-    let u_scale = (u_range.1 - u_range.0) / 2.0;
-    let u_mid = f64::midpoint(u_range.0, u_range.1);
-    let v_scale = (v_range.1 - v_range.0) / 2.0;
-    let v_mid = f64::midpoint(v_range.0, v_range.1);
+    // Composite quadrature: tile the domain into patches no larger than ~PI/4
+    // so one Gauss rule resolves curved and periodic integrands. A single patch
+    // over a torus's full 2*PI period in both u and v under-resolves it (~0.5%
+    // error); several patches per period converge to machine precision.
+    let patch = std::f64::consts::FRAC_PI_4;
+    let nu = (((u_range.1 - u_range.0).abs() / patch).ceil() as usize).max(1);
+    let nv = (((v_range.1 - v_range.0).abs() / patch).ceil() as usize).max(1);
+    let du_patch = (u_range.1 - u_range.0) / nu as f64;
+    let dv_patch = (v_range.1 - v_range.0) / nv as f64;
+    let u_scale = du_patch / 2.0;
+    let v_scale = dv_patch / 2.0;
 
     let mut area = 0.0;
     let mut vol = 0.0;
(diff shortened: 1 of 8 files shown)


andymai:
Thanks for the reviews — addressed in `8eae1ff` (boolean) and `bc11f5e` (check). These also fixed a real CI failure: the original per-face cavity copy left the cavity shell **non-manifold** (a box-in-box cut reported Euler 8 instead of 4). The metric-oracle corpus didn't catch it (per-face volume was still correct); CI's gridfinity Euler check did.

**Greptile P1 — orphaned solid / shared outer shell:** fixed. `build_contained_cut_hollow` no longer creates a second solid. It now mutates the copied blank in place via `solid_mut(...).add_inner_shell(...)` (new `Solid::add_inner_shell`), so there's no extra `Solid` referencing the blank's outer shell.

**Copilot — orphaned intermediate face:** fixed by the same rewrite. Each copied tool face is now reversed **in place** (`set_reversed`) and the copied tool shell is reused directly as the cavity, so no duplicate faces are created. (The rewrite also switched from per-face `copy_face` to a whole-solid `copy_solid`, which is what makes the cavity share edges/vertices and stay manifold.)

**Copilot — integrator patch-count perf:** fixed. Composite quadrature now caps at 16 patches per axis. A linear axis's integrand is low-degree so a bounded count stays exact, and angular axes never exceed 2π (8 patches), well under the cap — so a tall cylinder/cone no longer scales integration cost with model size.

**Greptile P2 — inner-wire vertices in `solid_strictly_inside`:** fixed. The strict-containment check now classifies inner-wire (hole) vertices in addition to the outer wire.

**Greptile P2 — `oe.oriented_start(e)` traversal:** not applied. `OrientedEdge` has no `oriented_start`/`oriented_end` method in this codebase, and the established convention (including the sibling `all_vertices_inside_or_on` helper) iterates `[e.start(), e.end()]`. The double-visit is harmless for a pure containment check, so I kept it consistent with the existing helper.

Also added a cavity-topology regression test asserting `Euler(result) == Euler(blank) + Euler(tool)` across cone/cylinder/sphere/box/torus, to cover the topological invariant the metric oracles can't.

PR #817, opened 2026-06-15
feat(wasm): add binary tessellateSolidGrouped (packed buffers, no JSON)

andymai:
## Summary

Adds `tessellateSolidGroupedBinary` — a binary counterpart to `tessellateSolidGrouped` that returns packed typed-array buffers instead of a JSON string — and refreshes the README performance table with re-measured numbers.

## Why

The grouped tessellation crossed the WASM boundary as a **JSON string**: every float was serialized to text in WASM, `JSON.parse`d back in JS, then re-packed into typed arrays. For a sphere at tol 0.01 (**56k triangles → a 4.2 MB JSON string**) that round-trip added **~46 ms** on top of ~35 ms of actual tessellation — roughly **doubling** the cost of meshing in brepjs.

Surfaced while re-running the brepjs kernel-comparison benchmarks: `mesh sphere` (81 ms) and `cut` (106 ms) were dominated by marshaling, and the native engine numbers (~33 ms) proved the cost was the JSON round-trip, not the engine.

## Change

- New `JsGroupedMesh` (positions/normals as `Float32Array`, indices/`faceOffsets` as `Uint32Array`) — the same packed-buffer pattern `tessellateSolid` → `JsMesh` already uses. `f32` matches what mesh consumers (GPU buffers) use and halves the transfer vs `f64`.
- New `tessellateSolidGroupedBinary` binding returning it; identical geometry to the JSON variant.
- The JSON `tessellateSolidGrouped` is **kept** for backward compatibility — consumers feature-detect the binary one.
- README performance table re-measured (kernel-comparison suite + native criterion) and the comparison baseline refreshed.

## Result (measured in brepjs after wiring the adapter to the binary path)

| op | before | after |
|---|---|---|
| mesh sphere (56k tris) | 81 ms | **33 ms** (2.4×) |
| cut(box,cyl) ×10 | 106 ms | **62 ms** (1.7×) |
| multi-boolean (16) | 13 ms | **7 ms** (1.8×) |

Speeds up real brepjs meshing of any complex model. (Companion brepjs adapter change uses the new binding with a JSON fallback.)

Verified: `cargo build` / `clippy` / wasm tests clean; the binary binding round-trips correct data (Float32Array positions, Uint32Array faceOffsets) in a node smoke test and the full brepjs kernel-comparison suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,18 +139,18 @@ Median times from the [brepjs benchmark suite](https://github.com/andymai/brepjs
 
 | Operation                | brepkit (WASM) | OCCT (WASM) | Speedup | brepkit (native) |
 | ------------------------ | -------------- | ----------- | ------- | ---------------- |
-| fuse(box, box)           | 5.7 ms         | 83.7 ms     | 15x     | 336 µs           |
-| cut(box, cylinder)       | 4.2 ms         | 123.8 ms    | 29x     | 221 µs           |
-| intersect(box, sphere)   | 31.9 ms        | 107.1 ms    | 3.4x    | 1.8 ms           |
-| box + chamfer            | 0.1 ms         | 7.8 ms      | 78x     | 55 µs            |
-| box + fillet             | 0.3 ms         | 8.1 ms      | 27x     | 75 µs            |
-| multi-boolean (16 holes) | 1.7 ms         | 52.0 ms     | 31x     | 1.2 ms           |
-| mesh sphere (tol=0.01)   | 20.0 ms        | 61.3 ms     | 3.1x    | 1.8 ms           |
-| exportSTEP (×10)         | 0.9 ms         | 19.2 ms     | 21x     | —                |
+| fuse(box, box)           | 0.6 ms         | 45.1 ms     | 75x     | 117 µs           |
+| cut(box, cylinder)       | 61.4 ms        | 71.2 ms     | 1.2x    | 24.8 ms          |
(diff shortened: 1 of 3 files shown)


andymai:
Thanks for the review — addressed in 3e0987a:

- ✅ **Move instead of clone** — `JsGroupedMesh::new` now takes `TriangleMesh` by value and moves `indices`; the call site passes by value.
- ✅ **Unit test for the binding's conversion** — covers f64→f32 packing of positions/normals, the verbatim index move, and the `faceOffsets` sentinel (`last() == indices.len()`). Runs on the host target, no wasm runtime needed.
- ✅ **brepjs type declarations + sync script** — handled in the companion PR (andymai/brepjs#1377): `tessellateSolidGroupedBinary` is registered in `sync-brepkit-types.ts` and the types are forward-declared.

**On `packedBuffer()`** — declining for now. The brepjs adapter consumes meshes through the individual typed-array getters and does **not** use `JsMesh::packedBuffer()` either (it reads `.positions`/`.normals`/`.indices` directly). Each `Float32Array`/`Uint32Array` getter already marshals as a single bulk memcpy, so the four getters move the same byte volume as one packed buffer — the only saving would be ~3 FFI call overheads, traded against JS-side header parsing and re-slicing. If the adapter ever moves to a packed-buffer convention wholesale, `JsGroupedMesh` should follow suit then.

Issue #813, opened 2026-06-14
Fillet robustness: variable-radius fillet returns wrong volume; double fillet fails across NURBS blend faces


andymai:
## Summary

Two fillet code paths produce wrong or failing results compared to OCCT/occt-wasm. Surfaced via brepjs's cross-kernel parity gate, where both are currently gated off for brepkit (`tests/helpers/kernelDivergences.ts`).

### 1. Variable-radius fillet returns physically impossible volume
- Divergence keys: `modifierFns.variableFilletRadius`, `modifierFns.variableFilletCallback`, plus the whole-suite `variableFillet`.
- A variable-radius fillet on a box yields `vol > 1000` where the input solid is far smaller. A fillet removes material, so the result volume must be **less** than the base solid. occt-wasm computes this correctly.
- Capability flag `variableFillet` is currently `false` for brepkit in brepjs.

### 2. Double fillet fails across NURBS blend faces
- Divergence key: `kernelCall.doubleFillet`.
- Filleting edges that are adjacent to a previously-created NURBS blend face fails. This is the standard "fillet a fillet" / sequential-fillet workflow.

## Acceptance

- Variable-radius fillet returns a valid solid with `vol(result) < vol(base)` for a material-removing fillet, matching occt-wasm within tolerance.
- Sequential fillet of an edge adjacent to an existing blend face succeeds.

## Context

Found while validating the brepkit-wasm 2.104.1 bump in brepjs (andymai/brepjs#1372). Per brepjs's "brepkit not held to parity" policy these don't block the bump, but they are genuine correctness gaps. occt-wasm is the reference.

andymai:
Diagnosed both cases against current `main` (post-2.105.0). They have very different status.

## Case 1 — variable-radius fillet volume: **cannot reproduce; appears resolved**

brepkit's variable fillet produces correct, material-removing, manifold volumes for the exact failing config (`box(10,10,10)`, the `[1,2]` → `Linear{start:1,end:2}` law):

| config | result vol | base | manifold |
|---|---|---|---|
| single edge, Linear{1,2} | 991–994 (all 12 edges) | 1000 | ✓ |
| two edges, Linear{1,2} | 982–990 | 1000 | ✓ |

Both the Rust `fillet_variable` and the `filletVariable` WASM binding (which maps `[1,2]` → `Linear{1,2}` correctly) give `vol < 1000`. I could not produce the reported `vol > 1000`.

[...]

PR #821, opened 2026-06-15
fix(fillet): skip edges bordering NURBS blend faces instead of emitting garbage (#813)

andymai:
## Summary

Addresses the **contained safety half of #813 Part 2** (double fillet across NURBS blend faces). Full diagnosis is in [the issue thread](https://github.com/andymai/brepkit/issues/813#issuecomment-4703904492).

### The bug

The walking (`fillet_v2`) and rolling-ball fillet engines build their blend section against *analytic* neighbor surfaces. An edge bordering a **NURBS face** — e.g. a previous fillet's blend face (the "fillet a fillet" workflow) — is unsupported, and **neither engine detects it**:

| second fillet on a plane↔NURBS-blend edge | before |
|---|---|
| `fillet_v2` | `Ok`, volume **unchanged** — silent **no-op** |
| `fillet_rolling_ball` | `Ok`, volume **grows to 1000.30 > 1000 base** — self-intersecting solid |

The garbage result is topologically *manifold*, so a watertightness post-condition can't catch it.

### The fix

Add `operations::query::filter_filletable_edges` — manifold edges (exactly 2 neighbors) whose **both** neighbors are analytic — and pre-filter in the wasm `try_fillet` best-effort path. We fillet only the analytic-neighbor edges and skip the rest (skip-and-continue, matching the binding's existing skip policy and brepjs's `fillet(allEdges)` expectation). If no edges qualify, the solid is returned unchanged.

Result: the double fillet returns a **valid material-removing (or no-op) manifold solid** instead of garbage.

### Scope

This stops the *silent wrong result*. Actually rounding an edge against a NURBS blend face ("fillet a fillet") is a separate, larger blend-engine feature (generalize the section solver to NURBS neighbor surfaces) — tracked in #813 Part 2.

Also confirmed in the same diagnosis: **#813 Part 1 (variable-radius fillet volume) does not reproduce** on current main (correct ~991–994 volumes) — likely a stale brepjs divergence key.

### Tests

- `query::filter_filletable_edges` unit tests: box → all 12 edges filletable; post-fillet → NURBS-blend-neighbor edges excluded, plane↔plane edges retained.
- `try_fillet_second_pass_does_not_break_solid`: a second fillet on blend-adjacent edges never adds material and stays manifold. **Confirmed to fail pre-fix** at `second=1000.30`, pass post-fix.

Refs #813

### DIFF
--- a/crates/operations/src/query.rs
+++ b/crates/operations/src/query.rs
@@ -49,3 +49,100 @@ pub fn filter_planar_edges(
     }
     Ok(result)
 }
+
+/// Filter edges to only those the blend engines can fillet: manifold edges
+/// (shared by exactly two faces) whose **both** neighbor faces are analytic
+/// (plane/cylinder/cone/sphere/torus, not NURBS).
+///
+/// The walking and rolling-ball fillet engines build the blend section against
+/// analytic neighbor surfaces. An edge bordering a NURBS face (e.g. a previous
+/// fillet's blend face) is unsupported — attempting it yields a silent no-op or
+/// a self-intersecting solid — so callers should skip such edges rather than
+/// feed them to the engine.
+///
+/// # Errors
+///
+/// Returns `OperationsError::Topology` if any entity ID is invalid.
+pub fn filter_filletable_edges(
+    topo: &Topology,
+    solid_id: SolidId,
+    edge_ids: &[EdgeId],
+) -> Result<Vec<EdgeId>, OperationsError> {
+    let solid_data = topo.solid(solid_id)?;
+    let shell = topo.shell(solid_data.outer_shell())?;
+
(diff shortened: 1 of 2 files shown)


andymai:
The latest review pass landed on the first commit (`c92ba6e`); all three points are already fixed in `be22cdf`:

- **Inner-wire edges (P1):** `filter_filletable_edges` now walks `outer_wire()` *and* `inner_wires()`, the same adjacency the fillet engine uses.
- **Seam double-count:** neighbors are collected into a `HashSet<FaceId>`, so an edge listed twice in one face's wire counts as a single distinct face; the manifold test is `distinct == 2`.
- **Fragile test ordering (P2):** the regression now fillets *every* result edge and asserts the no-material-gain + manifold invariant on each (and asserts the first fillet actually produced NURBS blend faces), instead of assuming `edge[0..2]`.

Issue #815, opened 2026-06-14
Geometry ops missing vs occt-wasm: projection/HLR, convex hull, Minkowski sum, interference detection, shape-evolution tracking


andymai:
## Summary

Five higher-level geometry/analysis operations that occt-wasm provides are not implemented in brepkit. All are gated off in brepjs's parity gate (`tests/helpers/kernelDivergences.ts`) and/or via capability flags in `tests/helpers/kernelRegistry.ts`.

## Cases

1. **Projection / HLR (hidden-line removal)** — capability flag `projection: false` for brepkit (`true` for occt-wasm). No way to project a 3D solid to 2D visible/hidden edges. Drives brepjs's `projection/` module.

2. **Convex hull** — divergence key `hullFns`. No convex hull of a point set / shape.

3. **Minkowski sum** — divergence key `minkowskiFns`. No Minkowski sum (used for offsetting/clearance shapes).

4. **Interference detection** — divergence key `interferenceFns`. No solid-solid interference/clash query (do two solids overlap, and by how much).

5. **Shape-evolution tracking** — divergence keys `shapeRef.*`, `shapeRefIntegration.*` (~10 tests). After a boolean/fillet, there is no way to ask which output faces were *generated* / *modified* / *deleted* from a given input face (OCCT `BRepAlgoAPI` / `BRepBuilderAPI` `Generated()` / `Modified()` / `IsDeleted()`). This underpins persistent face references across a modeling history.

## Acceptance

Each op returns results matching occt-wasm within tolerance (or, for shape-evolution, exposes generated/modified/deleted reporting on boolean and fillet operations).

## Context

Found while validating the brepkit-wasm 2.104.1 bump in brepjs (andymai/brepjs#1372). occt-wasm is the reference. Shape-evolution tracking is likely the highest-value item: it unblocks robust feature references in downstream history-based modeling.

andymai:
Audited all five items against current `main`. **Three of the five are already implemented in brepkit** — the brepjs divergence keys are stale (same pattern as #816 items 1–2 and #813 Part 1). Two are genuine gaps.

## Already in brepkit — clear the stale keys / wire the adapter

### Convex hull (`hullFns`) — ✅ exists
- `convexHull(coords)` WASM binding (`bindings/shapes.rs`) → `operations::primitives::make_convex_hull` → `math::convex_hull::convex_hull_3d` (incremental Quickhull). Returns a solid.
[...]

PR #822, opened 2026-06-15
feat(wasm): add filletWithEvolution face-provenance tracking (#815)

andymai:
## Summary

Closes the one tractable gap from the #815 audit: **shape-evolution for fillets**. brepjs's `shapeRef` suite tracks face provenance through booleans (the existing `fuseWithEvolution`/`cutWithEvolution`/`intersectWithEvolution`) *and* fillets (`filletWithEvolution`) — but brepkit only exposed the boolean half. This adds the fillet half.

(Per the [#815 audit](https://github.com/andymai/brepkit/issues/815#issuecomment-4704000927), the other items — convex hull, interference, boolean evolution — already exist; Minkowski sum and HLR remain large separate features.)

## Approach

The boolean already builds its `EvolutionMap` by **geometric face-signature matching** (face normal + centroid), not builder provenance — so it's directly reusable:

1. Extract that matcher into `evolution::build_evolution_by_geometry(inputs, outputs) -> EvolutionMap` (operation-agnostic, no topology deps): near-tied input → `modified`, unmatched output → `generated`, unmatched input → `deleted`.
2. Make `boolean::collect_face_signatures` public and refactor `boolean_with_evolution` onto the shared matcher — **behavior-preserving** (the 97 boolean tests stay green; net −95 lines there).
3. Add the `filletWithEvolution` binding: snapshot signatures around `try_fillet`, return `{"solid":<u32>,"evolution":{modified,generated,deleted}}` — the exact JSON shape `fuseWithEvolution` returns — so brepjs's adapter wires without representation loss.

## Tests

- `matcher_classifies_modified_generated_deleted` — synthetic signatures exercise all three buckets.
- `fillet_evolution_tracks_all_faces` — a real box fillet: 6 → 7 faces, no box face deleted, and **every** output face (including the blend) is attributed to an input, so a downstream face reference always resolves. (The blend face is attributed to both edge-adjacent faces — sound provenance under the generous heuristic.)

Refs #815

### DIFF
--- a/crates/operations/src/boolean/mod.rs
+++ b/crates/operations/src/boolean/mod.rs
@@ -829,8 +829,6 @@ pub fn boolean_with_evolution(
     a: SolidId,
     b: SolidId,
 ) -> Result<(SolidId, crate::evolution::EvolutionMap), crate::OperationsError> {
-    use crate::evolution::EvolutionMap;
-
     // Collect input face normals + centroids before the operation mutates topology.
     let input_faces_a = collect_face_signatures(topo, a)?;
     let input_faces_b = collect_face_signatures(topo, b)?;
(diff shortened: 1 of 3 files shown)


andymai:
Addressed in the latest commit:

- **P1 — missing `catch_unwind` (Greptile/Copilot):** ✅ Fixed. `filletWithEvolution` now wraps the fillet + evolution build in `catch_unwind(AssertUnwindSafe(...))` exactly like the `fillet` binding, so a fillet panic returns an error instead of aborting the WASM instance.
- **`collect_face_signatures` "outward normal" doc (Copilot):** ✅ Fixed. Reworded to "face normal" — the stored plane normal (not re-oriented by `reversed`); matching is consistent because input and output use the same convention.
- **`matched_inputs` marks generating inputs as matched (Copilot):** This is verbatim-extracted from the existing `boolean_with_evolution` — behavior-preserving by design (the 97 boolean tests stay green). It's the established boolean-evolution semantics; changing it would alter boolean behavior and belongs in a separate change if revisited.
- **P2 — not wired in the brepjs adapter (Greptile):** Expected — `filletWithEvolution` is wired on the brepjs side (its `shapeRef` module already imports it), like the existing `fuseWithEvolution`. No brepkit-side action.

Issue #814, opened 2026-06-14
Advanced sweeps: multi-profile sweep and guided sweep (auxiliary spine) not implemented


andymai:
## Summary

Two sweep variants that occt-wasm supports are missing in brepkit. Both are gated off in brepjs's parity gate (`tests/helpers/kernelDivergences.ts`).

### 1. Multi-profile sweep
- Divergence key (whole suite): `multiSweepFns`.
- Sweeping a pipe through **multiple section profiles** along a spine (lofted sweep, OCCT `BRepOffsetAPI_MakePipeShell` with several `Add(profile)` sections). brepkit supports single-profile sweep only.

### 2. Guided sweep (auxiliary spine)
- Divergence key (whole suite): `guidedSweepFns`.
- Sweep driven by an **auxiliary spine** that controls profile orientation/scaling along the path (OCCT `SetMode(auxiliarySpine, ...)`).

## Acceptance

- A pipe swept through 2+ distinct profiles produces a solid interpolating between them, matching occt-wasm bounds/volume within tolerance.
- A guided sweep with an auxiliary spine orients/scales the profile per the spine, matching occt-wasm.

## Context

Found while validating the brepkit-wasm 2.104.1 bump in brepjs (andymai/brepjs#1372). occt-wasm is the reference implementation.

andymai:
Audited against current `main`. Unlike #815/#816, **both #814 items are genuine gaps** — but it's worth recording the substantial sweep surface that *does* exist, since these build on it.

## Existing sweep capability (single-profile)
- `sweep` (profile along a path), `sweepSmooth`, `sweepAlongEdges`, `pipe`, `helicalSweep`
- `sweepWithOptions(profile, …)` — single profile with **scaling along the path** (`scale_values`) and contact/orientation modes (`rmf` / `fixed` / `constantNormal:x,y,z`)

So path-following, twist, scale, and frame control exist — but always for **one** profile section.

## Genuine gaps

### Multi-profile sweep (`multiSweepFns`)
No path accepts **multiple section profiles** to interpolate between along a single spine (the `MakePipeShell` + multiple `Add(section)` model). Closest primitives: `sweepWithOptions` (one profile, scaled) and `loft` (multiple profiles but **no spine path** — straight interpolation). Neither covers a multi-section *swept* solid. Real feature.

### Guided sweep / auxiliary spine (`guidedSweepFns`)
[...]

PR #825, opened 2026-06-15
feat(sweep): native multi-section sweep with RMF frame transport (#814)

andymai:
## Summary

Implements **multi-profile sweep** (#814, the `multiSweepFns` gap): sweep through multiple section profiles along a spine, interpolating between them — the lofted-pipe workflow.

### Context from the audit

brepjs already *composes* this in JS — `multiSectionSweep` = position each profile along the spine (`positionOnCurve`) + `loft`, both of which reduce to existing brepkit primitives (`transformWire`, `loftWithOptions`). So the basic case works today. **But** that JS `positionOnCurve` uses a single swing-rotation with no rotation-minimizing frame, so it **twists profiles on curved spines** (the tests only exercise line spines).

This PR adds a **first-class native operation** that's more robust and a single call.

### Approach

`operations::sweep::multi_section_sweep(spine, sections, ruled)`:
- Places each planar profile at its spine parameter using the sweep engine's existing **rotation-minimizing frame** (`compute_frames`, double-reflection) — twist-free on curved paths.
- The placement maps the profile's centroid → spine point, normal → tangent, and plane → the frame's right/up (a rigid `Mat4` via `copy_face` + `transform_face`).
- Lofts the placed profiles — `loft` (ruled) or `loft_smooth`.

Exposed as the `multiSectionSweep` WASM binding (spine as raw NURBS data, parallel `face_handles`/`params` arrays, `ruled` flag), mirroring the existing `sweep` binding's shape.

### Why native vs. leaving it to JS composition

- **RMF** — profiles stay perpendicular and twist-free on curved spines; the JS swing-rotation does not.
- One kernel call instead of N positioning calls + a loft.
- A foundation for guided sweep (#814's second half).

### Tests

- `multi_section_sweep_line_spine_tapered_volume` — frustum between 10×10 and 6×6 squares; volume strictly between the two prism bounds.
- `multi_section_sweep_curved_spine_positive_volume` — 3 sections on a quarter-circle spine; valid positive volume.
- `multi_section_sweep_rejects_single_section`, `…_rejects_out_of_range_param`.
- WASM contract test: lofts two circles along a line spine → positive volume.

(Validation error paths can't be exercised at the WASM boundary off-wasm because `JsError` can't be constructed there — they're covered at the operations layer.)

Refs #814 (multi-profile half). Guided sweep / auxiliary spine remains the second half.

### DIFF
--- a/crates/operations/src/sweep.rs
+++ b/crates/operations/src/sweep.rs
@@ -5,6 +5,7 @@
 //! point. Uses rotation-minimizing frames (double-reflection method) to avoid
 //! Frenet-frame singularities on straight segments and inflection points.
 
+use brepkit_math::mat::Mat4;
 use brepkit_math::nurbs::curve::NurbsCurve;
 use brepkit_math::nurbs::surface_fitting::interpolate_surface;
 use brepkit_math::tolerance::Tolerance;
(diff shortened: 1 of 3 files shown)


andymai:
Addressed in the latest commit (14a38de):

- **det = −1 reflection (P1):** `profile_to_frame_matrix` now builds a left-handed profile basis (`p_y = p_x × normal`) to match the `(right, up, tangent)` frame, so R is a proper rotation. Added `profile_to_frame_matrix_is_proper_rotation` asserting `det = +1` (fails pre-fix).
- **`multiSectionSweep` missing from `dispatch_op` (P1):** registered in `batch.rs` (spine passed by NURBS edge, matching the `sweep` arm) + a batch contract test.
- **Centroid breaks for single-closed-edge profiles + `f64 == 0.0` (P2 / Copilot):** centroid now comes from the sampled boundary (`face_polygon` + `polygon_centroid`), which is correct for circle profiles and removes the float-equality count check.
- **Sections not ordered (P2):** sections are sorted by parameter before lofting; added `multi_section_sweep_unsorted_params_match_sorted`.
- **Raw / unnormalized plane normal (Copilot):** the normal is normalized before building the basis. The face's `reversed` flag doesn't affect the result — `loft` applies its own CCW profile correction.

PR #827, opened 2026-06-15
feat(sweep): guided (two-rail) sweep via auxiliary spine (#814)

andymai:
## Summary

Implements the **guided-sweep / auxiliary-spine** half of #814 (the `guidedSweepFns` gap). A guided sweep orients the profile so its up-vector points toward a *guide curve* at each path parameter — the profile rolls to track the guide rather than holding a fixed or rotation-minimizing orientation.

### Why this is a genuine gap (not stale)

Unlike multi-section sweep (composable from existing primitives), the brepjs adapter's `sweepPipeShell` **silently dropped the `auxiliary` guide** — it only did contact-mode/smooth/simple-pipe fallbacks, so the guide had no effect. This makes the guide actually work.

### Approach — additive, no enum ripple

`SweepContactMode` derives `Copy`, so a `Guided(NurbsCurve)` variant would break `Copy` and ripple to `sweep_miter`. Instead:

- `SweepOptions` gains an optional **`aux_spine: Option<NurbsCurve>`** (the struct already derives `Default` and is non-`Copy`).
- When set, `sweep_with_options` computes **guided frames** — `up = orthogonalize(aux(t) − spine(t), tangent)` — instead of the `contact_mode` frames, reusing the entire existing swept-solid construction (scale law, caps, side faces). When `None`, behaviour is byte-for-byte unchanged.

Exposed as `operations::sweep::sweep_guided` + the `guidedSweep` WASM binding (profile + spine NURBS + aux NURBS) + batch dispatch (spine/aux by NURBS edge).

### Proof that the guide controls orientation

The key test: a wide rectangle (16×2) swept along a straight spine.
- **Plain sweep** (no guide): stays flat — Y-extent **2.0**.
- **Guided** by a curve that rotates +X→+Y over the sweep: the rectangle rolls 90°, sweeping its wide axis into Y — Y-extent **16.0**.

So the auxiliary spine demonstrably rolls the profile (it is not ignored).

### Tests

- `sweep_guided_produces_valid_solid`, `sweep_guided_rotating_aux_rolls_profile` (the 2→16 orientation proof).
- WASM `guided_sweep_produces_solid` (direct) + `guided_sweep_batch_dispatch`.

Refs #814 — this completes both halves (multi-section sweep landed in #825).

### DIFF
--- a/crates/operations/src/sweep.rs
+++ b/crates/operations/src/sweep.rs
@@ -1116,21 +1116,32 @@ pub fn sweep_with_options(
         // Guided (two-rail) sweep: orient the profile so its up-vector points
         // toward the auxiliary spine at each path parameter. The tangent still
         // follows the main path, so this overrides `contact_mode`.
-        (0..=num_segments)
-            .map(|k| {
-                #[allow(clippy::cast_precision_loss)]
-                let t = k as f64 / num_segments as f64;
-                let origin = path.evaluate(t);
-                let tangent = path.tangent(t).unwrap_or(path_tangent_0);
-                let up = orthogonalize(aux.evaluate(t) - origin, tangent);
-                Frame {
-                    origin,
-                    tangent,
-                    up,
-                    right: tangent.cross(up),
-                }
-            })
-            .collect()
+        let mut out = Vec::with_capacity(num_segments + 1);
+        let mut prev_up: Option<Vec3> = None;
+        for k in 0..=num_segments {
+            #[allow(clippy::cast_precision_loss)]
+            let t = k as f64 / num_segments as f64;
+            let origin = path.evaluate(t);
+            let tangent = path.tangent(t).unwrap_or(path_tangent_0);
+            let guide_dir = aux.evaluate(t) - origin;
+            // Where the guide momentarily coincides with the spine the up-vector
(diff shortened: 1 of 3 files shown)


andymai:
Addressed in 60049ce:

- **Degenerate guide (P2, Greptile + Copilot):** the guided frame computation is now a stateful pass that carries the previous frame's up-vector when the guide momentarily coincides with the spine (`|aux(t) − spine(t)| < 1e-9`), so orientation stays continuous instead of snapping to an arbitrary world axis. Added `sweep_guided_handles_guide_meeting_spine` (a guide that starts on the spine then diverges → still a valid finite solid).
- **degree validation (P2):** the `guidedSweep` binding now rejects `degree < 1` for both the spine and aux curves.
- **Batch error clarity (Copilot):** the batch `guidedSweep` error now names the offending argument (`spineEdge` vs `auxEdge`).

On the **P1 — "guide still dropped for brepjs JS consumers"**: that's the brepjs adapter (`sweepPipeShell`) not yet calling this new binding — a brepjs-side wiring task, the same pattern as the other audited items (the kernel capability now exists; the adapter needs to use it). No brepkit-side change.

PR #828, opened 2026-06-15
feat(operations): convex Minkowski sum of two solids (#815)

andymai:
## Summary

Implements **Minkowski sum** (#815, the `minkowskiFns` gap): `make_minkowski_sum(A, B)` = the convex hull of all pairwise vertex sums.

### Context

The approach is the standard convex-Minkowski identity (`A ⊕ B = hull{a + b}`), exact when both inputs are convex — boxes, or a shape with a tessellated sphere as the rolling tool. It was already implicitly validated by brepkit's `convex_hull_minkowski_*` tests; this wraps it into a first-class operation + binding. (brepjs's adapter also routes a *sphere* tool to `offset` — that path already works via `offset_solid`; this covers the general convex case.)

### Tests

- `minkowski_sum_unit_boxes_volume_8` — `[0,1]³ ⊕ [0,1]³ = [0,2]³` (volume 8).
- `minkowski_sum_box10_box2_volume_1728` — `box(10) ⊕ box(2) = box(12)` (volume 1728, matching the brepjs bar).
- WASM `minkowskiSum` direct + batch contract tests.

Refs #815 (Minkowski). Remaining #815 gap: HLR/projection.

### DIFF
--- a/crates/operations/src/primitives.rs
+++ b/crates/operations/src/primitives.rs
@@ -802,6 +802,46 @@ pub fn make_convex_hull(
     Ok(topo.add_solid(solid))
 }
 
+/// Convex Minkowski sum of two solids: `A ⊕ B = { a + b | a ∈ A, b ∈ B }`.
+///
+/// Computed as the convex hull of all pairwise vertex sums, which is exact when
+/// both inputs are convex (e.g. boxes, or a shape with a tessellated sphere as
+/// the rolling tool) and a convex over-approximation otherwise.
+///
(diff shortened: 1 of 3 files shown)


andymai:
Follow-up #832 addresses the review findings (docstring overstated 'exact' → convex *polytopes*; `saturating_mul` capacity hint; `# Errors` widened; test comment no longer assumes box origin). Apologies — this auto-merged before the review was handled.

PR #830, opened 2026-06-15
feat(operations): edge projection with hidden-line removal (#815)

andymai:
## Summary

Implements **HLR / projection** (#815, the last #815 gap): project a solid's edges to a view plane and classify each segment **visible** or **hidden**.

### Why it's a genuine gap

The brepkit adapter's `projectEdges` was a stub ("return all edges as visible outlines, no hidden line removal"). This adds real occlusion.

### Approach (no tessellation in the occlusion test)

`operations::projection::project_edges`:
- Samples each edge to a polyline (the projected drawing is inherently polygonal).
- Classifies each segment by an **exact point-in-solid test**: a boundary sample is hidden iff stepping it toward the camera (`p − ε·dir`) lands `Inside` the solid (a face is in front). Reuses `classify_point` (ray-casting) — no face tessellation.
- Merges consecutive same-visibility segments into polylines and projects via the camera frame `(u,v) = ((p−origin)·x, (p−origin)·y)`.

Exposed as the `projectEdges` WASM binding (returns `{visible, hidden}` flat 2D polylines) + batch dispatch.

### Tests

- `project_box_oblique_view_has_visible_and_hidden_edges` — an oblique (1,1,1) view of a box yields **both** visible and hidden edges (the three far-corner edges are occluded).
- `project_box_without_hidden_lines_drops_hidden`.
- Batch contract test asserting non-empty visible + hidden.

### Known limitation

A view aligned *exactly* to faces (e.g. an exact front view of a box) is degenerate — back edges graze coplanar faces, which point-classification can't disambiguate (OCCT resolves it by convention). The general oblique case is exact; a silhouette-aware refinement could handle the coincident case later.

Refs #815 (projection/HLR). This closes the #815 implementation gaps; convex hull / interference / boolean-evolution were already present (documented on the issue).

### DIFF
--- a/crates/operations/src/lib.rs
+++ b/crates/operations/src/lib.rs
@@ -27,6 +27,7 @@ pub mod helix;
 pub mod loft;
 pub mod pipe;
 pub mod primitives;
+pub mod projection;
 pub mod revolve;
 pub mod sweep;
 
--- a/crates/operations/src/projection.rs
+++ b/crates/operations/src/projection.rs
@@ -1,0 +1,176 @@
+//! Edge projection with hidden-line removal (HLR).
+//!
+//! Projects a solid's edges onto a view plane and splits each edge into visible
+//! and hidden polylines. Occlusion is an **exact** point-in-solid test — no
+//! tessellation of faces: a boundary point is hidden when stepping it toward the
+//! camera enters the solid (a face is in front of it). Edges themselves are
+//! sampled to polylines because a projected drawing is inherently polygonal.
+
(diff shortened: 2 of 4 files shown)


andymai:
Follow-up #832 addresses the review findings, including both P1s: `projectEdges` now validates all nine float inputs are finite, and `project_edges` propagates classification errors instead of silently treating them as visible. Plus the epsilon clamp and the deflection/tolerance doc fix. Apologies — this auto-merged before the review was handled.

Issue #834, opened 2026-06-15
Fillet an edge adjacent to a NURBS blend face ("fillet a fillet")


andymai:
## Summary

Round an edge whose neighbour is a **NURBS blend face** — the sequential "fillet a fillet" workflow. Today the fillet engine cleanly **skips** such edges (safety fix #821, so it never returns garbage), but it doesn't yet round them.

Split out of #813 (whose other parts — the stale variable-radius-volume report and the safety fix — are resolved/closed).

## Reproduce

```
box(10,10,10) → fillet edges[0..2] @ r=1   // creates 2 NURBS blend faces (vol ≈ 994, manifold)
             → fillet a result edge bordering a blend face @ r=0.5
```
Edges between two planar faces fillet fine; an edge between a planar face and a NURBS blend face is skipped.

## Root cause — three independent layers

This was mis-diagnosed earlier as a walker/NURBS-convergence issue. It is **not** — the walker structurally supports NURBS neighbours (`surface_ref_or_adapter` maps `FaceSurface::Nurbs` to a `ParametricSurface`). The real chain:

1. **Unshared edges.** The first fillet produces a result that is *geometrically* manifold but *topologically* uses unshared edge IDs (the blend face and the trimmed planar neighbour reference different edges for the same boundary). So `adjacency.faces_for_edge(edge).len() != 2` → `compute_stripe_for_edge` returns `StartSolutionFailure` at `crates/blend/src/fillet_builder.rs:308`, **before** the walker. All target edges fail → the builder returns the unchanged solid (the silent no-op).
2. **Sew correctness.** `sew::sew_faces` on the first result *does* fix the sharing (the edge then has 2 distinct neighbours), but the sewn solid is wrong (volume 1041 vs ~990 input / 1000 base) — sewing doesn't cleanly reconstruct a fillet result.
3. **Blend-surface build.** Once sharing is fixed and the walker runs on the plane↔NURBS edge, `approximate_blend_surface` fails `InvalidKnotVector { expected: 24, got: 6 }` — the NURBS-neighbour contact curve has a different knot structure than the analytic path assumes.

Same family as the `d5` filleted-lip watertightness work.

## Fix path

- A watertight **edge-reconciliation** pass in `fillet_builder` so blend-face boundaries are shared with the trimmed neighbours (so a later fillet's adjacency sees 2 faces). Reconcile only truly-coincident edges; the `d5` notes flag a curve-vs-chord case that can't be canonicalised directly.
[...]

PR #842, opened 2026-06-15
fix(fillet): watertight rolling-ball fillet of two edges sharing a corner

### LINKED ISSUE #841: Fillet: two edges sharing a corner produce a non-watertight junction
## Summary

Filleting two edges that **share a corner vertex** in a single operation produces a **non-watertight** solid — the junction where the two fillet strips meet leaves 4 free edges (euler = 1 instead of 2). Filleting two edges that do **not** share a corner is watertight, and single-edge fillets are watertight, so the defect is localized to the **multi-fillet corner junction**.

This is the remaining blocker for the literal 2-edge repro in #834: a non-watertight 2-edge fillet can't serve as a watertight input to a subsequent "fillet a fillet" operation. (The single-edge → blend-adjacent workflow of #834 is fully fixed by #837 + #839.)

## Reproduce

`fillet_rolling_ball` on a 10×10×10 box, varying which two edges are filleted together:

| edges | share a corner? | F | E | V | euler | free edges | watertight |
|------|----|---|---|---|-------|-----------|------------|
| `[0], [1]` | **yes** | 8 | 19 | 12 | **1** | **4** | ✗ |
| `[0], [2]` | no | 8 | 18 | 12 | 2 | 0 | ✓ |
| `[0], [3]` | **yes** | 8 | 19 | 12 | **1** | **4** | ✗ |
| `[1], [2]` | **yes** | 8 | 19 | 12 | **1** | **4** | ✗ |

```rust
let cube = make_box(&mut topo, 10.0, 10.0, 10.0)?;
let edges = solid_edges(&topo, cube);
// edges[0] and edges[1] share a corner:
let s = fillet_rolling_ball(&mut topo, cube, &[edges[0], edges[1]], 1.0)?;
// validate_shell_closed(shell, &topo) -> Err (4 free edges, euler = 1)
```

The non-watertightness appears **iff the two filleted edges meet at a shared vertex**. Volume is sane (≈994), and the result is "manifold" in the weak sense (no edge used by >2 faces), so `validate_shell_manifold` passes but `validate_shell_closed` fails.

## Suspected mechanism

At a shared-corner junction the rolling-ball engine sets each strip back from the vertex and builds a spherical-triangle corner patch (Phase 5b in `crates/operations/src/fillet/rolling_ball.rs`). The 4 free edges suggest the corner patch boundary and/or the setback-trimmed strip ends and the trimmed neighbour-face corners are **not stitched into a shared boundary** — `assemble_solid_mixed` dedups by position, so the leftover free edges indicate the patch/strip/trim endpoints don't coincide (or the patch doesn't fully close the gap) at the junction.

Next step: dump the 4 free edges' positions for the `[0],[1]` case and check them against the corner-patch boundary, the two strip setback-ends, and the trimmed-face corner vertices to see which boundary isn't shared.

## Acceptance

- Filleting N edges meeting at a shared vertex produces a valid **watertight** manifold (`validate_shell_closed` passes, euler = 2), matching the non-shared-corner and single-edge cases.
- Unblocks the 2-edge variant of #834.

## Refs

- #834 (single-edge → blend-adjacent fillet, fixed in #837 + #839)
- #837 (curved-neighbour contact geometry), #839 (consumer-path wiring)
- Engine: `crates/operations/src/fillet/rolling_ball.rs` (Phase 5b corner patches, setback map, `assemble_solid_mixed`).

andymai:
## Summary

Closes #841. Filleting two edges that **share a corner vertex** in a single rolling-ball pass produced a **non-watertight** solid (4 free edges, `euler = 1`). This unblocks the literal 2-edge "fillet a fillet" repro from #834.

## Root cause

At a shared-corner junction the engine sets both strips back from the vertex (`setback_map`), expecting a corner patch to fill the gap. But two things were missing for the **2-edge** case:

1. **Phase 5b's corner patch was gated at `fillet_count >= 3`** — so for exactly two edges nothing closed the gap.
2. **Phase 3's side-face trim collapsed the unfilleted third edge onto the far corner** (e.g. it connected the setback contact straight to `(0,0,10)`) instead of preserving the unfilleted edge near the corner.

The result was a 4-free-edge loop A–B–D–C (A = the far end of the still-sharp unfilleted edge).

## Fix

`crates/operations/src/fillet/rolling_ball.rs`:

- **Phase 3** — at a 2-edge setback corner, preserve the unfilleted edge by emitting a trim point `P` on it (recorded in `corner_preserved`). Both side faces compute the same `P` from the unfilleted edge's far endpoint, so the sub-edge survives as a shared boundary. Gated on `setback_map` containing the filleted edge → single-edge and G1-chain fillets are untouched (no setback there).
- **Phase 5b** — gate lowered to `>= 2`; a new branch builds a four-sided corner patch `P–B–D–C` (`build_two_edge_corner_patch`): the two `near→far` boundaries are sphere arcs matching the strip ends, the two `P→near` boundaries are straight, matching the trimmed side faces. The 3-edge spherical-triangle path is unchanged (the patch's third arc would face the still-sharp unfilleted edge with nothing to share it, so the eighth-sphere isn't applicable here).

The change only fires for `fillet_count == 2` corners with a setback; the 3-edge, single-edge, and G1-chain paths are provably unaffected.

## Verification

- The repro is now **watertight**: 9 faces, 0 free edges, `validate_shell_closed` passes, `euler = 2`, volume ≈ 994.4 (matches the issue's ~994 and the non-shared-corner case).
- New regression tests: `rolling_ball_two_edges_shared_corner_watertight` (6 corner-sharing pairs: closed + manifold + genus-0 + sane volume) and `rolling_ball_two_edges_no_shared_corner_watertight` (no spurious patch).
- Full `brepkit-operations` suite: all pass (690 lib + integration, 0 failed).
- `brepkit-wasm` native suite: 224 passed, 0 failed.
- `cargo clippy --all-targets --all-features` clean; `check-boundaries.sh` clean.

## Risk

HIGH — core fillet engine (`rolling_ball.rs`). Tightly gated and regression-clean, but flagging for human review before merge.

### DIFF
--- a/crates/operations/src/fillet/rolling_ball.rs
+++ b/crates/operations/src/fillet/rolling_ball.rs
@@ -978,7 +978,7 @@ pub fn fillet_rolling_ball(
                             if let Ok(dir) = (next_pos - pos).normalize() {
                                 let p = pos + dir * radius;
                                 trimmed_verts.push(p);
-                                corner_preserved.insert(vi, p);
+                                corner_preserved.entry(vi).or_insert(p);
                             }
                         }
                     }
@@ -990,7 +990,7 @@ pub fn fillet_rolling_ball(
                             if let Ok(dir) = (prev_pos - pos).normalize() {
                                 let p = pos + dir * radius;
                                 trimmed_verts.push(p);
-                                corner_preserved.insert(vi, p);
+                                corner_preserved.entry(vi).or_insert(p);
                             }
                         }
                         if let Some(&pt) = fillet_contact_map.get(&(vi, ei, fi)) {
@@ -1123,7 +1123,7 @@ pub fn fillet_rolling_ball(
                         if let Ok(dir) = (next_pos - pos).normalize() {
                             let p = pos + dir * radius;
                             new_verts.push(p);
-                            corner_preserved.insert(vi, p);
+                            corner_preserved.entry(vi).or_insert(p);
                         }
                     }
                 }
@@ -1136,7 +1136,7 @@ pub fn fillet_rolling_ball(
                         if let Ok(dir) = (prev_pos - pos).normalize() {
                             let p = pos + dir * radius;
                             new_verts.push(p);
-                            corner_preserved.insert(vi, p);
+                            corner_preserved.entry(vi).or_insert(p);
                         }
                     }
                     if let Some(&pt) = fillet_contact_map.get(&(vi, ei, fi)) {
@@ -1528,6 +1528,7 @@ pub fn fillet_rolling_ball(
         // corners does not apply here because one of its arcs would face the
         // (still sharp) unfilleted edge with nothing to share it.
         if fillet_count == 2 {
+            let mut built = false;
             if let (Some(&p_pt), Some(v_pos)) = (corner_preserved.get(&vi), original_vertex) {
                 if blend_points.len() == 3 {
                     // Sphere centre: corner offset inward by R along each distinct
@@ -1583,8 +1584,21 @@ pub fn fillet_rolling_ball(
                         is_concave,
                     ) {
                         all_specs.push(spec);
+                        built = true;
                     }
                 }
+            }
+            if !built {
+                // The setback gap could not be closed (no preserved point on the
+                // unfilleted edge, contacts that didn't deduplicate to a triangle,
+                // or a degenerate patch). Surface it — the junction may be left
+                // non-watertight rather than failing silently.
+                log::warn!(
+                    "2-edge fillet corner at vertex {vi}: corner patch not built \
+                     (preserved={}, unique_contacts={}); junction may be non-watertight",
+                    corner_preserved.contains_key(&vi),
+                    blend_points.len()
+                );
             }
             continue;
         }
--- a/crates/operations/src/fillet/rolling_ball.rs
+++ b/crates/operations/src/fillet/rolling_ball.rs
@@ -825,6 +825,16 @@ pub fn fillet_rolling_ball(
     // Phase 3: Build modified (trimmed) planar faces.
     let mut all_specs: Vec<FaceSpec> = Vec::new();
 
+    // At a corner where exactly two filleted edges meet (sharing one face), the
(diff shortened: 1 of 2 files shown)


andymai:
Thanks for the review — addressed all four points in 2387972:

1. **Concave 2-edge corners** (correctness): extracted the concavity test the 3-edge path uses (summed outward face normals vs summed edge tangents) into a `corner_is_concave` helper, and the 2-edge branch now flips both the sphere-centre offset sign and the patch's outward-orientation reference. Convex corners are unchanged — `offset_sign == -1` reproduces the previous `v_pos - r·Σn` formula bit-for-bit, and all existing tests still pass.
2. **Phase 5b header comment**: updated to "2+ edges" and now describes the four-sided 2-edge patch alongside the 3-edge eighth-sphere triangle.
3. **"bottom-ring" test comment**: corrected — the test collects all vertex-sharing edge pairs.
4. **"distinct corner orientations" comment**: reworded to say it samples the first several pairs in iteration order (capped for speed), without overstating coverage.

andymai:
Thanks @greptile — addressed all three in 7c508d4:

- **P1 — `corner_preserved` overwrite (both paths):** switched `insert()` → `entry().or_insert()` in the planar and non-planar trim branches, so the first computed `P` wins deterministically and matches the first emitted face polygon (same convention already used for `fillet_contact_map` at G1 junctions). Box corners are unaffected — the single unfilleted edge gives both side faces the same `P`. (For a corner with 2+ *unfilleted* edges a single patch still can't match every differing `P`; that case now surfaces via the P2 warning rather than silently — see below. Fully general multi-unfilleted-edge corners are out of scope for this fix.)
- **P2 — silent no-op:** the `fillet_count == 2` branch now emits a `log::warn!` when the corner patch can't be built (no preserved point, contacts that didn't deduplicate to a triangle, or a degenerate patch), reporting `preserved` and `unique_contacts` instead of leaving the junction non-watertight with no diagnostic.

CI green on the prior commit; re-running on 7c508d4. Still flagged for human review before merge (core fillet engine).

PR #865, opened 2026-06-15
feat(topology): add make_ellipse_arc trimmed-ellipse-arc constructor + wasm export

### LINKED ISSUE #858: Ellipse-arc 2D→3D lift is geometrically wrong (sketcher3d ellipseTo/halfEllipseTo)
brepjs parity gap: `sketcher3d.ellipseTo` / `sketcher3d.halfEllipseTo` (divergence registry). Extruded volumes are far off — `ellipseTo` 24.4 vs **115.5**, `halfEllipseTo` 33.3 vs **89.3** — not a precision margin, a wrong arc.

## Diagnosis (this session)
Path: `Sketcher.ellipseTo/halfEllipseTo` → `baseSketcher2d.ellipseTo` → `makeEllipseArcFromSvgParams` (ellipseUtils.ts) → `make2dEllipseArc` → `getKernel().makeEllipseArc2d` ⇒ bk2d `{__bk2d:'trimmed', basis: makeEllipse2d(...), tStart, tEnd}` (`direct=true`, sense always CCW). The lift in `curvesAsEdgesOnPlane` dense-samples ellipse arcs into a NURBS that collapses the swept region.

Two layered root causes:
1. **`reverseCurve2d` is a silent no-op** (`brepjs/src/kernel/brepkit/kernel2dOps.ts:288` — `/* no-op for immutable objects */`). `makeEllipseArcFromSvgParams` does `if (!clockwise) arc.reverse()`, so **non-clockwise SVG arcs are never actually reversed on brepkit** — broader than ellipses (any reversed 2D curve). bk2d objects are `readonly`, so a correct fix needs a reversed-curve representation threaded through `evaluateCurve2d`/`curveBounds`/the lift, or changing the kernel's in-place `reverse` contract to return-value.
2. **Exact-ellipse trim direction** must be reconciled with `sense` when the SVG params swapped start/end (clockwise `ellipseTo` still reconstructs wrong).

## Foundation already built
brepkit branch **`feat/ellipse-arc-3d-primitive`** (pushed, not merged): `topology::builder::make_ellipse_arc` + `makeEllipseArc3d` wasm export (exact trimmed-ellipse edge, unit-tested). Held until the adapter consumer works to avoid merging unused API. The brepjs adapter ellipse branch in `curvesAsEdgesOnPlane` was prototyped (uses bk2d fields cx/cy/majorRadius/minorRadius/xDirAngle/sense) and verified via wasm-swap but still wrong pending the two root causes above.

**Fix touches:** brepjs adapter (`reverseCurve2d`, `curvesAsEdgesOnPlane`) primarily; brepkit primitive PR lands alongside.

andymai:
## Phase A of #858 — the brepkit ellipse-arc primitive

The ellipse-arc 2D→3D lift parity gap (`sketcher3d.ellipseTo`/`halfEllipseTo`) needs an exact elliptical-arc edge so the brepjs adapter can stop dense-sampling ellipse arcs into a NURBS that collapses the swept region. This PR lands the **brepkit-side foundation** — additive and self-contained:

- `topology::builder::make_ellipse_arc` — builds an exact `EdgeCurve::Ellipse` edge trimmed by distinct start/end vertices. The CCW angular range comes from `EdgeCurve::domain_with_endpoints` (which already handles ellipses), so the edge traces exactly the arc. Coincident endpoints ⇒ closed full-`[0, 2π]` edge.
- `makeEllipseArc3d` wasm export — the elliptical analog of the existing `makeCircleArc3d`, same validate-and-delegate shape (`validate_finite`/`validate_positive`, returns a `u32` edge handle). Produces real CIRCLE/ELLIPSE-class geometry, not a NURBS approximation.

## Why now

Extracted as a clean single commit onto current `main` (the brepkit half of the fix), so a release can ship the primitive and the brepjs adapter can consume it. It was previously held back to avoid shipping unused API; landing it is the unblocker for the cross-repo Phase B.

## Tests

- `make_ellipse_arc_traces_quarter_arc` (topology) — trimmed `[0, π/2]` domain; midpoint at `π/4` lands on the ellipse at `(5·√½, 2·√½, 0)`.
- Builds clean for the `wasm32-unknown-unknown` target; the binding mirrors `makeCircleArc3d`, which is itself covered via the gridfinity integration tests.

## Scope / follow-up (not in this PR)

This does **not** by itself close #858. The user-visible parity fix is **Phase B in brepjs** (fix the `reverseCurve2d` no-op and add an exact ellipse-arc branch in `liftCurve2dToPlane` that calls `makeEllipseArc3d`) — a separate, cross-repo change that lands after this releases and brepjs re-syncs types. #858 stays open until that ships.

### DIFF
--- a/crates/topology/src/builder.rs
+++ b/crates/topology/src/builder.rs
@@ -171,6 +171,44 @@ pub fn make_ellipse_edge(
     )
 }
 
+/// Create a trimmed elliptical arc edge between `start` and `end`.
+///
+/// Builds an [`Ellipse3D`] from `center`, `normal`, the semi-axes, and the
+/// `ref_dir` major-axis direction, then an edge whose distinct start/end
+/// vertices trim it. [`EdgeCurve::domain_with_endpoints`] projects those
+/// endpoints onto the ellipse and returns the CCW angular range, so the edge
+/// traces exactly the arc. `start`/`end` must lie on the ellipse. When the
+/// endpoints coincide the edge is closed (full `[0, 2π]` domain).
+///
+/// # Errors
+///
+/// Returns an error if either semi-axis is non-positive, `semi_minor` exceeds
+/// `semi_major`, or `normal`/`ref_dir` is degenerate.
+#[allow(clippy::too_many_arguments)]
+pub fn make_ellipse_arc(
+    topo: &mut Topology,
+    center: Point3,
+    normal: Vec3,
+    semi_major: f64,
+    semi_minor: f64,
+    ref_dir: Vec3,
+    start: Point3,
+    end: Point3,
+    tolerance: f64,
+) -> Result<EdgeId, crate::TopologyError> {
+    let ellipse = Ellipse3D::new_with_ref(center, normal, semi_major, semi_minor, ref_dir)
+        .map_err(|e| crate::TopologyError::NonManifold {
+            reason: format!("invalid ellipse: {e}"),
+        })?;
+    let v_start = topo.add_vertex(Vertex::new(start, tolerance));
+    let v_end = if (start - end).length() < tolerance * 100.0 {
+        v_start
+    } else {
+        topo.add_vertex(Vertex::new(end, tolerance))
+    };
+    Ok(topo.add_edge(Edge::new(v_start, v_end, EdgeCurve::Ellipse(ellipse))))
+}
+
 /// Create a closed wire from an ordered list of points.
 ///
 /// Each consecutive pair of points becomes a line edge, and the last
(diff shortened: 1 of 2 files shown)


andymai:
Re: Greptile — `makeEllipseArc3d` not in `dispatch_op`: intentional, and matching the existing `makeCircleArc3d` precedent. The arc constructors are low-level **edge** primitives meant to be called directly while assembling wires/faces (as in `gridfinity_tests.rs`), not standalone batch operations. Neither `makeCircleArc3d` nor `makeEllipseArc3d` belongs in the JSON batch dispatch table; the batch API operates at the solid/op level. Documenting that as the explicit decision.

Issue #859, opened 2026-06-15
Non-square lip fuse drops the tool in GFA fuse (coincident contact-ring)


andymai:
Uncovered while fixing the non-square sweep overshoot (#854). With the lip now geometrically correct, `fuse(hollow_box, translated_lip)` for a **non-square** (42×126) rounded-rect lip returns the box **unchanged** — `lipAdded = -0.0`, `fusedZMax` stays 21 — while the **square** 84×84 case fuses correctly (lip present, zMax 23.88).

## Diagnosis
The lip's bottom contact-ring is coincident with the box rim. This is the recurring GFA **same-domain / coincident-face** theme (see `project_gfa-samedomain-scope` notes): SD detection/selection drops the coincident tool ring on the asymmetric (non-square) case. The square case is symmetric enough to survive.

This is what still blocks brepjs `gridfinity.rectLipSweep` end-to-end (the sweep half is fixed + released in 2.112.1).

**Risk:** HIGH — core GFA boolean engine (same-domain overlap detection). Warrants careful, isolated work + heavy regression checks. Repro: build the 1×3 lip via the gridfinity lip profile, translate to the box rim, fuse; compare to the 84×84 case.

**Fix touches:** brepkit `algo` (GFA same-domain pass).

andymai:
## Investigation update — the headline root cause needs revising

Re-investigated against current `main`. The **symptom is real and reproduces**, but the proposed root cause (GFA same-domain detection dropping the coincident tool ring) does **not** hold up. Correcting the diagnosis so whoever picks this up starts in the right place.

### What I ruled out
- **Non-square-ness is not the trigger.** Built non-square *planar* box-frame lips natively (10×30, 8×24, 10×14) and fused them onto the rim — all fuse correctly (F=28, Euler 2, lip survives), identical to the square case.
- **Same-domain selection isn't dropping the lip.** `apply_sd_selection` (`algo/src/bop.rs:140`) only ever *discards* coincident faces; there's no keep-then-drop path that could lose an Outside-classified lip. `planar_faces_overlap` (`same_domain.rs:394`) is a containment test and isn't on this path.

### Actual mechanism — arc contact-edge reconciliation + a silent drop gate
The lip's bottom contact ring meets the box rim only at **coincident arc (`Circle`) corner edges**, and edge reconciliation in the assembler is **Line-only**:
[...]

PR #867, opened 2026-06-16
fix(algo): weld near-coincident vertices in solid assembly (#859)

andymai:
## ⚠️ Core GFA assembly change — requesting human review before merge (#859)

Fixes the non-square rounded-rect lip fuse that dropped the tool (blocks gridfinity `rectLipSweep`). **This investigation overturned the originally-posted diagnosis** — see below.

### Corrected root cause (the issue's "same-domain" theory and the triage's "arc reconciliation" theory were both wrong)
I first implemented the triage's proposed fix (generalize `split_edges_at_collinear_vertices` to sub-split `Circle`/`Ellipse` arcs). It compiled but **fired zero times** — gated behind a flag, the whole suite stayed green without it — because the coincident arcs already share endpoints; there was nothing to sub-split.

Dumping the literal free edges of the GFA growth shell (before the mesh fallback) showed the real cause: the pavefiller places a coincident-ring intersection vertex a few ULPs short of the body's exact vertex — e.g. `x = -11.999999` vs `-12.0` (~1e-6 apart). `merge_duplicate_edges` keys edges on `quantize_point(p, MERGE_TOL)` with `MERGE_TOL = 1e-7`, so these near-duplicates land in **different cells** and the two faces' shared boundary is never unified. It stays open → `has_free_edges` → the boolean rejects the GFA result and falls back to the mesh boolean (≈correct volume for the symmetric square case, badly wrong — `5383` vs expected `9656` — for the asymmetric non-square one).

### Fix
New `weld_coincident_vertices` pass in `crates/algo/src/builder/builder_solid.rs`, run as "Step 0a-pre3" before the existing edge-split/merge passes:
- Clusters selected-face vertices by **actual distance** at `snap = 10 · MERGE_TOL = 1e-6` (the tolerance `split_edges_at_collinear_vertices` already treats as coincident) via a coarse spatial hash (cell = snap ⇒ a 26-neighbour probe is provably sufficient).
- Snaps each near-duplicate onto the lowest-`VertexId` canonical vertex, rebuilds touched wires, and drops edges that collapse to a point.
- Deterministic (vertex-index order, fixed probe order, shared-edge cache) so no HashMap iteration order leaks into entity IDs.

### Before / after
| case | before | after |
|---|---|---|
| square (15×15) | manifold=false, free edges, vol 9791 | **manifold, no free edges, vol 9791** (exp 9792) |
| non-square (10×22) | **broken: vol 5383**, manifold=false | **manifold, no free edges, vol 9655** (exp 9656) |

### Verification (re-run independently, not just the agent's report)
- `cargo test -p brepkit-operations` — 684 lib + all integration suites, **0 failed**. Canaries green: `d4_shelled_box_fuse_lip`, `fuse_stacked_rounded_rect_arc_prisms_{same,nested}_footprint`, `fuse_overlapping_rounded_rect_arc_prisms_same_footprint`, `cut_concentric_rounded_rect_arc_prisms_overshoot`, `rounded_rect_arc_prism_volume_baseline`.
- `cargo test -p brepkit-algo` — 115 passed, 0 failed.
- `cargo clippy -p brepkit-algo -p brepkit-operations --all-targets -- -D warnings` — clean.
- `./scripts/check-boundaries.sh` — valid.
- New regression: `fuse_arc_frame_lip_is_manifold_{square,nonsquare}`.

### 🔍 What I'd like a reviewer to scrutinize
1. **This runs on every `build_solid` (every boolean).** It's a no-op (early return) when nothing is within `snap`, and the full suites are green — but it's a new pass in the hottest core path. Worth a read of `weld_coincident_vertices` and a thought on perf for very large solids.
2. **`snap = 1e-6` welding moves vertices** by up to 1e-6 (10× `MERGE_TOL`). That matches the engine's existing on-edge coincidence tolerance, but it is a deliberate sub-tolerance snap.
3. **Greedy clustering is single-linkage / non-transitive** — it can *under*-merge (leave the bug unfixed in a pathological chain) but won't *over*-merge here, since real vertices are either ~1e-6 coincident or millimetres apart. If geometry ever has legitimately-distinct vertices ~1e-6 apart they'd be ambiguous to the kernel regardless.
4. **Alternative considered:** fixing it at the source (pavefiller emitting exact coincident vertices) would be deeper but riskier and broader; the downstream weld is the conservative, localized choice.

### Out of scope
`fuse_shelled_box_with_socket_loft` stays `#[ignore]`d — un-ignoring it shows a *different* defect (non-manifold edge shared by 3 faces, `euler=-54`), not the near-coincident-vertex bug. The issue's brepjs `rectLipSweep` un-skip is a downstream follow-up after this releases.

### DIFF
--- a/crates/algo/src/builder/builder_solid.rs
+++ b/crates/algo/src/builder/builder_solid.rs
@@ -75,6 +75,16 @@ pub fn build_solid(topo: &mut Topology, selected: &[SelectedFace]) -> Result<Sol
     // slivers removed above, so they would survive as free edges).
     remove_zero_length_edges(topo, &mut face_ids)?;
 
+    // Step 0a-pre3: Weld vertices that are coincident within snap tolerance.
+    // Intersection in the pavefiller can place a vertex a few ULPs short of an
+    // exact pre-existing vertex (e.g. a coincident-arc tangent point landing at
+    // -11.999999 vs the body's -12.0). Such near-duplicates quantize to
+    // different cells at MERGE_TOL, so the duplicate-edge merge never unifies
+    // the two faces' partitions and the shared boundary stays open. Snapping
+    // them to one canonical vertex (and dropping the resulting zero-length
+    // slivers) lets the merge below see identical partitions.
+    weld_coincident_vertices(topo, &mut face_ids)?;
+
     // Step 0a: Split Line edges at intermediate collinear vertices.
     // Adjacent faces can partition the same geometric boundary differently
     // (one whole edge vs several sub-edges split at paves); refining every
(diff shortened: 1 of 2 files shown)


andymai:
Addressed review feedback in 1398b7b:

- **Canonical selection (Copilot):** `weld_coincident_vertices` now picks the **lowest-index** canonical within `snap` across all 27 probed cells, instead of breaking on the first match — so the choice no longer depends on cell iteration order and matches the documented rule.
- **Function doc (Greptile):** corrected — an arc whose once-distinct endpoints weld together is **dropped** (not kept), while a genuinely closed input arc is preserved; also noted the pass is O(V log V) and early-returns without rebuilding when nothing welds.
- **Collapse-site comment (Copilot):** clarified that a closed input *arc* is preserved while a zero-length line is always dropped (the `matches!(EdgeCurve::Line)` term).

Re-ran locally: `operations` boolean module (99) + `algo` (115) green, clippy clean.

PR #871, opened 2026-06-16
fix(algo): analytic gridfinity bin — multi-section loft, coaxial cone cut, shelled-lip fuse

andymai:
## What

Makes the full gridfinity **1×1 bin (d4)** analytic and closed-manifold end to end, instead of collapsing to a non-manifold mesh-boolean fallback (was Euler≈−85, ~290 all-planar faces, ~5× volume). The bin's stacking lip exercises three GFA weaknesses in sequence — a multi-section curve-preserving loft, a coaxial-cone cut, and a coincident-ring fuse — each fixed here. They are landed together because the loft fix alone regresses d3/d4 (the now-analytic lip drives the still-broken cut/fuse differently), so the chain must be atomic.

### Loft (analytic lip profiles)
1. Relax the curve-preserving loft's `>2`-profile gate (`loft.rs`) — the section loops already build one ruled band per adjacent pair, so allow N≥2.
2. Merge co-circular corner arcs in `profile_oriented_edges` — `drawRoundedRectangle` emits a 90° corner as one arc on some sections and two co-circular sub-arcs on others (lip sections came in 8/12/12/12 edges), defeating the per-index matcher. Normalize to canonical one-arc-per-corner.

### Cut (coaxial cone cavity)
3. Sample cone/cylinder cavity walls at the axial midpoint for cut classification — the inner cavity wall's interior sample landed exactly on the coincident top rim (a shared boundary), where the analytic classifier returns None and ray-cast says Outside, dropping the wall.
4. Exact algebraic coaxial cone-cone intersection yielding one clean `Circle` (`math`) + emitting it as `EdgeCurve::Circle` so it links to the shared cap boundary (was ~97 degenerate micro-curves from the grid marcher).
5. Robust interior-point + partial-cylinder ray classification; bail the composite classifier to ray-cast for corner-fillet cylinders; trim partial closed ellipse/NURBS FF curves to their in-both arc.

### Fuse (shelled lip → body)
6. Split CW cavity walls and nest holes in `split_face_2d` (`face_splitter`): an inner-shell cavity corner winds CW in UV, so the CW sign-flip must fire independent of loop count (it was gated to `loops.len() <= 1`), else both v-bands of a section-split corner read as holes and collapse to one sub-face — orphaning the wall's top edge. And nested holes from a section-split holed annulus are assigned to the **innermost** containing sub-face via mutual point-in-polygon (robust to arc-wire UV-area under-measurement), not the first container.

## Result (the real gate — wasm gridfinity bin tests, not the ops --lib suite)

- **d4 full 1×1 bin: faces=68, Euler 2 (genus-0, inner_loops=2), volume 10760 mm³, lip present (z→25.4)** — analytic, closed-manifold. Was the mesh fallback.
- d3 shelled box + lip: pass (196 faces). d1/d1a/d1a2/d1a1c/d1b/d2: pass.
- New native repro `fuse_shelled_cup_lip_covers_cavity_opening` (fast proxy for the lip→body fuse).
- `cargo test --workspace`: 2316 passed, 0 failed (51 binaries). clippy `-D warnings`, fmt, boundaries clean.

d5 (filleted lip) remains ignored — the boolean is clean (Euler 2) but the fillet stage re-introduces boundary edges; separate follow-up.

### DIFF
--- a/crates/algo/src/builder/face_splitter/containment.rs
+++ b/crates/algo/src/builder/face_splitter/containment.rs
@@ -34,15 +34,22 @@ pub(super) fn find_point_outside_holes(
             (outer_pts[i].x() + outer_pts[j].x()) * 0.5,
             (outer_pts[i].y() + outer_pts[j].y()) * 0.5,
         );
-        // Nudge the edge midpoint slightly toward the centroid.
-        let candidate = Point2::new(
-            edge_mid.x() * 0.9 + centroid_x * 0.1,
-            edge_mid.y() * 0.9 + centroid_y * 0.1,
-        );
-        if super::super::classify_2d::point_in_polygon_2d(candidate, outer_pts)
-            && !is_inside_any_hole(&candidate, inner_wires)
-        {
-            return candidate;
+        // Step inward from the edge midpoint toward the centroid in small
+        // increments; the first point inside the outer wire and outside every
+        // hole wins. Small steps handle THIN rings (e.g. the ~1.2mm gridfinity
+        // lip annulus on an 83mm cap), where a single large nudge overshoots
+        // straight into the hole and no ring point is ever found.
+        for k in 1..=99 {
+            let t = f64::from(k) * 0.005;
+            let candidate = Point2::new(
+                edge_mid.x() * (1.0 - t) + centroid_x * t,
+                edge_mid.y() * (1.0 - t) + centroid_y * t,
+            );
+            if super::super::classify_2d::point_in_polygon_2d(candidate, outer_pts)
+                && !is_inside_any_hole(&candidate, inner_wires)
+            {
+                return candidate;
+            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


andymai:
Converting to **draft** — holding the merge. The loft fixes here are correct (the gridfinity lip's multi-section lofts now come out analytic), but landing them **alone** regresses two integration tests because the lip's downstream booleans still fall back to mesh on a coincident top cap:

- `gridfinity_d3_shelled_box_with_lip`: fused face count 200 → 372 (analytic lofts tessellate to a finer mesh-fallback in the still-faceting `cut(outer, inner)`).
- `gridfinity_d4_full_1x1_bin`: lip dropped (Z extent 21.0 vs 25.4) — the analytic-loft lip's cut/fuse now throws and shellStage's catch returns the body lip-less.

Root cause downstream: `cut(outer, inner)` and the lip→body fuse hit a **same-domain coincident face** (outer & inner frustums coincide at the top cap) and fall back to a non-manifold mesh boolean. These loft fixes are the prerequisite; they'll land together with the GFA same-domain cut + fuse fixes so the bin reaches analytic Euler-2 and d3/d4 pass.

PR #877, opened 2026-06-17
fix(algo): treat near-collinear wire-builder junctions as continuations

andymai:
## Summary

Fixes the multi-tongue gridfinity baseplate dovetail fuse going non-manifold (~20 free/over-shared edges) when a tongue is fused into a wall face that an earlier tongue already split.

**Stacks on #875** (the dovetail FUSE-hang fix). Base that branch first if reviewing the full series; this PR is diffed against `main` and may need a trivial rebase once #875 merges.

## Root cause (confirmed by native repro, not assumed)

The GFA face splitter's minimum-clockwise-angle wire-builder traversal (`clockwise_angle` in `crates/algo/src/builder/wire_builder.rs`) wove a **zero-area slit** instead of splitting a face when a section edge met two **collinear** boundary edges at a T-junction.

The gridfinity tongue cap is exactly this: a trapezoid whose slanted base edge passes straight through the wall-line section point (the 0.01 mm overlap puts that point exactly on the base line). At the junction the boundary "bar" continues straight and the section "stem" branches perpendicular; the min-clockwise rule must pick the stem (the face split), and a straight continuation is meant to score TAU (worst).

`clockwise_angle` only flagged a straight continuation when the swept angle `da < 1e-14`. But at a collinear junction `da` is pure rounding noise (the angles flow through `atan2`, subtraction, `rem_euclid`) and lands anywhere up to ~1e-12 depending on the vertex coordinates. The **same tongue** split cleanly at `y=9` (`da=1.8e-15`, under the cutoff → TAU) but failed 40 mm away at `y=49` (`da=1.7e-13`, over the cutoff → scored ~0). When it scored ~0 the boundary continuation **beat** the real section branch, so the section never split the cap: the cap stayed whole and overlapped its own sliver, producing 3 edges shared by 3 faces plus 1 free edge per tongue (~20 across the baseplate), which forced the slow mesh-boolean fallback.

This was localized by dumping, per cap face: the section edges, the parent loop, the wire-builder loop set, each loop's area, and finally the traversal's per-junction turn scores — which showed identical UV geometry but a `da` flipping across the 1e-14 boundary between the two tongues.

## Fix

Widen the continuation band in `clockwise_angle` from `1e-14` to `1e-9` rad (and fold the symmetric near-TAU case). Safe by construction:
- At a 2-edge junction the lone continuation is still the minimum and is taken regardless of its score.
- At a 3+-edge junction a numerically-collinear continuation should always lose to a genuine branch — which is exactly the face split the traversal exists to find.

## Tests

- `wire_builder`: `clockwise_angle_treats_near_collinear_as_continuation` (the exact `da=1.7e-13` case — fails under the old 1e-14 cutoff) and `t_junction_with_collinear_bar_splits_into_two` (the minimal 2D form).
- `boolean`: new `baseplate_two_tongues_far_apart_resplit_is_watertight` (tongue A on a clean slab is watertight; tongue B 40 mm away fused into the A-split wall must stay watertight — failed before, passes now), and `baseplate_dovetail_sequential_fuse_no_mesh_blowup` now asserts a closed manifold at **every** step plus a constant per-tongue volume increment.

## Scope note

This fixes the multi-tongue re-split of **plain trapezoidal** tongues. The gridfinity tool's *basic* dovetail tile still falls to mesh because each tongue is additionally **relieved** with a through-cut gridfinity pocket (a 5-section ruled loft with rounded-rect/tapered sections), so the fused operand carries curved + sloped surfaces — a separate, deeper boolean path. Confirmed via 5 native repros (two-tongue, realistic-baseplate perpendicular-edge, slab-with-pocket, grid-of-pockets, box-relieved tongue) that the plain-tongue path is now watertight; the relieved-tongue/curved-boolean fuse is out of scope for this change.

## Gates
- `cargo test -p brepkit-algo` — 120 passed
- `cargo test -p brepkit-operations --lib` — 694 passed
- `cargo test -p brepkit-wasm --lib gridfinity` — 25 passed, 1 ignored (d5, pre-existing)
- `cargo fmt --all --check`, `cargo clippy --all-targets -D warnings`, `./scripts/check-boundaries.sh` — clean

### DIFF
--- a/crates/algo/src/builder/builder_solid.rs
+++ b/crates/algo/src/builder/builder_solid.rs
@@ -439,10 +439,37 @@ fn shell_is_closed(topo: &Topology, faces: &[FaceId]) -> bool {
     !edge_counts.is_empty() && edge_counts.values().all(|&c| c % 2 == 0)
 }
 
+/// Newell's method normal for a polygon (unnormalized; magnitude = 2·area).
+/// Robust to non-planar / non-convex loops.
+fn newell_normal(verts: &[Point3]) -> Vec3 {
+    let n = verts.len();
+    let mut nx = 0.0;
+    let mut ny = 0.0;
+    let mut nz = 0.0;
+    for i in 0..n {
+        let a = verts[i];
+        let b = verts[(i + 1) % n];
+        nx += (a.y() - b.y()) * (a.z() + b.z());
+        ny += (a.z() - b.z()) * (a.x() + b.x());
+        nz += (a.x() - b.x()) * (a.y() + b.y());
+    }
+    Vec3::new(nx, ny, nz)
+}
+
 /// Compute a signed volume estimate for a shell using the divergence theorem.
 ///
 /// Positive = outward-oriented normals (growth shell).
 /// Negative = inward-oriented normals (hole shell).
+///
+/// Each face's fan-triangulation contribution is oriented by the face's actual
+/// geometric surface normal (which already accounts for `is_reversed`), not by
+/// the raw outer-wire winding. The two agree for solids built with a
+/// CCW-against-the-outward-normal convention (e.g. `make_box`), but diverge for
+/// equally valid solids whose wires were wound the other way (e.g. a profile
+/// extruded *opposite* its face normal). Trusting the wire winding alone made
+/// such a solid read as negative volume, so every shell of a fuse that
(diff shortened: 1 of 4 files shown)


andymai:
Superseded by #879 — rebased clean onto main (this branch's #875 base was squash-merged, leaving it conflicting). Same fix, identical diff.

PR #882, opened 2026-06-17
fix(algo): build the enlarged mouth when a cut deepens a blind-pocket notch

andymai:
## Problem

The snap-clip baseplate connector cuts **blind, ledged pockets** into the baseplate's seam-side walls. Where a later cut **deepens** an existing pocket notch, the GFA boolean produced a non-watertight result (free + non-manifold edges), so the exported STL failed the scenario assertions (`nonManifoldEdges==0`, `boundaryEdges==0`).

## Confirmed mechanism (native repro)

A blind-pocket cutter that overhangs a wall (so its opening reaches the wall edge) opens a **concave notch** in that wall face. When a second cut deepens the pocket:

- the **old notch-floor edge** ends up inside the freshly-removed area — removed material on its interior side — yet it's still carried as a face boundary edge;
- the min-clockwise-turn wire builder traces the *shallow* old notch and orphans the deepening sections into zero-area spurs, so the **enlarged notch mouth is never built** → free / non-manifold edges.

Isolated to the planar face splitter with a minimal box-slab repro: a wall notched to z=-2.8, then deepened to z=-4.4, goes from watertight to 10 free + 4 non-manifold edges purely from the second cut. Reproduces for both a flush meeting and a `COPLANAR_OVERLAP`-style overlap; clean when the cuts leave a wall between them.

## Fix

A planar-face pass (`clip_buried_cut_edges`) drops the buried edges before wire building:

- **flush** deepening — `drop_collinear_bridge_edges` drops the old floor edge that is bridged across the notch line at both endpoints (a boundary neighbour one way, a section continuing collinearly the other into the deeper cut);
- **overlapping** deepening — `drop_buried_loop_edges` rebuilds the removed-region polygon from the section edges' own closed loop, then drops the old floor edge enclosed by it and the tool-lid section running through the old notch's air.

Each detector fires only on its specific signature, so ordinary cuts are untouched.

## Scope / not yet fixed

This covers the wall-**NOTCH** deepening case (the cutter opening reaches the wall edge). The wall-**HOLE** deepening case — where the opening is *fully enclosed* in the wall (the gridfinity snap-clip throat, whose retaining-wall gap leaves the lower band as an enclosed window that the chamber cut then deepens) — routes through a different splitter path (`split_face_with_internal_loops`, which forms two overlapping holes instead of merging them) and is **not** fixed here. The full snap-clip baseplate is therefore not yet watertight; that remaining variant is a clean follow-up.

## Tests / gates

- New regression `deepened_blind_notch_in_overhanging_wall_is_watertight` (flush, +0.01 overlap, deep overlap) — passing, no `#[ignore]`.
- `cargo test -p brepkit-algo` (118), `cargo test -p brepkit-operations --lib` (694 + 5 ignored), `cargo test -p brepkit-wasm --lib gridfinity` (25) — all green, zero regressions.
- `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `./scripts/check-boundaries.sh` — clean.

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -17,7 +17,7 @@ use brepkit_topology::Topology;
 use brepkit_topology::edge::EdgeCurve;
 use brepkit_topology::face::{FaceId, FaceSurface};
 
-use super::classify_2d::{sample_interior_point, signed_area_2d};
+use super::classify_2d::{point_in_polygon_2d, sample_interior_point, signed_area_2d};
 use super::pcurve_compute::evaluate_edge_at_t;
 use super::plane_frame::PlaneFrame;
 use super::split_types::{OrientedPCurveEdge, SectionEdge, SplitSubFace, SurfaceInfo};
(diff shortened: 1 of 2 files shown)


andymai:
Superseded by #884 — rebased clean onto current main (#882's base predated #878/#879/#881).

PR #884, opened 2026-06-17
fix(algo): build the enlarged mouth when a cut deepens a blind-pocket notch

andymai:
Rebased-clean replacement for #882 (its base predated #878/#879/#881, leaving it conflicting).

Fixes a class of deepened-blind-pocket cuts going non-manifold (the baseplate snap-clip connector's overhanging cutters). When a later cut deepens an earlier cut's opening on an overhanging wall face, the face splitter left the old opening's floor edge as a face boundary; the min-clockwise-turn wire builder then traced the *shallow* old opening and orphaned the deepening into zero-area spurs → free/non-manifold edges.

Fix: `face_splitter/mod.rs::clip_buried_cut_edges` drops the buried floor edge (and the tool-lid section running through the old opening's air) before wire building, for both flush and `COPLANAR_OVERLAP` deepening — so the enlarged mouth is built. Fully fixes the **wall-NOTCH** variant (opening reaching the wall top edge).

Regression: `deepened_blind_notch_in_overhanging_wall_is_watertight` (flush / +0.01 overlap / deep overlap), passing. Zero regressions across algo + operations + gridfinity.

Known remaining: the **wall-HOLE** variant (enclosed window, which the real snap-clip throat band hits) still needs a 2D polygon-union primitive that doesn't exist yet — a self-contained follow-up. So the snapClip scenario is not yet fully watertight; this is a contained engine correctness improvement toward it.

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -17,7 +17,7 @@ use brepkit_topology::Topology;
 use brepkit_topology::edge::EdgeCurve;
 use brepkit_topology::face::{FaceId, FaceSurface};
 
-use super::classify_2d::{sample_interior_point, signed_area_2d};
+use super::classify_2d::{point_in_polygon_2d, sample_interior_point, signed_area_2d};
 use super::pcurve_compute::evaluate_edge_at_t;
 use super::plane_frame::PlaneFrame;
 use super::split_types::{OrientedPCurveEdge, SectionEdge, SplitSubFace, SurfaceInfo};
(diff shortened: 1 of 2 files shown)


andymai:
Closing: this wall-NOTCH fix regresses inclusion_exclusion_near_coincident_faces (usize underflow at face_splitter/mod.rs:999 — clip_buried_cut_edges false-fires on near-coincident faces and over-removes edges). It's also only a partial snapClip fix (the real snap-clip hits the wall-HOLE variant, which needs a 2D polygon-union primitive brepkit lacks). snapClip deferred to a proper effort: sound deepened-notch detection that doesn't false-positive on near-coincident faces + polygon union for the enclosed-window case.

PR #897, opened 2026-06-17
fix(algo): assemble faceted-ramp arcs on corner cylinders in plane×cylinder FF

andymai:
## Problem

Fusing a faceted scoop-ramp prism into a gridfinity compartmented bin fell back to the mesh boolean (hundreds of tessellated faces, wrong volume) because the result was non-manifold at the bin's rounded **corner cylinders**.

The scoop's sharp staircase (~12 thin tread planes) pokes into each rounded corner cylinder. Each tread plane ∩ cylinder is a sub-millimetre ellipse arc, but the arcs were trimmed independently → endpoints ~0.08 mm apart → they never chained into a split curve → the cylinder stayed whole → its whole-face interior sample landed inside the scoop → the cylinder was dropped entirely → free edges → mesh fallback.

## Fix

**Exact arc assembly (`phase_ff.rs`):** for a `{planar tread} × {cylinder/cone}` section, trim the closed ellipse at the EXACT points where the tread's straight boundary lines and the cylinder's seam edges cross it (analytic line-segment × surface quadratics) instead of uniform-t sampling. Adjacent treads share a boundary crossing, so a fine-grained endpoint registry chains the arcs through one shared vertex into a continuous boundary-to-boundary split curve. A seam-line crossing and the line-cylinder quadratic for the *same* geometric point can disagree by just over 1e-6 at bin coordinates, so the crossing dedup runs at 1e-5 (treads are ~1e-4 apart), keeping the exact-on-seam endpoint so the seam boundary edge splits and the chain connects.

**Band classification (`face_splitter/mod.rs`):**
- A partial cylinder/cone quarter split boundary-to-boundary by a section chain produces two complementary bands of **opposite UV winding**; signed-area wrongly nests the reversed one as a hole. Force both to outer *only* when two seam-carrying loops have opposite-sign effective areas (the flipped-pair signature) — a genuinely nested band/hole winds the *same* way, so a single plane×corner-cylinder lip section is correctly left to signed-area (this is what keeps the gridfinity lip booleans intact).
- The section-arc pcurves on a lateral band can land in a different 2π window than the boundary edges, self-crossing the UV polygon and mislabelling the interior sample. Unwrap the sampled points to one continuous u-window before computing the interior point.

Also: record the partial-arc face's excluded angular u-gap in `FaceExtent`, and scale the planar `FaceExtent` margin to the smaller in-plane extent so a thin full-width tread does not overshoot onto the cylinder seam. (Both verified load-bearing for this fix.)

## Verification

New regression test `crates/io/tests/scoop_corner_cylinder_fuse.rs` (STEP fixtures from the tool) drives the **public** `operations::boolean(Fuse)` and asserts the result is watertight (free=0, over=0), keeps all 8 analytic corner cylinders, and stays small (<100 faces — no mesh fallback). It fails on `main` (28 free edges → fallback) and passes here.

Zero regressions:
- `brepkit-algo` 126/126
- `brepkit-operations` 696 + all integration binaries, 0 failures (incl. parity corpus, same-domain #895)
- `brepkit-io` 131 + 25 + `multicavity_cut` (#891) + the new scoop test
- gridfinity 26/26 ×3 (deterministic)
- check / heal / offset green
- fmt, clippy `-D warnings`, layer boundaries all pass


andymai:
Superseded by #905, which cherry-picked this branch's corner-cylinder fix (d3afdb7) as part 1 of a coordinated scoop-fuse fix and added the section-over-extension, dangling-chain, and NURBS-wall-recognition fixes on top. The faithful repro now reaches free=0 via #905. Closing — its content is in main.

PR #898, opened 2026-06-17
fix(algo): trim notch-corner sections to the true arc + opposing face

andymai:
## Summary

A rounded-rect **notch cut whose top corners straddle the wall-top edge** went non-manifold and the operations boolean fell back to a mesh result. This was a cascade of analytic-trimming gaps, not one bug. This PR lands four regression-free fixes that take the isolated single-wall straddle cut from **free=16/euler=1 → free=0/euler=2, fully analytic** (all corner cylinders preserved).

## Root causes fixed

1. **Arc boundary edges clipped as chords.** `clip_line_to_face_boundary` collected boundary edges from their start/end vertices, so a section exiting a convex rounded corner clipped to the chord (x=±12) instead of the true arc (x=±13.24). Fix: intersect the section with the true `Circle3D` arc and **merge** that crossing into the chord crossings, taking the outermost pair — so the section reaches the arc without dropping any chord crossing the existing cuts rely on (the lip cuts graze corner chords and depend on them).

2. **Plane section ending mid-arc on the boundary.** The section's endpoint landed strictly inside a convex corner arc; the arc wasn't split there, so the wire builder pruned the section as a pendant. Fix: split the plane boundary arc at section endpoints using the **shorter-arc** convention `evaluate_edge_at_t` uses (`domain_with_endpoints`'s CCW span misses a clockwise-traversed corner arc — the inconsistency that made the naive approach fail).

3. **Adjacent region misclassified as a hole.** A plane face split by a single section into two side-by-side regions could hand back the second region wound clockwise (negative area), so it was treated as a *hole* of the first and kept inside it. Fix: promote a negative-area loop to a separate outer when a point strictly inside it lies in no other outer (adjacent ≠ nested).

4. **FF line section clipped only to its own face.** The same intersection line, clipped independently per face, came out at different extents because a wide face (box cap, x=±21) extended it past the narrower opposing face (notch tool front, x=±13.24). Fix: clip to the opposing FF face as well, taking the intersection of the two clips. (`trim_t_range_to_aabb` uses the bbox **union**, so the FF line was over-extended upstream; the per-face opposing clip corrects it without touching that shared trim.)

## Verification

- New committed regression `cut_wall_notch_straddling_top_edge_is_watertight`: free=0, over=0, euler=2, closed-manifold, 8 corner cylinders (analytic, no mesh fallback).
- **Zero regressions**: `brepkit-algo` (126) + `brepkit-operations` full suite (697 lib + all integration) + gridfinity 26/26, all green. The lip-cut tests that the naive arc-clip regressed (`cut_concentric_rounded_rect_arc_prisms_*`, `fuse_arc_frame_lip_*`, `fuse_shelled_cup_lip_*`, `fuse_*_rounded_rect_arc_prisms_*`) pass unchanged.
- fmt, clippy `-D warnings` (algo + operations), boundaries: clean.

## Known remaining (not in this PR)

The real tool-side `2×2 wall cutouts` bin uses a **wide** tool (≈70% of the wall span) whose sides cross the body's **own rounded corners**, producing tool-front × body-corner-cylinder sections. That case is still non-manifold (free=20, cyl=6 in a native repro at hw=18 vs free=0 at hw=16) and continues to mesh-fall-back tool-side. It is a deeper analytic configuration beyond the corner-straddle scope here; documented for a follow-up. This PR is a strict, regression-free improvement that fixes the isolated straddle class and lays the groundwork.

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -18,7 +18,9 @@ use brepkit_topology::edge::EdgeCurve;
 use brepkit_topology::face::{FaceId, FaceSurface};
 
 use super::classify_2d::{sample_interior_point, signed_area_2d};
-use super::pcurve_compute::evaluate_edge_at_t;
+use super::pcurve_compute::{
+    compute_pcurve_on_surface, evaluate_edge_at_t, project_point_on_surface,
+};
 use super::plane_frame::PlaneFrame;
 use super::split_types::{OrientedPCurveEdge, SectionEdge, SplitSubFace, SurfaceInfo};
 use super::wire_builder::{build_wire_loops, build_wire_loops_with_winding};
@@ -29,7 +31,10 @@ use conversion::{
     boundary_edges_to_pcurve, extract_plane_normal, is_point_on_boundary_uv,
     uv_endpoints_from_pcurve,
 };
-use edge_splitting::split_boundary_edges_at_3d_points;
+use edge_splitting::{
+    find_splits_on_circle, find_splits_on_ellipse, find_splits_on_line,
+    split_boundary_edges_at_3d_points,
+};
 use sampling::{sample_wire_loop_uv, sample_wire_loop_uv_periodic};
 use special_cases::{
     split_face_with_internal_loops, split_noseam_face_direct, split_periodic_face_into_bands,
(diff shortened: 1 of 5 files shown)


andymai:
Superseded by #899 (re-applied the notch-corner trim as 9ae470c) and #903 (arc-aware `split_plane_face_by_arrangement` — the real fix for the rounded U-notch, which rewrote this same splitter). This branch's `b2f888c` now conflicts hard against main's newer splitter and cannot rebase cleanly. The straddle geometry it targeted is not reached by any gridfinity parity case (the real bin's corner r=1.485 < overshoot). The one unique commit, `a82b162` (perpendicular cylinder-cylinder analytic intersection), is a genuine general math improvement but is tangential to the current suite and conflicts with #899/#903; it remains recoverable in git for a clean standalone re-extract if a future case needs it. Closing to keep the PR queue accurate.

PR #917, opened 2026-06-19
fix(algo): keep convex boundary arcs whole in plane arrangement (2×2 compartments+scoop fuse)

andymai:
## Summary

Flips the **2×2 compartments+scoop** bin from a 135-face all-planar mesh fallback to a **102-face analytic solid** (cyl:8, valid), tool-confirmed. Tenth of eleven gridfinity cases now analytic.

## Root cause

The bin floor is a rounded rectangle (4 line walls + 4 **convex circle corner arcs**), cut by 4 coplanar scoop-base section loops. The angular wire builder produced a self-crossing loop (two scoop bases touching on a shared divider line → figure-8). The planar-arrangement fallback that exists to repair self-crossing traces **bailed** at its `is_arc && ts.len() > 2` guard — because a scoop-base section line crossed a convex corner arc's straight **chord** ~1mm clear of the actual outward-bulging arc, registering a *phantom* break that subdivided an **uncrossed** arc. On bail, the malformed self-crossing floor face survived → mesh fallback.

(An earlier attempt relaxed this bail *blindly* — subdividing the arc at the chord crossing — which broke the boundary and flipped 0/3 tool cases. The correct fix is the opposite: keep the arc whole.)

## Fix

In `split_plane_face_by_arrangement` (`face_splitter/mod.rs`), gate arc break-point registration on the point actually lying on the arc geometry: sample the arc (`ARR_ARC_SAMPLES`) and reject a break farther than `tol` from it. A chord-only crossing no longer subdivides the arc, so the corner stays a single sub-edge and the arrangement decomposes the floor into clean regions. Gated on `is_arc` — Line inputs are unchanged.

## Verification

- Faithful **arena-captured** repro (the in-memory operands — STEP/nudge normalize the trigger away): was 5 free / mesh fallback → now **watertight, 102 faces, 8 corner cylinders, 0 free**.
- **Tool oracle**: `2×2 compartments+scoop` 135 all-planar → **102 faces {cyl:8, plane:94}, valid**. ✓
- Full suites green: algo (135), operations (700+ incl. `rounded_rect_arc_prism`/`coincident_rrect_cap`/`wall_cutout`/`parity_boolean`/`same_domain` — the arc-arrangement neighbors), io (133 + in-mem guards), wasm (226 incl. gridfinity d1–d5). Zero regressions.
- New faithful guard `gridfinity_compartments_scoop_fuse_2x2_inmem_is_watertight` (+ arena `.bin` fixtures).
- fmt + clippy (`-D warnings`) + check-boundaries clean.

## Scope

`3×3 scoop+label+lip` (the 11th case) remains a fallback — a *distinct, deeper* root (the label-bracket fuse fragmenting the lip corners' curved FF sections at a 4-surface junction), reported separately. This fix's arrangement bail fires zero times for that case, so it's unaffected.


andymai:
Good catch — fixed in 2f4f399. `evaluate_with_endpoints` takes the native parameter (radians for Circle/Ellipse, knot value for NURBS), so the `[0,1]` sampling was probing the wrong arc portion. Now mapping the sample fraction onto `domain_with_endpoints(start,end)` so the probe points lie on the real trimmed arc. Verified: comp+scoop in-mem guard, lip guards, rounded_rect_arc corpus, the #903 wall-cutout (which relies on real arc breaks being honored), and same_domain all green.

PR #919, opened 2026-06-19
fix(algo): base EF containment margin on curved boundary edges only

andymai:
## Problem

The edge-face containment test in `pave_filler::phase_ef::build_face_containment` bounds each candidate crossing against the face's sampled boundary, applying a sagitta margin of `max_chord * 0.5` to retain true near-boundary crossings on chord-sampled **curved** arcs.

That margin was accumulated over **all** boundary edges — including straight `Line` edges, which have zero sagitta (their sampled chords coincide with the edge exactly). For a thin planar strip with long straight sides (the gridfinity label bracket's ramp facets are ~123mm wide × ~1mm tall), the long straight edge inflated the margin to **~1.9mm**, so edge-face crossings up to ~1.9mm *outside* the strip's bounded region passed the containment check.

In the 3×3 scoop+label+lip bin's label-bracket fuse, the bin's lip-corner step circles (z = 13.3 / 14.7 / 16.6, up to ~1.6mm outside a ramp strip's z-band) were wrongly accepted as IN-edges → fed to the face splitter as ~20 spurious sections per strip → realized as ~50 degenerate zero-area out-and-back "ribbon" sub-faces that over-shared their boundary circles → non-manifold → mesh fallback.

## Fix

Base the sagitta margin only on **curved** boundary edges. An all-`Line` strip now gets a tolerance-scale margin and the off-band crossings are correctly rejected. Effect on the faithful in-memory repro (operands captured via `serializeSolid`): degenerate sliver faces 50→0, over-shared edges 20→0.

## Scope / risk

A necessary correctness fix the 3×3 scoop+label+lip bin exposed; **not by itself sufficient** to flip that case (a separate ramp corner-split seam remains — fix in progress). Shipped on its own merits: a straight edge contributing to a chord-deviation margin is wrong, and the over-extended band can admit spurious EF crossings in any fuse where a thin planar face with long straight sides meets curved features at a near-but-off-band parameter.

## Verification
- Faithful regression guard `crates/io/tests/scooplabel_ef_margin_inmem.rs` (asserts over-shared == 0 + no spur faces; **fails without** this fix).
- Full suites green: `brepkit-algo` (135), `brepkit-operations` (all), `brepkit-io` (incl. #915 `gridfinity_lip_fuse_3x3_inmem`, `scoop_fix_inmem`, `lipfuse_fixture`), `brepkit-wasm`. `same_domain`, all lip/scoop/coaxial/parity/gridfinity tests pass.


andymai:
Addressed in b27a69f — the guard now tracks distinct `FaceId`s per quantized edge key (a manifold edge has exactly 2; >2 distinct faces = the non-manifold over-share), so an edge repeated across one face's outer+inner wire no longer reads as over-shared. Guard still passes.

PR #921, opened 2026-06-19
fix(algo): split holed planar cap whose cut bridges material between holes

andymai:
## Problem

Fusing a stacking lip onto a **compartmented** (interior-divider-walled) bin body mesh-falls-back — and that's the **default** gridfinity bin config (almost every real bin has compartments). The result drops to a 228-facet all-planar mesh, losing every analytic surface.

Isolated with clean controls (faithful operands captured from the live tool via `serializeSolid`, #915):
- plain lip body (no dividers): clean analytic ✓
- multi-cavity cut alone: clean ✓ (negative control — `gridfinity_cavitycut_inmem`)
- `fuse(cavity-cut body, lip)`: **14 free LINE edges at z=23** tracing the interior divider "+" cross → mesh fallback ✗

## Root cause

A **face-splitter** bug (not classification — point classification against the lip is correct: the divider-"+" interior is Outside→keep, the perimeter is Inside→drop). The body's z=23 top face is a holed planar cap (rounded square with the 4 compartment openings as holes; the divider cross is the material between them). It receives the footprint split-lines it needs (4 lines at the lip inner wall ±39.25 + outer-corner arcs), but two splitter gaps left it unsplit, so it was classified *whole* by a single interior sample and dropped:

1. `integrate_holes_plane` bailed on **any** arc section (`if any non-Line { return None }`), so the holed-cap integration never ran when the cap also carried outer-corner arc sections.
2. No decomposition handled the resulting **multiply-connected** topology — the footprint cut bridges *across* the divider material *between* the holes, which the angular wire builder mis-traces into one winding loop.

## Fix (`crates/algo/src/builder/face_splitter/mod.rs`)

1. `integrate_holes_plane`: partition sections into Line vs arc; trim Lines at hole crossings as before, carry arc sections through whole (bail only if an arc crosses a hole's straight wall).
2. New even-odd arrangement decomposition (`arrangement_regions_from_combined`): depth-even = solid (with direct-child loops as inner wires → the outer disk minus the inner hole = the perimeter annulus); depth-odd = hole; air regions filling original holes dropped. Containment uses all-vertices-inside + strictly-larger-area (a single probe is symmetric for concentric disks). Classifier seed uses `find_point_outside_holes` so an annular region isn't sampled in its own hole.
3. **Gating (regression-critical):** run the wire builder first; use the arrangement only when `original_inner_wires.len() >= 2`. The even-odd path is better for multi-hole caps (divider-lip) but worse for the single-cavity shelled wall-cutout rim, which the wire builder already partitions cleanly.

## Result

`fuse(compartmented body, lip)` production: **228 all-planar fallback → 120 faces, 36 curved, 0 free edges, watertight + analytic** (volume preserved). The 14 divider-cross free edges are resolved.

## Verification

- New faithful guards: `gridfinity_lipfuse_dividers_inmem` (asserts watertight + analytic) + `gridfinity_cavitycut_inmem` (negative control: the cavity cut alone is clean).
- **Full workspace green** (2389 tests); independently re-confirmed operations+io (1076) + the regression-critical set: all `gridfinity_stacking_lip_fuse_*`, `scoop_fix_inmem`, `scoop_corner_cylinder_fuse`, `scooplabel_ef_margin_inmem`, `same_domain`, `wallcut_fixture`, wall-notch booleans, `coaxial_*`, `multicavity_cut`, `*rounded_rect_arc*`, `parity_boolean_*`, gridfinity d1–d5.

## Notes

- 8 *pre-existing* lip-inner-wall out-and-back spur stubs remain in the raw GFA (present on the plain shelled-cup lip too, on rank-B faces this change never touches); production's sew heals them, so the result is watertight+analytic. A separate scoped follow-up.
- This is the same coincident-contact face-split frontier as the remaining 3×3 scoop+label case; a general fix here is a step toward that too.


andymai:
Addressed all three findings in 35aebf5:
- **L551 (arc pcurve):** `integrate_holes_plane` now recomputes each arc section's pcurve in the current face frame via `compute_pcurve_on_surface` (threaded `surface`/`wire_pts` through), matching the main section path's plane refit — no more unconditional `pcurve_a.clone()`, so it's correct regardless of rank/source frame.
- **L541 (arc-in-hole):** the over-air drop test now samples the arc's *geometric* midpoint (its bulge via `evaluate_with_endpoints` at the domain midpoint) instead of the chord midpoint, catching an arc that bows into a hole while its chord sits clear.
- **L193 (guard):** relaxed to `free <= 8` so it survives the separate lip-spur-removal follow-up.

algo (135) + io (133) green after the changes; divider-lip guard still flips to watertight+analytic; operations suite re-running.

PR #923, opened 2026-06-20
fix(algo): keep minuend wall for opposite-oriented coincident Cut pair

andymai:
## Problem

Two gridfinity wall-cutout tools applied in sequence to a compartmented bin produced a **non-manifold** result, even though each cut is watertight on its own:
- `Cut(body, cut0)` → 150 faces, watertight ✓
- `Cut(body, cut1)` → 120 faces, watertight ✓
- **`Cut(Cut(body, cut0), cut1)` → 62 free edges** ✗ → downstream collapse

(Faithful 2-operand `.bin` arena repro, captured via `serializeSolid` #915; survives a STEP round-trip, so it's a genuine geometric defect, not an in-memory arena-id artifact.)

## Root cause

Both wall-cutout tools carve to the same back-plane (x = ±28.385) and overlap there. After cut0, the body carries a back-wall face at x=28.385 that is **fully contained in** cut1's coincident, **opposite-oriented** tool face. `detect_same_domain` correctly flags this as a geometric-overlap SD pair, but `apply_sd_selection`'s **Cut** branch for an opposite-oriented pair kept the minuend wall (A) *only when the faces were exactly touching (coextensive)* and otherwise **discarded both** — so the contained back-wall was dropped, leaving the notch outline on the adjacent floor/side faces with no partner edge → 62 free edges (all classified PARTIAL = dropped-face/T-junction, not the #859/#867 near-duplicate class).

## Fix

For an **opposite-oriented** coincident Cut SD pair, always keep the minuend's wall (A). Opposite orientation means the minuend's material and the tool's material lie on **opposite sides** of the shared plane (the faces abut back-to-back), so the cut removes nothing at that plane and the minuend wall is a genuine result boundary — whether the faces are coextensive **or one is contained in the other**.

The same-**orientation** Cut path (overlap cut away from A, cap from the B-Inside face — the #462 `cut_overlapping_cubes` case) is untouched. `bop.rs::apply_sd_selection` only; plus doc/comment accuracy cleanup in `same_domain.rs` (`b_contained_in_a` is now read only by tests).

## Result

- Raw `gfa::boolean` sequential cut: free **62 → 0**, over-shared 0, **156 faces**, analytic preserved (12 cone + 40 cylinder + 104 plane — the 6 extra vs the broken 150 are the correctly-retained back-walls).
- `operations::boolean`: watertight + manifold, `validate_solid` valid, volume preserved (16082.8 → 16078.4; cut1 removes the small notch remainder).

## Verification

- New guard `gridfinity_wallcut_seq_inmem::wallcut_sequential_cuts_stay_watertight` (faithful `.bin` repro).
- **Full workspace: 1436 passed, 0 failed.** Scrutinized the #462-protected `boolean_stress` (`cut_complement_identity`, `anti_commutativity`, `volume_cut_overlapping_boxes`, `half_overlap_cut` — all green), `boolean_invariants`, `same_domain`, all `gridfinity_*` (incl. `gridfinity_lipfuse_dividers_inmem` #921, `scoop_fix_inmem`, lip-fuse 2×1/3×3, multicavity, cavitycut), `coaxial_*`, `parity_boolean_*`, `wallcut_fixture`.
- fmt + clippy (`--all-targets -D warnings`) + `check-boundaries.sh` clean.


andymai:
Addressed all three in 675f45d:
- **wallcut_seq L2:** fixed the stale module doc (it now documents the watertight fixed behavior, with the pre-fix defect as context).
- **same_domain L43 (×2):** renamed `b_contained_in_a` → `geometric_overlap` (it fires for A-in-B, B-in-A, *and* partial overlap — the new name matches both the semantics and the internal variable). Field is test-read-only; compiler verified all sites + shorthand applied.

algo same_domain (29) + the wallcut_seq guard green after the rename.

PR #925, opened 2026-06-20
fix(algo): use planar arrangement when wire-builder loops overlap (honeycomb cap)

andymai:
## Problem

Cutting a gridfinity honeycomb wall-pattern into a bin that already has wall cutouts caused `gfa::boolean(Cut, body, pcut0)` to fail with **`AssemblyFailed("no outer shell found (all shells classified as holes)")`**. In the tool this is the **multi-minute hang**: `cutAllBisect` retries the failed cut by recursively bisecting (re-running the expensive NURBS honeycomb cut each time) and never converges. It blocks two real scenarios (honeycomb + wall-cutouts, honeycomb + scoop).

(Surfaced precisely *because* #923 made the wall-cut body watertight — the failure moved downstream to the honeycomb cut. Faithful 2-operand `.bin` arena repro via `serializeSolid` #915.)

## Root cause

`pcut0`'s z=23 **top cap** is split by the body's lip + wall-cutout faces. Because the wall cutouts **vent the cavity to the exterior**, the cap's inner-wall ring is an *incomplete* dividing boundary (gaps where the cutouts open). The angular wire builder (`wire_builder::build_wire_loops`) discards incomplete loops, so it returns a partition that is **not edge-disjoint**: one greedy outer loop traces the whole ±41.75 cap perimeter while sibling outer loops trace corner sub-regions sitting *inside its material*. The full-perimeter sub-face mis-samples (single interior point) as Inside, is kept reversed, and double-covers the z=23 plane → open + **inverted** shell (signed volume < 0) → `perform_areas` classifies every shell as a hole.

(Ruled out same-domain first per the debugging playbook: the coincident ±41.75 walls produce `same_orientation` SD pairs that Cut correctly discards — verified with minimal coincident-wall Cut repros.)

## Fix

`split_plane_face_by_arrangement` already builds an **edge-disjoint** decomposition that resolves the cap, but its gate only triggered when the arrangement produced *more* regions or the greedy loops *self-cross* — and here neither holds (both yield 6 non-self-crossing regions). Added a third trigger `greedy_outer_loops_nested`: fires when one greedy outer loop's polygon **contains another outer loop with no hole between them** (the non-edge-disjoint signature). The **hole-separation guard is essential** — without it, it false-positives on a valid material island (washer-with-post) and regresses `multi_cavity_shell_cut`.

Deliberately **not** broadened to "always arrange": that makes pcut0 fully watertight but provably regresses `parity_boolean_planar::corner_cut_notch` (volume 0.833 vs 0.75).

## Result

| tool | before | after |
|------|--------|-------|
| pcut0 | **AssemblyFailed** | **OK**, 59 faces, free=8, over=0 |
| pcut1/2/3 | OK (free 52/35/15) | OK (unchanged) |

The catastrophic `AssemblyFailed` — the **hang trigger** — is eliminated: all four cuts return Ok, so the tool's bisect-retry no longer spins → **no hang** (multi-minute freeze → fast result).

## Residual (precisely diagnosed, deferred)

pcut0's free=8 is a **separate cross-face partition-consistency gap** (the cap's correct arrangement introduces vertices its neighbor analytic faces don't share, because their greedy split picks a *different valid* partition). Forcing every planar neighbor to arrange makes pcut0 free=0 — proving it's pure neighbor inconsistency — but that's the `corner_cut_notch`-regressing path. The principled fix belongs in the reconciliation layer (`builder_solid::split_edges_at_collinear_vertices`), not the per-face splitter; left for a separate scoped change.

## Verification

- New guards in `gridfinity_honeycomb_cut_inmem.rs`: `honeycomb_cut_no_longer_throws` (pcut0..3 all Ok + manifold) + `honeycomb_cut_residual_documented` (pins remaining free counts) + `wallcut_step_is_watertight` (#923).
- **Full workspace green**: algo 135, operations 902 (incl. `boolean_stress`, `parity_boolean_planar::corner_cut_notch`, `multi_cavity_shell_cut`, all `gridfinity_*`), io all, wasm 0 failed.


andymai:
Addressed both in b52b795:
- **L682 (poly_contains concave):** now tests each edge **midpoint** in addition to the vertices, so a concave `outer` can't spuriously 'contain' an `inner` edge that exits and re-enters between vertices. Honeycomb fix + the gated guards (corner_cut_notch, multi_cavity_shell_cut) still green.
- **L188 (residual pins):** changed exact `==` to `<=` ceilings (the quantized-coord counts are ULP-fragile on cross-compiled targets); the qualitative invariant (all cuts Ok + manifold) stays in `honeycomb_cut_no_longer_throws`.

operations lib 700 + the honeycomb guard green after the changes.

PR #926, opened 2026-06-20
perf(algo): spatial-hash the builder's O(N²) collinear-split + same-domain passes

andymai:
## Problem

Sequential boolean cuts scale **quadratically**: `regress_hexwall_cuts::hexwall_sequential_cuts_full` (133 hex-prism cuts through a wall) took **19.4s release / ~219s debug**, and its own doc says it "must complete in seconds." Each `boolean(Cut, growing_body, small_tool)` was O(body²), so the sequence was quadratic-of-quadratic. brepkit must be performant; a slow test is an engine perf smell.

## Root cause (profiled, not assumed)

The PaveFiller is **not** the culprit — it already AABB-pre-filters face pairs (~13ms/cut). Stage-level timing of the heavy end (cut 131, ~800 body faces, ~309ms) attributes the cost to **three all-pairs passes in the builder/assembly stage**:

| Pass | File | @cut131 | Pattern |
|---|---|---:|---|
| `split_edges_at_collinear_vertices` | `builder_solid.rs` | ~140ms | every vertex × every Line edge — O(V·E) |
| `split_arc_edges_at_collinear_vertices` | `builder_solid.rs` | (O(V·E)) | same, arc edges |
| `detect_same_domain` Steps 3b/3c | `same_domain.rs` | ~140ms | every coplanar/coaxial face pair × polygon-boolean — O(F²) |

## Fix — broad-phase locality at the layer where the cost lives

- A uniform spatial hash `PointGrid` (over candidate vertices) and an AABB-grid candidate-pair generator `overlap_candidate_pairs` (over faces).
- Each pass now probes only the grid cells an edge's/face's AABB spans → **O(near) instead of O(N²)**.
- **Provably result-preserving:** a vertex beyond the snap band from a segment can never be a collinear cut; two faces whose AABBs don't overlap can never pass the overlap test. Candidate pairs are emitted in the **same ascending (i,j) order** the former nested loop used, so the same-domain union-find sequence and representative selection are **byte-identical**.

## `cut_all` batch — investigated and rejected

I implemented a `merge_disjoint_solids`-then-single-cut batch path and found it produces **non-manifold** results for through-hole patterns (the merged disconnected tool shell breaks the cut's shell assembly — fails `compound_cut_matches_sequential_{2x2,3x3,4x4}_grid` manifold assertions, volume right / manifoldness wrong). Dropped it; `compound_cut` stays sequential and correct. The locality fix above is the provable one.

## Results

| Profile | Baseline | Fixed | Speedup |
|---|---:|---:|---:|
| release | 19.40s | **5.18s** | 3.7× |
| debug | ~219s | 62.85s | 3.5× |

The residual is the fundamental "a sequential boolean reprocesses the whole body each cut" floor (now genuinely **linear per cut**, O(body²) total with a 3.5× smaller constant); flattening further needs incremental far-field carry-through — an architecture change, out of scope for a provable local fix.

## Verification

- **Full suites, zero regressions/diffs (1439 tests):** `brepkit-algo` (135, all `same_domain` geometric-overlap cases), `brepkit-operations` (902, incl. `parity_boolean_*` VOLUME oracles, `boolean_stress`/`boolean_invariants`, `coaxial_*`, `compound_cut` grid/shelled), `brepkit-io` (176), `brepkit-wasm` (226, gridfinity d1–d5 + honeycomb + lipfuse). `hexwall_sequential_cuts_20`/`_full` pass (each cut removes the exact prism volume).

## Follow-up (same O(N²) class, not in scope here)

The per-cut O(body) rebuild of `detect_same_domain`'s edge-sets/AABBs is the remaining linear-per-cut floor; eliminating it needs an incremental boolean that carries the untouched far-field through unchanged.

### DIFF
--- a/crates/algo/src/builder/builder_solid.rs
+++ b/crates/algo/src/builder/builder_solid.rs
@@ -1593,7 +1593,7 @@ fn split_arc_edges_at_collinear_vertices(
         // span, so the spatial query covers every vertex that could lie on it.
         let mut amin = sp;
         let mut amax = sp;
-        let arc_samples = 12;
+        let arc_samples = ARC_AABB_SAMPLES;
         for k in 0..=arc_samples {
             let f = f64::from(k) / f64::from(arc_samples);
             let a = a0 + (a1 - a0) * f;
@@ -1950,6 +1950,11 @@ fn build_edge_face_map(
 
 /// Tolerance for position quantization (matches system linear tolerance).
 const MERGE_TOL: f64 = 1e-7;
+
+/// Samples per arc when building its broad-phase AABB for the collinear-split
+/// query (its bulge is covered by inflating the query band with the per-step
+/// sagitta — see `split_arc_edges_at_collinear_vertices`).
+const ARC_AABB_SAMPLES: u32 = 12;
 
 /// Get all edge keys (quantized position-pair) for a face's wires.
 fn face_edge_keys(topo: &Topology, fid: FaceId) -> Result<Vec<VPair>, AlgoError> {
--- a/crates/algo/src/builder/builder_solid.rs
+++ b/crates/algo/src/builder/builder_solid.rs
@@ -951,9 +951,12 @@ impl PointGrid {
         // a long edge): iterating every empty cell would defeat the speedup.
         // Iterating the populated buckets directly is still a superset and
         // bounded by the point count, so correctness is preserved.
-        let cells = (chi.0 - clo.0 + 1)
-            .saturating_mul(chi.1 - clo.1 + 1)
-            .saturating_mul(chi.2 - clo.2 + 1);
+        let cells = chi
+            .0
+            .saturating_sub(clo.0)
+            .saturating_add(1)
+            .saturating_mul(chi.1.saturating_sub(clo.1).saturating_add(1))
+            .saturating_mul(chi.2.saturating_sub(clo.2).saturating_add(1));
         let bucket_budget = i64::try_from(self.buckets.len())
             .unwrap_or(i64::MAX)
             .saturating_mul(4);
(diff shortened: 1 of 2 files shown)


andymai:
Addressed all three in c6f9150:
- **L1596 (arc AABB conservativeness):** the 12-sample arc AABB could under-cover the bulge by up to one step's sagitta (≫ snap for large radii). Now inflating the query band by `radius·(1−cos(step/2))` — provably conservative, no missed collinear split. (The suite was passing only because the test arcs' vertices didn't land in the missed band; this closes the latent gap.)
- **L956 (cell-count overflow):** `saturating_sub`/`saturating_add` so an extreme float→int cell index can't overflow-panic before the `saturating_mul` guard.
- **L1081 (face_outer_aabb):** sample `0..N` (not `..=N`) to match `planar_faces_overlap` exactly + drop the redundant per-vertex duplicate (AABB unchanged).

Verified: parity volume oracles, same_domain (29), coaxial_*, rounded_rect_arc, gridfinity io, hex-wall release still 5.3s; clippy clean.

andymai:
Addressed both re-review findings in d5f1773:
- **same_domain L1167 (cell-budget guard):** `overlap_candidate_pairs` now mirrors `PointGrid::box_candidates` — a face whose expanded AABB spans more cells than ~4× the face count is AABB-tested against all faces directly instead of gridding O(cells) (a broad patch among tiny facets). The candidate set stays a superset of true overlaps (narrow phase filters, pairs sorted before union-find), so the same-domain result is unchanged. same_domain (29) green.
- **builder_solid L1596:** named the literal — `ARC_AABB_SAMPLES = 12` as a module const next to `MERGE_TOL`.

PR #932, opened 2026-06-20
fix(operations): reverse periodic-curve parameterization for reversed extrude edges

andymai:
## What

Follow-up to #930 (the reversed-edge limitation it documented). Extrude's translated **top** edge was built with wire-traversal-order vertices but a curve from `translate_edge_curve` that keeps the bottom edge's **stored** parameterization. For an edge used reversed in its wire those disagree:

- a periodic curve (ellipse/circle) resolves to the **complementary sub-arc** via `domain_with_endpoints`, and
- a NURBS curve's endpoints no longer match the stored vertices.

The swept side then sweeps the wrong arc — the reversed half-ellipse came out **98.33 vs 65.45**.

## Fix

Add `reverse_edge_curve` (flip `normal`+`v_axis` for circle/ellipse so `project` → `-θ`; reverse control net, weights, and mirrored knots for NURBS) and apply it to the translated top curve when the source edge is reversed. **No-op for forward edges**, so the common path (all gridfinity profiles, every existing test) is byte-identical.

Single fix site: the shared `extrude_wire_vertices_with` top-edge loop (covers both outer and inner wires).

## Verification

- Un-ignores `extrude_half_ellipse_reversed_edge_volume` (now passes) + adds `extrude_half_circle_reversed_edge_volume`.
- Full `brepkit-operations` suite green (704 passed, 0 failed); clippy clean.
- Forward path unchanged (guarded by `is_forward()`).

Closes the reversed-edge follow-up tracked from #930.

### DIFF
--- a/crates/operations/src/extrude.rs
+++ b/crates/operations/src/extrude.rs
@@ -353,7 +353,16 @@ fn extrude_wire_vertices_with(
     for i in 0..n {
         let next = (i + 1) % n;
         let bottom_curve = topo.edge(edge_ids[i])?.curve().clone();
-        let top_curve = translate_edge_curve(&bottom_curve, offset)?;
+        let mut top_curve = translate_edge_curve(&bottom_curve, offset)?;
+        // The top edge's stored vertices (top_verts[i], top_verts[next]) follow
+        // wire-traversal order. For a reversed edge those are swapped relative
+        // to the curve's parameterization, so a periodic curve (ellipse/circle)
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the 4/5 finding: added `extrude_half_disc_reversed_nurbs_edge_volume` (8953911) — a reversed NURBS half-disc profile now guards the NURBS arm of `reverse_edge_curve`, matching the Circle/Ellipse volume regressions.

PR #936, opened 2026-06-20
fix(math): bounded oblique plane-cone conic (was unbounded both-nappe sweep)

andymai:
## What

`exact_plane_cone` only handled the perpendicular (Circle) case. The oblique fallback (`sample_plane_cone`) accepted any `|v| <= 100`, so it swept the **unbounded double-napped infinite cone** — including the phantom `v<0` nappe and points running toward the asymptote — and returned a garbage `Points` chain.

Probe on a real oblique cut (45° cone, steep plane → hyperbola):
- before: `Points n=63, z=[3.66, 28.37], max_cone_err=24.9, wrong_nappe=29`
- after: `Points n=249, z=[7.69, 14.75], max_cone_err=3.5e-14, wrong_nappe=0`

## Fix

Classify the conic in the plane-aligned frame by `A = sin²θ − sin²(half_angle)`:
- **Ellipse** (`A<0`): exact closed-form `Ellipse3D` (center, semi-axes derived analytically) — strictly more correct than the old sampled fallback.
- **Parabola / hyperbola**: bounded, branch-separated sample chains on the real (`v≥0`) nappe only, capped at `8·v_min`; near-asymptote samples dropped; contiguous-`u` runs split into separate chains.
- Perpendicular Circle path unchanged.

3 unit tests assert every returned point lies on both the plane (<1e-9) and cone (<1e-7), on the real nappe, within bound (ellipse / parabola / hyperbola, hyperbola using gridfinity scooplabel geometry).

## Verification

brepkit-math (455), -algo (135), -operations (904), -io (177) all green — no regressions (incl. all coaxial-cone + cone-boolean suites). clippy + fmt + layer-boundaries clean.

## Notes

Foundation for the gridfinity scooplabel scoop+label fuse (the oblique tread×lip-cone hyperbola). It does NOT by itself close that case: a phase_ff hyperbola-weld attempt on top was measured to regress (cone orphaned in shell assembly, 21→30 free edges) and reverted — the scoop close needs a separate shell-assembly z-step ring-reconciliation, tracked separately. A sibling `intersect_plane_cone` (used by `section.rs`) has the same negative-`v` bug (~L538), left as a follow-up.

### DIFF
--- a/crates/math/src/analytic_intersection.rs
+++ b/crates/math/src/analytic_intersection.rs
@@ -153,8 +153,12 @@ fn exact_plane_sphere(
 
 /// Exact plane-cone intersection.
 ///
-/// - Plane perpendicular to axis → `Circle3D`
-/// - Otherwise → `Points` fallback (could be ellipse, parabola, or hyperbola)
+/// The conic type is set by the cone's half-opening angle from the axis
+/// (`γ = π/2 − half_angle`) versus the plane-axis angle `ψ`:
+/// - Plane perpendicular to axis (`ψ = π/2`) → `Circle3D`
+/// - `ψ > γ` (ellipse) → closed-form `Ellipse3D`
+/// - `ψ ≤ γ` (parabola/hyperbola) → bounded single-branch `Points` (one chain
+///   per branch — a hyperbola's two nappes never share a chain)
 fn exact_plane_cone(
     cone: &ConicalSurface,
     normal: Vec3,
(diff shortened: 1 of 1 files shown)


andymai:
Addressed the 3/5 finding (e9408d4): added the real-nappe guard `if e*c < 0.0 { return Ok(vec![]); }` to the closed-form ellipse path — when the plane is offset to the far side of the apex the section lies entirely on the phantom v<0 nappe, so it now returns empty instead of a spurious ellipse. Added regression test `oblique_plane_cone_wrong_nappe_is_empty`. Full math suite (456) + clippy green.

PR #938, opened 2026-06-21
fix(algo): drop doubled faces in solid assembly (baseplate dovetail groove cut)

andymai:
## What

The baseplate dovetail **groove cut** produced non-manifold results (over-shared edges → mesh fallback → 5/11 dovetail scenario tests failed in the downstream tool). Root cause (captured via faithful in-memory operands, NOT reproducible from clean primitives): `Cut(base, groove_tool)` over-produced a **duplicate face** — the slab's slanted outer wall (rank A) and the groove flank (rank B) each emitted the *same* triangle (one reversed), so two boundary edges became 4-face-incident. The upstream cause is `build_section_edges` feeding a planar face's split some foreign sections with endpoints up to ~1.3 mm off its plane; the arrangement weaves them into a spurious triangle duplicating the flank's.

Not a weld/dedup bug (a near-duplicate-vertex scan found zero near-dup pairs at every tolerance band — the vertices were *exactly* shared). It is duplicate-face over-production.

## Fix

`remove_doubled_faces` in `builder_solid.rs`, run after `merge_duplicate_edges`: group selected faces by their sorted outer-wire edge-ID multiset; any group of ≥2 is geometrically coincident copies bounding zero volume → dropped. Exact key (no tolerance), sound (a coincident pair with an identical boundary cancels). An alternative section-coplanarity filter was tried first and **rejected** — it regressed the honeycomb-cut guard (free 52→65); this output-side removal has minimal blast radius.

## Result

Both groove cuts (right + back join edge): over=2, free=0, 23 faces → **over=0, free=0, 21 faces**, all-planar/analytic (no mesh fallback).

## Verification

New guard `dovetail_groove_cut_inmem.rs` (2 tests, faithful captured `.bin` operands) + full `brepkit-algo` (135), `brepkit-operations` (702+), `brepkit-io` (incl. `scoop_fix_inmem`, `scooplabel_ef_margin_inmem`, `gridfinity_honeycomb_cut_inmem`, `gridfinity_lipfuse_dividers_inmem`) all green; clippy + fmt + layer-boundaries clean.

## Residual (not addressed)

The deeper section-attribution bug (why `build_section_edges` feeds off-plane foreign sections to a planar face) remains upstream; this removes its non-manifold symptom soundly. The dovetail scenario's 5.3 h runtime was the *lightweight floor cut*, not the dovetail (the dovetail cut is 0.02 s) — separate, unverified here.


andymai:
Addressed the 4/5 finding: the grouping key now includes inner-wire edges (not just the outer wire), so a holed face only matches a truly identical holed face — never a coincident spurious non-holed copy. The dovetail duplicate (a hole-less triangle) is still caught. Dovetail/honeycomb/scoop guards + algo suite green.

PR #940, opened 2026-06-21
fix(algo): harden periodic-face arc splitting (shorter-arc convention + analytic-face v-band clip)

andymai:
Two correct, general soundness fixes for periodic-face arc splitting in the GFA builder, discovered while driving the gridfinity scoop+label fuse close. Net effect on that fuse: free edges 21→20 (the bin corner cylinder now splits per z-band; 6 of 8 staircase ellipse arcs gain partners). All guards green; the remaining residual is a separate, precisely-localized z=13.3 ring-closure (tracked for a follow-up).

## Fix 1 — shorter-arc co-circular split convention (`face_splitter/edge_splitting.rs`)
`find_splits_on_circle`/`find_splits_on_ellipse` mapped a candidate split angle to `[0,1]` via `normalize_angle_in_span` over the CCW span from `domain_with_endpoints` (always positive 0..2π), but `evaluate_edge_at_t` traces the **shorter arc** (signed −π..π). On a circle whose two arcs are split, the genuine crossing projects onto both arcs and the old convention assigned it to the wrong one → a bogus 3D split point. Compute `t` under the shorter-arc convention (matching the existing `split_plane_boundary_arcs_at_points` precedent + `evaluate_edge_at_t`), select the owning arc by a 3D round-trip, and reject endpoint-coincident crossings within a snap tolerance. Two unit tests added. (Reconciled the honeycomb near-endpoint regression via the snap-tol guard.)

## Fix 2 — clip analytic-face section arcs to their v-band (`builder/fill_images_faces.rs`)
A plane×cylinder/cone section arc could be emitted with an endpoint **overshooting the analytic face's v-range** (e.g. v=13.912 on a face whose top is z=13.3). The dangling endpoint above the boundary circle prevented `split_face_2d` from closing a boundary-to-boundary cut, so the face stayed whole and was dropped by a single interior-sample classification. New `clip_arc_to_face_v_range` bisects the monotone-in-v arc to land the overshooting endpoint exactly on the face's v_min/v_max boundary circle (analytic side only; the planar partner keeps its full arc). Load-bearing detail: clip to the **exact** band edge, not the containment margin.

## Verification
`brepkit-algo` (137), `brepkit-io` (incl. scoop_fix_inmem, scooplabel_ef_margin_inmem, gridfinity_honeycomb_cut_inmem, gridfinity_lipfuse_dividers_inmem, scoop_corner_cylinder_fuse, dovetail_groove_cut_inmem, cavitycut, wallcut_seq), `brepkit-operations` (705+) — all green, 0 failed. clippy + fmt + layer-boundaries clean. Builds on #936 (oblique plane-cone conic, already in main).

### DIFF
--- a/crates/algo/src/builder/face_splitter/edge_splitting.rs
+++ b/crates/algo/src/builder/face_splitter/edge_splitting.rs
@@ -5,11 +5,10 @@ use brepkit_topology::edge::EdgeCurve;
 use brepkit_topology::face::FaceSurface;
 
 use super::super::pcurve_compute::{
-    compute_pcurve_on_surface, evaluate_edge_at_t, project_point_on_surface,
+    compute_pcurve_on_surface, evaluate_edge_at_t, project_point_on_surface, shorter_arc_delta,
 };
 use super::super::plane_frame::PlaneFrame;
 use super::super::split_types::OrientedPCurveEdge;
-use super::sampling::normalize_angle_in_span;
 
 /// Split boundary edges at 3D points where section edges start/end.
 ///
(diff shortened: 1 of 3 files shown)


andymai:
Closing — root-caused as net-negative.

The find_splits co-circular convention change (commit 5866888) regresses `gridfinity_d1_lip_ring_loft_cut` (validation 0→15, euler −50) — a shipped wasm-crate case. Caught by the full-workspace CI; the local verification ran -p algo/io/operations but not -p brepkit-wasm, so it was missed.

The arc-clip (commit dcc54db) is innocent (full workspace green on its own), but its apparent scoop benefit was entirely dependent on the unsafe find_splits splits: arc-clip alone leaves the scoop+label fuse free-edge count at 21, unchanged from main. Net effect of the pair vs main: scoop 21→20 (1 edge) at the cost of breaking d1. Not viable.

The genuine scoop close path (10th investigation, definitive): cone234 is kept-whole while ~75% is buried in the bracket; it needs a coordinated cone phase_ff split (θ=21°) + bracket-cap re-trim sharing the new edge. Tracked in memory; will be approached fresh, not on this branch.

PR #959, opened 2026-06-23
fix(measure): clamp volume tessellation deflection for accurate curved-face volume

andymai:
## Summary

`solid_volume` tessellated curved analytic faces at whatever deflection the caller passed. A coarse **preview** deflection (brepjs passes `DEFAULT_DEFLECTION = 0.01`) inscribes too few facets in a cylinder/cone/sphere face, so the inscribed mesh **under-counts** the volume by ~1-2% — e.g. a swept cylinder, or a box with a cylindrical hole.

Volume is a precise query, not a preview, so it now clamps the tessellation deflection to a small fraction of the solid's bounding-box diagonal (never coarsening a finer request). Primitive solids are unaffected — the exact analytic fast path (`try_analytic_solid_volume`) runs first.

## Why this surfaced

This fixes three pre-existing `--project brepkit` failures in brepjs (`sweeps a circle along a straight spine`, `extrudes a compound sketch with a hole`, `sweep with transition mode`) — all swept/holed cylinders whose volume measured ~2% low at the preview deflection. Because the clamp is internal to the kernel, **no brepjs change is needed**: every consumer's `volume()` call gets an accurate result once this releases.

## Evidence

- New regression test `extrude_analytic_circle_hole_volume_is_exact`: a 20×20 plate with a true (4-arc) circular hole, extruded into four `CylindricalSurface` walls. It passes a deliberately **coarse** deflection (0.5) and asserts the volume is within 1e-3 of the exact `box − π·r²·h` — only possible if the clamp engages.
- Golden `boolean_box_minus_cylinder` volume improves **719.159 → 717.293** (true **717.257**): a ~50× accuracy gain. The mesh (46v/88t) is unchanged, confirming the change is isolated to volume accuracy.

## Tested

`cargo test` for `check`, `operations`, `io`, `wasm` all green (incl. the boolean/shell volume suites); `cargo fmt` + `clippy` clean; pre-commit hook passed.

### DIFF
--- a/crates/operations/src/extrude/tests.rs
+++ b/crates/operations/src/extrude/tests.rs
@@ -497,6 +497,79 @@ fn extrude_face_with_ccw_circle_hole_volume() {
 }
 
 #[test]
+fn extrude_analytic_circle_hole_volume_is_exact() {
+    use brepkit_math::curves::Circle3D;
+
+    // 20×20 plate with a true (analytic) circular hole, radius 3, extruded by
+    // 10. The hole is four quarter-circle arc edges (how a full circle is
+    // lifted), so the extruded wall is four CylindricalSurface faces and the
+    // volume is exactly box − cylinder = 4000 − π·9·10. Tessellation (and a
+    // chord-polygonised cap) mis-count the curved boundary by ~1-2%; exact
+    // divergence-theorem integration over the analytic faces must hit it.
+    let mut topo = Topology::new();
+
+    let outer_pts = vec![
+        Point3::new(0.0, 0.0, 0.0),
+        Point3::new(20.0, 0.0, 0.0),
+        Point3::new(20.0, 20.0, 0.0),
+        Point3::new(0.0, 20.0, 0.0),
+    ];
+    let outer_wire =
+        brepkit_topology::builder::make_polygon_wire(&mut topo, &outer_pts, 1e-7).unwrap();
+
+    let (cx, cy, r) = (10.0, 10.0, 3.0);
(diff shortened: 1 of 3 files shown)


andymai:
Thanks both — addressed in `0f129c5` (documentation; behaviour unchanged and still green).

**@greptile-apps (P1, clamp local curvature too):** Good catch on the long-solid-with-tiny-feature case. I prototyped the curvature-radius cap (`min curved-face radius · 5e-4`) and it **regresses a real invariant**: `boolean::tests::fuse_coincident_rrect_cap_with_frustum` breaks (`fused 29475.0 != a+b 29473.8`, Δ1.2 > tol 1.0). Reason: a per-solid `min_radius` tessellates a *shared* face differently depending on context — the frustum's cone face is coarse alone (its own large radius) but fine in the fuse (the rrect's small fillet radius dominates), so `volume(A∪B) == volume(A) + volume(B)` no longer holds. The vertex-bbox scale doesn't vary wildly between a part and its fuse, so it preserves that invariant while still fixing the target under-count. The fully-correct answer is **per-face** deflection (each face fine per its own curvature, consistently across contexts), which needs the volume tessellation to drop the watertight whole-solid path — a larger change worth its own PR. Documented the trade-off inline.

**@copilot (vertex-AABB vs true AABB):** Correct — it's the vertex bounding box, not the curvature-inclusive AABB. Since curvature only bulges *outward*, the vertex diagonal under-estimates the extent, which makes the cap conservatively **finer** (safe). Clarified the doc to say so rather than switch to `solid_bounding_box()` (whose extra cost buys nothing here).

Issue #863, opened 2026-06-15
Faithful shape-evolution (persistent face identity through the GFA builder)


andymai:
Umbrella tracking issue for genuinely-unimplemented kernel features that brepjs parity tests exercise (each is a substantial feature, not a bug). From the brepjs divergence registry:

- **Shape evolution tracking** — `shapeRef.*` + `shapeRefIntegration.*` (~11 tests). `fuseWithEvolution`/`cutWithEvolution` exist as wasm exports but the face-identity-tracking semantics aren't wired through. Largest single subsystem.
- **variableFillet** — present but broken (`modifierFns.variableFilletRadius/Callback` produce physically-impossible volume > 1000).
- **multiSweepFns** — multi-profile pipe.
- **guidedSweepFns** — auxiliary-spine guided sweep.
- **interferenceFns** — interference detection.
- **hullFns** — convex hull.
- **minkowskiFns** — Minkowski sum.

Each warrants its own issue/PR when picked up; listed here so the parity backlog is visible. Not blocking the gridfinity north-star (those use primitives/booleans/sweep/fillet/shell, mostly covered).

**Note:** the raw-OCCT-API divergences (`*.nullShapeValidation`, `cast.*`, `meshFns.stepReadError`, `occtBoundary`, `disposal`, `batchOps.cacheReset`) are NOT real gaps — they exercise an `oc` object brepkit deliberately doesn't expose; leave skipped.

andymai:
## Grooming — 5 of 7 items are already implemented; re-scope to shape-evolution

Audited each item against current `main` + the brepjs adapter. Most of this list is stale.

### Already implemented & exported (recommend striking from this tracker)
| Item | brepkit location | Notes |
|------|------------------|-------|
| `hullFns` (convex hull) | `wasm shapes.rs:716` → `primitives.rs:721` | ungated brepjs `convexHullFns.test.ts` |
| `minkowskiFns` | `wasm operations.rs:1279` → `primitives.rs:800` | brepjs `minkowski()` built on `hullFromPoints` |
| `multiSweepFns` | `sweep.rs:1915` (`make_section_sweep`) | passing Rust tests + ungated brepjs `multiGuidedSweepNative.test.ts` |
| `guidedSweepFns` | `sweep.rs:2068` (`sweep_guided`) | same suite |
| `interferenceFns` | — | **not a kernel feature**: pure-TS brepjs `checkInterference` over `solidToSolidDistance` (`measure.rs:241`) |

[...]

Issue #801, opened 2026-06-14
booleans: fuse drops coincident-face operand — (a−b) ∪ (a∩b) loses vol(a∩b) (2.103.1)


andymai:
## Summary

`fuse` of two solids that share coincident boundary faces silently drops one operand. The classic CSG identity `(a − b) ∪ (a ∩ b) === a` fails: the result equals `a − b`, losing exactly `vol(a ∩ b)`.

Discovered via brepjs parity suite (`tests/parity/booleans.test.ts` → `INVARIANT: cut-then-fuse-back recovers original volume`) against `brepkit-wasm@2.103.1`. `cut` and `intersect` are each individually correct; only the recombining `fuse` is wrong.

## Minimal repro

Two axis-aligned cubes sharing the origin corner, `b` fully nested in `a`:

```
a = box(2, 2, 2)          // vol 8
b = box(1, 1, 1)          // vol 1, nested in a, shares 3 faces at the corner

aMinusB   = cut(a, b)        // vol 7   ✓ correct
aAndB     = intersect(a, b)  // vol 1   ✓ correct
recombined = fuse(aMinusB, aAndB)
```

| Step | brepkit-wasm 2.103.1 | OCCT (occt-wasm / opencascade.js) |
|------|----------------------|-----------------------------------|
| `cut(a, b)` | vol **7** ✓ | vol 7 |
| `intersect(a, b)` | vol **1** ✓ | vol 1 |
| `fuse(aMinusB, aAndB)` | vol **7** ✗ (want 8) | vol **8** ✓ |

The fused result discards the `a ∩ b` operand entirely instead of merging it across the shared cut-plane faces.

## Scaling — error is always exactly `vol(a ∩ b)`

| a | b (nested, dx=0) | recombined (brepkit) | expected | error |
|---|------------------|----------------------|----------|-------|
| box(2) | box(1) | 7 | 8 | 1 = 1³ |
| box(3) | box(1) | 26 | 27 | 1 |
| box(10) | box(5) | 875 | 1000 | 125 = 5³ |
| box(3) | box(2) @dx=1 | 19 | 27 | 8 = 2³ |
| box(10) | box(5) @dx=2.5 | 875 | 1000 | 125 |

The dropped volume is always exactly the second `fuse` operand's volume, confirming `fuse` returns only the first operand when the two share coincident boundary faces.

## Notes

[...]

PR #962, opened 2026-06-23
feat(algo): faithful shape-evolution via GFA face provenance

andymai:
## Summary

Implements the one genuinely-remaining item in #863: **faithful shape-evolution** (persistent face identity through booleans), replacing the geometry heuristic that left the ~12 brepjs `shapeRef`/`shapeRefIntegration` tests skipped.

`boolean_with_evolution` previously matched each output face back to an input by **normal + centroid similarity** — fine for a single fillet, but unable to resolve ambiguous multi-step splits. Now each result face carries its **true input source**, threaded through the GFA.

## How the provenance flows

```
fill_images_faces: SubFace.source_face = the input face being split
        │
BOP select_faces: SelectedFace.source_face forwarded (kept rep for SD pairs)
        │
build_solid_with_origins: (result face → source) pairs, a vector PARALLEL to
        face_ids — in-place edge/vertex rebuilds preserve position; only the
        length-changing steps (sliver retain, doubled-face removal, cap
        synthesis) sync explicitly. **Purely additive: face_ids handling is
        unchanged, so the assembled geometry is identical.**
        │
GfaShapeStore (deep_copy_solid now returns its source→target face map):
        input remap retained + result remap exported
        │
gfa::boolean_with_face_origins: translates store-local provenance to CALLER
        face indices (the convention EvolutionMap uses)
        │
boolean_with_evolution: builds EvolutionMap (modified / deleted)
```

The faithful path applies the **id-preserving** result heals (`remove_degenerate_edges` + `remove_wire_spurs` for #801) so the result is as correct as the standard boolean; `unify_faces` is intentionally skipped — it merges coplanar faces into new entities, discarding per-face provenance. Non-GFA paths (identical operands) fall back to the geometry heuristic.

## Validation

New `fuse_evolution_is_faithful`: an overlapping two-cube fuse — **every** result face traces to a real input face (no geometry guessing). Full `algo`, `operations`, `io`, `wasm` suites green; `clippy`/`fmt` clean.

The wasm `fuseWithEvolution` / `cutWithEvolution` / `intersectWithEvolution` bindings pick this up automatically. The brepjs `shapeRef` un-skip + the stale `kernelDivergences.ts` reason refresh follow once `brepkit-wasm` releases and re-syncs.

## Known follow-ups (documented on #863)

- Same-domain merges record the kept representative's source only (the merged-away operand maps to `deleted`); dual-origin tracking is a refinement.
- A face synthesised with no input (rare partial-overlap cap) is left unattributed (`None`).

Refs #863

### DIFF
--- a/crates/algo/src/bop.rs
+++ b/crates/algo/src/bop.rs
@@ -124,6 +124,7 @@ pub(crate) fn select_faces(
             if keep {
                 Some(SelectedFace {
                     face_id: sf.face_id,
+                    source_face: sf.source_face,
                     reversed: op == BooleanOp::Cut && sf.rank == Rank::B,
                 })
             } else {
@@ -181,6 +182,7 @@ fn apply_sd_selection(
             if op == BooleanOp::Cut {
                 selected.push(SelectedFace {
                     face_id: sf_a.face_id,
+                    source_face: sf_a.source_face,
                     reversed: false,
                 });
                 continue;
@@ -215,6 +217,7 @@ fn apply_sd_selection(
             };
             selected.push(SelectedFace {
                 face_id: sub_faces[keep].face_id,
+                source_face: sub_faces[keep].source_face,
                 reversed: false,
             });
         } else {
@@ -242,6 +245,10 @@ fn apply_sd_selection(
 pub(crate) struct SelectedFace {
     /// The topology face to include.
     pub face_id: brepkit_topology::face::FaceId,
+    /// The original input face this selection derives from (shape-evolution
+    /// provenance). For a same-domain pair it is the kept representative's
+    /// source.
+    pub source_face: brepkit_topology::face::FaceId,
     /// Whether to reverse this face's orientation in the result.
     pub reversed: bool,
 }
@@ -277,8 +284,10 @@ mod tests {
 
     /// Helper to create a SubFace for testing.
     fn make_sub_face(topo: &mut Topology, rank: Rank, classification: FaceClass) -> SubFace {
+        let fid = dummy_face_id(topo);
         SubFace {
-            face_id: dummy_face_id(topo),
+            face_id: fid,
+            source_face: fid,
             classification,
             rank,
             interior_point: None,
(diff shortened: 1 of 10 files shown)


andymai:
Thanks @copilot — all four addressed in \`7833b2d\`:

1. **\`build_solid_with_origins\` `unwrap_or_default`** → now propagates the \`solid_faces\` error with \`?\`. A miss there is a real assembly/provenance bug and should surface, not return a partial map.
2. **\`boolean_with_face_origins\` silent `filter_map`** → a result face missing from the export map is a provenance desync and now returns an \`AlgoError\` (the export map is total over result faces). An unresolvable *source* still maps to \`None\` — that is by design (a synthesised face with no input origin), not a dropped input.
3. **No result validation on the faithful path** → it now runs \`validate_boolean_result\` and, if the GFA result is invalid, **falls through** to \`boolean()\`'s full pipeline (fast paths + mesh fallback + validation), so it honours the same correctness contract as \`boolean\`.
4. **Docstring vs behaviour** → corrected. The faithful path intentionally prefers the GFA for *every* `a != b` pair (that's where provenance lives), so it can diverge from \`boolean\`'s fast paths; the doc now says exactly that, plus the validate-or-fall-back semantics. Running the full GFA for a provenance query (rather than an AABB shortcut) is an acceptable trade for faithful tracking.

`clippy`/`fmt` clean; `algo` (137) + `operations` boolean (113, incl. `fuse_evolution_is_faithful`) green.

andymai:
Thanks @greptile — all three addressed in \`ad3cad0\`:

1. **Heal errors propagate after topology mutation** → the heals are now `is_ok()`-gated; a heal failure (like an invalid result) **falls through** to `boolean()`'s full pipeline instead of propagating. The orphaned result solid is harmless dead topology in the arena.
2. **No caller-observable signal on degradation** → the heuristic fallback now emits a `log::debug` noting faithful provenance was unavailable. (A typed faithful-vs-approximate signal would change the public return type — left for a maintainer call; the log covers observability for now.)
3. **`.get(fi).copied().flatten()` indirection** → replaced with `sources[fi]`. `sources` is parallel to `face_ids` so `fi` is always in bounds; direct indexing makes the invariant explicit and panics loudly in debug if it's ever violated, rather than silently mislabelling a tracked face as a synthesised cap.

`clippy`/`fmt` clean; `fuse_evolution_is_faithful` + the boolean suite green.

Issue #967, opened 2026-06-23
Constant-radius fillet on a cylinder's circular edge removes ~37–47% of the solid (should round the rim only)


andymai:
## Summary

A constant-radius `fillet` applied to the circular rim edge(s) of a plain cylinder produces a grossly wrong solid in brepkit. Even a tiny fillet radius (r=0.5 on a 6283 mm³ cylinder, which should remove well under 1% of material) collapses the volume by ~37%. A box fillet of the same radius is correct, so the bug is specific to **circular (cylindrical) edges**.

occt-wasm and occt (OpenCascade.js) both produce correct, physically sensible filleted cylinders.

## Numbers

Cylinder radius 10, height 20 (raw volume = π·100·20 = **6283.185 mm³**). Fillet all edges:

| radius | occt-wasm / occt (correct) | brepkit (wrong) | brepkit error |
| ------ | -------------------------- | --------------- | ------------- |
| 0.5    | 6276.519                   | **3978.824**    | −36.6%        |
| 1.0    | 6256.820                   | **3769.412**    | −39.8%        |
| 2.0    | 6180.134                   | **3350.585**    | −45.8%        |

Single circular edge, r=2: occt-wasm = **6231.659**, brepkit = **3769.297** (−39.5%).

The filleted topology also differs: occt-wasm yields 5 faces / 7 edges (area 1773.53), brepkit yields 3 faces / 3 edges (area 1004.98). Both kernels report the result as `valid=true`, so this is a silent geometry error, not a failed operation.

A fillet of radius r on a rim can only remove material from a band of width ≤ r around the edge; removing nearly half the cylinder is physically impossible. brepkit appears to mis-resolve the blend against the cylindrical face rather than rounding the rim.

## Minimal repro (brepjs public API)

```ts
import { cylinder, fillet, measureVolume, unwrap } from 'brepjs';

const cyl = cylinder(10, 20);             // volume 6283.185
const filleted = unwrap(fillet(cyl, 0.5)); // round both circular rims, r=0.5
console.log(unwrap(measureVolume(filleted)));
// occt-wasm/occt: 6276.519  (correct — tiny rim round)
// brepkit:        3978.824  (wrong — ~37% of the solid gone)
```

## Notes
- Box fillet works on brepkit (a 10-cube filleted r=1 gives 975.3, matching occt-wasm's 975.6), so the regression is isolated to filleting cylindrical/curved edges, not fillets in general.
[...]

PR #971, opened 2026-06-23
fix(fillet): reject degenerate cylinder-rim fillet instead of silent volume corruption

andymai:
## Summary

Filleting a plain cylinder's circular rim silently removed **37–47% of the solid** (#967) — a *valid-but-wrong* result that every cheap check accepted. A cylinder/cone rim is a **closed** circular edge (start vertex == end vertex), unlike a box's open line edges, which triggered two layered defects:

1. The deprecated **rolling-ball** engine (tried first) assumes polygonal wires with distinct corner vertices. On a closed circular edge it collapses the blend strip and the disc caps to **zero-area faces** while staying manifold — shrinking the wall and dropping the caps (−37% at r=0.5).
2. The walking **fillet_v2** engine (fallback) can't trim a disc cap to an interior contact circle, so it leaves an **open shell** (free edges) that tessellates to a plausible-but-wrong volume.
3. The dispatch gate `validate_shell_manifold` only forbids edges shared by 3+ faces, so it accepted **both**.

## Fix

- **`rolling_ball.rs`** — reject an assembly with any near-zero-area face (keyed on area, not topology, so a correct fillet is never rejected).
- **`wasm/helpers.rs`** — `try_fillet` now requires a **closed** shell (`validate_shell_closed`), not just a manifold one, rejecting open-shell results.
- **`fillet/tests.rs`** — corrected two tests that asserted the buggy rolling-ball output (whose "NURBS faces" were zero-area).

With both broken engine outputs rejected, `try_fillet` falls back to returning the input unchanged: the cylinder rim fillet is now a **valid no-op** (raw volume `6283.185`, **0% change**, within 0.1% of occt's `6276.519`) instead of silently-corrupt geometry. New regression test `try_fillet_cylinder_rim_does_not_corrupt_volume`.

## Scope — this stops the corruption, it does NOT yet round the rim

This fix eliminates the **reported defect** (the silent 37–47% volume loss). It does **not** implement genuine rim rounding (occt removes ~0.1% and yields 5 faces). That is a separate blend-engine **feature**, blocked on:
- `blend/src/analytic.rs::plane_cylinder_fillet` keying the blend side only on `is_reversed()` — wrong for a bare disc cap; the inward geometry is `major = r_c − r` (derived + prototyped), and
- `FilletBuilder`'s line-based trimmer being unable to rebuild a disc cap bounded by an interior contact circle.

Left as a follow-up; `Refs #967` (not `Closes`) so the rounding feature stays tracked.

## Verification

- `cargo test -p brepkit-operations -p brepkit-wasm -p brepkit-blend` — **1236 pass, 0 fail** (incl. the gridfinity fillet suite and the box `try_fillet` tests, proving the stricter gate rejects no legitimate fillet)
- clippy clean; pre-commit hooks passed

Refs #967

### DIFF
--- a/crates/operations/src/fillet/rolling_ball.rs
+++ b/crates/operations/src/fillet/rolling_ball.rs
@@ -2155,6 +2155,30 @@ pub fn fillet_rolling_ball(
     // the mesh boolean fallback on moderate-complexity filleted solids.
     let _ = crate::heal::unify_faces(topo, solid_id);
 
+    // Reject a geometrically-degenerate assembly. This engine is built around
+    // polygonal wires with distinct corner vertices; a closed circular edge
+    // (a cylinder/cone rim, where start vertex == end vertex) collapses its
+    // blend strip and the adjacent disc cap to zero-area faces while still
+    // passing the manifold check. Such a result is silently wrong (it drops
+    // the caps and shrinks the wall), so surface it as an error and let the
+    // caller fall through to the walking blend engine, which handles closed
+    // spines. The guard is keyed on actual face area, not edge topology, so a
+    // valid fillet (every face has real area) is never rejected.
+    let faces = brepkit_topology::explorer::solid_faces(topo, solid_id)
+        .map_err(crate::OperationsError::Topology)?;
+    let area_floor = (radius * radius) * 1e-6;
+    for fid in faces {
+        let area = crate::measure::face_area(topo, fid, radius * 0.1)?;
+        if area <= area_floor {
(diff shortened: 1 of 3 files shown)


andymai:
Superseded by #972, which actually rounds the cylinder rim into an exact quarter-torus (watertight 5-face solid, ~0.12% rim round matching occt) rather than turning the corruption into a validated no-op. #972 includes this PR's defensive guards (rolling-ball zero-area rejection + the `try_fillet` closed-shell gate) plus the genuine fix.

PR #974, opened 2026-06-23
feat(operations): support non-planar profiles in loft

andymai:
## What

Lifts the planar-only restriction on `loft` and `loft_smooth`. A profile may now lie on a cylinder, cone, sphere, torus, or NURBS surface — not just a plane.

This is **PR1 of 3** for the README's only *Planned* feature (Sweeps → *"Non-planar profiles for revolve, sweep, loft, pipe"*). PR1 establishes the reusable cap pattern; PR2 covers sweep + pipe, PR3 covers revolve. The README Status row and Known-Limitations bullet are updated with the final PR (they describe all four ops as one row).

## Why it's small

The loft side-face machinery (`face_polygon`, the `winding` helpers, the ruled side surfaces) was already 3D-point based and planarity-agnostic. The only blockers were:
1. an explicit gate `match face.surface() { Plane => {}, _ => Err(...) }`, and
2. the flat `FaceSurface::Plane` end caps.

## Approach

**The cap *is* the section.** brepkit profiles are always faces, so every profile already carries a real surface. Each end cap now reuses the section's own surface (`surface().clone()`) instead of reconstructing a plane — so a curved section stays curved. No new filling primitive is needed.

- **Planar profiles are byte-for-byte unchanged.** A planar profile keeps the existing Newell-corrected `Plane` cap (the stored profile normal can't be trusted, so the Newell recompute is preserved) and is never reversed. Locked by a regression test.
- **Curved caps** pick their `Face` `reversed` flag by probing the surface normal at the UV projections of two boundary vertices against the outward (away-from-body) direction. The probe uses real boundary vertices rather than a fixed `(0.5, 0.5)` because a cap can occupy any sub-region of its parent surface.
- New `build_cap_face` helper is written to be lifted into a shared module in PR2/PR3, whose end caps will pass a rigidly transformed copy of the profile surface (`transform_face` already handles every surface variant).

Holes in profiles remain ignored (unchanged); resampled curved boundaries are a chord approximation of the true section (documented).

## Tests

- `loft_planar_caps_unchanged` — regression: planar caps stay `Plane`, ±Z, never reversed.
- `loft_two_nonplanar_coons_patches_is_valid_solid` — non-coplanar Coons sections → NURBS caps, `validate_solid` valid, bounded volume.
- `loft_sphere_cap_profiles_keep_analytic_caps` — analytic (`Sphere`) caps; positive-volume assertion guards cap orientation (a wrong `reversed` flips the divergence-theorem volume sign).
- `loft_smooth_three_nonplanar_patches_positive_volume` — covers the `loft_smooth` cap path.

All 22 loft tests + the full `brepkit-operations` suite (717 lib + integration) pass; clippy `-D warnings` clean; layer boundaries valid.

### DIFF
--- a/crates/operations/src/loft.rs
+++ b/crates/operations/src/loft.rs
@@ -82,19 +82,123 @@ fn cap_normal_from_verts(verts: &[Point3], inward: bool) -> Result<Vec3, crate::
     Ok(if inward { unit * -1.0 } else { unit })
 }
 
-/// Loft two or more planar profiles into a solid.
+/// A cap ring whose vertices deviate from their best-fit plane by less than
+/// this fraction of the ring's size is treated as planar (capped by an exact
+/// `Plane`); a larger deviation is filled by a bilinear patch.
+const CAP_PLANARITY_TOL: f64 = 1e-6;
+
+/// Characteristic size of a cap ring: the largest distance from its first
+/// vertex to any other, used to scale the planarity test.
+fn ring_scale(verts: &[Point3]) -> f64 {
+    let c = verts[0];
+    verts.iter().map(|p| (*p - c).length()).fold(0.0, f64::max)
+}
+
+/// Bilinear (degree-1) NURBS patch through a 4-corner ring, in ring order.
 ///
-/// Each profile is a planar face. All profiles must have the same
-/// number of boundary vertices. The loft connects corresponding
-/// vertices between adjacent profiles with ruled (linear) surfaces,
-/// and caps the first and last profiles as the solid's end faces.
+/// Its four boundary iso-curves are the straight segments between consecutive
+/// corners — exactly the ring's chord edges — so the cap face shares its
+/// boundary with the side faces and tessellates/integrates clipped to the
+/// section. (Reusing the section's own parent surface would not: brepkit
+/// tessellates a non-planar face over its full u/v extent, not clipped to a
+/// chord-polygon wire, so a reused surface overfills past the section.)
+fn bilinear_cap_patch(corners: &[Point3]) -> Result<NurbsSurface, brepkit_math::MathError> {
+    NurbsSurface::new(
+        1,
+        1,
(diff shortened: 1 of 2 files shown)


andymai:
Thanks — the T-Rex repro on the NURBS overfill was the key catch. I dug into the tessellator and confirmed the root cause is broader than NURBS:

- `tessellate_nurbs` grids the full `domain_u × domain_v` (never consults the wire), and
- `tessellate_analytic` (cylinder/cone/sphere/torus) fills the surface's u/v **extent** from `compute_*_range`, not the chord-polygon wire.

So **only `Plane` faces clip to the polygon** (via CDT). Reusing a section's parent surface as a cap is therefore unsound for a faceted loft ring for *every* non-planar surface type, not just NURBS — the cap fills the natural surface region and overfills past the section.

I've reworked the approach to **fill the section boundary** (standard thru-sections) instead of reusing the surface:

- Planar section boundary → exact `Plane` cap (CDT-clipped; planar profiles are byte-for-byte unchanged).
- Non-planar 4-sided boundary → a **bilinear (degree-1) NURBS patch through the ring corners**, whose four boundary iso-curves *are* the ring chord edges, so `domain == wire` and it can't overfill.
- Non-planar boundary with >4 edges → rejected (no supported fill yet) rather than approximated.

This also dissolves the two probe concerns: there's no longer any analytic `project_point`/boundary-normal probing (so no cancellation and no equatorial ⊥-to-outward ambiguity) — the only orientation check is the stable center-normal of a near-flat bilinear lid.

New/updated tests: planar-cap regression guard; non-planar Coons-patch loft (bilinear caps, volume bounded near the prism to catch overfill); non-planar-surface/planar-boundary loft (exact flat caps, **exact** prism volume = 180); >4-edge non-planar rejection; smooth multi-section non-planar loft. Full `brepkit-operations` suite + clippy `-D warnings` + layer boundaries all green.

PR #976, opened 2026-06-23
feat(operations): support non-planar profiles in sweep and pipe

andymai:
## What

**PR2 of the non-planar sweep-family series** (PR1 = loft, [#974](https://github.com/andymai/brepkit/pull/974), merged). Extends non-planar profile support to the basic `sweep` and `pipe`, and lifts the cap logic into a shared `crate::cap` module that loft now also uses.

## Approach (same as PR1: fill the boundary)

The end caps are built by filling the section **boundary**, never by reusing the section's parent surface — brepkit tessellates/integrates a non-planar face over its full u/v extent rather than clipping to a chord-polygon wire, so a reused surface overfills. So:

- **planar ring** → exact `Plane` cap (CDT-clipped; planar profiles unchanged),
- **non-planar 4-sided ring** → bilinear patch whose four iso-curves *are* the ring chords (`domain == wire`, can't overfill),
- **non-planar >4 edges, or holes on a non-planar cap** → unsupported (clear error).

For sweep/pipe the cap's `outward` normal is the ring's Newell normal oriented along the path tangent, and the frame up-hint is derived from the section boundary (replacing the old `match surface { Plane }` gate). A planar perpendicular profile yields byte-for-byte the same `-tangent` cap as before.

`crate::cap::build_cap_face` threads pre-built inner-wire (hole) loops; holes are supported on planar caps, rejected on non-planar ones.

## Scope

Basic `sweep` and `pipe` (the canonical entry points). The specialised sweep variants — `sweep_smooth`, `sweep_with_options`, `sweep_miter`, `multi_section_sweep` — **keep their planar gate** for now: `sweep_smooth`'s NURBS rail interpolation (rails skip the middle rings) degrades to a ~3× volume undercount on a non-planar profile, which needs separate work. Revolve is the next PR. The README row is updated when the series completes.

## Tests

- `sweep_planar_profile_caps_are_planar` — planar regression.
- `sweep_nonplanar_saddle_profile_is_valid_solid` — non-planar (saddle) sweep → `validate_solid` valid, bilinear caps, volume bounded near the prism (overfill would fail).
- `pipe_nonplanar_saddle_profile_is_valid_solid` — same for pipe.

Full `brepkit-operations` suite (incl. loft's 23, sweep's 41, pipe's 8) green; clippy `-D warnings` clean; layer boundaries valid; `brepkit-wasm` builds.

### DIFF
--- a/crates/operations/src/cap.rs
+++ b/crates/operations/src/cap.rs
@@ -1,0 +1,163 @@
+//! Shared loft/sweep/pipe/revolve end-cap construction.
+//!
+//! A swept/lofted/revolved end is closed by filling its section *boundary* — a
+//! ring of chord edges — rather than by reusing the section's own surface.
+//! brepkit tessellates and integrates a non-planar face over its full u/v
+//! extent rather than clipping to a chord-polygon wire, so a reused parent
+//! surface would overfill past the section. Instead: a planar ring gets an exact
+//! `Plane` cap (the planar tessellator clips it to the polygon), and a
+//! non-planar 4-sided ring is filled by a bilinear patch whose boundary
+//! iso-curves are exactly the ring chords (`domain == wire`, so it can't
+//! overfill).
+
+use brepkit_math::nurbs::surface::NurbsSurface;
+use brepkit_math::vec::{Point3, Vec3};
+use brepkit_topology::Topology;
+use brepkit_topology::edge::EdgeId;
+use brepkit_topology::face::{Face, FaceId, FaceSurface};
+use brepkit_topology::vertex::VertexId;
+use brepkit_topology::wire::{OrientedEdge, Wire, WireId};
+
+use crate::dot_normal_point;
+
+/// Collect the 3D positions of a ring of vertices (a cap's outer boundary).
+///
+/// # Errors
+///
+/// Returns an error if a vertex id is missing from the arena.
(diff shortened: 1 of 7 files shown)


andymai:
Thanks — addressed all six:

1. **`build_cap_face` panic safety** — added an up-front guard: returns `InvalidInput` when `cap_verts` is empty or its length doesn't match `outer_ring_edges` (was previously relying on callers).
2. **Sweep regression test bypassing the cap path** — great catch; `straight_z_path` short-circuited to `try_straight_extrude`. The test now uses a curved degree-2 path whose initial tangent is still +Z, so the general sweep (and the new cap code) actually runs.
3. **`sweep` rustdoc** — updated to drop "planar profile" / the "not planar" error and document the new cap behavior + limitations (non-planar caps only for hole-free 4-edge rings).
4. **`pipe` rustdoc** — same update to the summary and `# Errors`.
5. **Duplicated `make_saddle_profile`** — moved to `crate::test_helpers` (cfg(test)); sweep and pipe now share the single source.
6. **Missing pipe planar regression** — added `pipe_planar_profile_caps_are_planar` (pipe has no straight-extrude fast path, so it exercises the cap code directly).

Full `brepkit-operations` suite green, clippy `-D warnings` clean, layer boundaries valid.

PR #979, opened 2026-06-23
feat(operations): support non-planar profiles in revolve

andymai:
## What

**PR3 — the capstone of the non-planar profile series** (loft [#974](https://github.com/andymai/brepkit/pull/974), sweep+pipe [#976](https://github.com/andymai/brepkit/pull/976), both merged). Lifts revolve's planar-only gate and updates the README to reflect the finished series.

## Approach

Revolve only uses the profile's **boundary**, so the surface-type gate was artificial. Now:

- A **planar** profile keeps its stored plane normal, corrected to the CCW convention — **byte-for-byte unchanged** (the existing CW-winding path is preserved).
- A **non-planar surface** profile has its profile normal derived from the boundary's Newell normal.

Revolve is the structurally different member of the series: its caps reuse the profile's **own wire** (with per-edge orientations, possibly arcs), so the shared `cap::build_cap_face` (which builds a fresh forward-ordered chord ring) doesn't apply. Instead:

- A **full** revolution has no caps → accepts **any** boundary.
- A **partial** revolution closes its ends with planar caps → still requires a **planar** boundary; a non-planar boundary on a partial revolve is rejected with a clear error.

## README

The non-planar-profiles Status row moves **Planned → Beta**; the Known-Limitations and Roadmap entries now describe the remaining gaps (advanced sweep variants — smooth/guided/miter/multi-section — still planar-only; section boundaries with >4 edges; partial revolutions of non-planar boundaries).

## Tests

- `revolve_nonplanar_surface_full_turn_volume` — cylinder-surfaced rectangle revolved 2π → 36π (matches the planar case).
- `revolve_nonplanar_surface_partial_volume` — same profile revolved π → 18π (planar caps, exact).
- `revolve_partial_nonplanar_boundary_is_rejected` — a genuinely non-planar boundary on a partial revolve errors.

All 17 revolve tests + full `brepkit-operations` suite green; clippy `-D warnings` clean; layer boundaries valid; `brepkit-wasm` builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ brepkit is in active development. Core modeling is solid. Each feature below is 
 | **Sweeps**              | Extrude (planar + NURBS profiles)                                            | Stable       |
 | **Sweeps**              | Revolve, sweep, loft, pipe (planar profiles)                                 | Stable       |
 | **Sweeps**              | Helical sweep                                                                | Stable       |
-| **Sweeps**              | Non-planar profiles for revolve, sweep, loft, pipe                           | Planned      |
+| **Sweeps**              | Non-planar profiles for loft, sweep, pipe, revolve                           | Beta         |
 | **Construction**        | Coons-patch face fill, sew, untrim                                           | Stable       |
 | **Sectioning**          | Cross-section faces, split by plane                                          | Stable       |
 | **Measurement**         | Bounding box, area, volume, center of mass                                   | Stable       |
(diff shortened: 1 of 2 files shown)


andymai:
Thanks — addressed all four:

1. **Full revolve rejecting 1–2 edge non-planar profiles** — good catch. The boundary-vertex-count check (and the Newell/planarity derivation) now only runs for *partial* revolutions. A full revolution has no caps and the profile normal is unused, so it accepts any boundary; the normal is derived best-effort with an axis fallback when the wire is degenerate.
2. **New `_ => None` wildcard on `FaceSurface`** — enumerated the variants explicitly (`Cylinder | Cone | Sphere | Torus | Nurbs => None`), so a future surface variant fails at the compiler instead of silently taking the `None` path (per the project's match-arm convention).
3. **Magic `1e-6`** — hoisted to a documented `const PLANARITY_REL_TOL` alongside `MIN_RADIAL_LEN`.
4. **Untested full-revolve non-planar boundary** — added `revolve_full_nonplanar_boundary_is_accepted` (the twisted quad from the rejection test, revolved 2π → valid, positive volume), locking in the "full revolution accepts any boundary" guarantee.

Full `brepkit-operations` suite green (18 revolve tests), clippy `-D warnings` clean, layer boundaries valid.

PR #981, opened 2026-06-23
fix(operations): correct sweep_smooth side-face rails and orientation

andymai:
## What

Fixes the `sweep_smooth` bug flagged during the non-planar sweep work ([#976](https://github.com/andymai/brepkit/pull/976)): it produced **non-manifold solids whose volume integrated to ⅓ of the true value** — for *every* profile, not just non-planar ones. The existing tests only asserted `vol > 0`, so it slipped through.

## Root cause (two compounding defects)

1. **Duplicated rails.** Each side face built its *own* left/right rail edge between the same two ring vertices, so adjacent faces never shared the rail → the shell was non-manifold (every rail edge belonged to a single face).
2. **Inward side-face normals.** The NURBS side surface's natural normal `∂u×∂v = path×edge` points *into* the swept body for a CCW profile, but the face was added without the `reversed` flag → the sides contributed inward-facing flux and cancelled most of the volume (the telltale ⅓).

A diagnostic made it concrete — a 2×2 square swept 6 along +Z gave **volume 8, invalid**; expected 24.

## Fix

- Build **one shared rail edge per profile vertex**, reused by both adjacent side faces → manifold shell.
- Set the face `reversed` flag when the surface normal opposes the geometric outward (`edge×path`), probed at a consistent `(u=0, v=0.5)` location so it stays correct as a curved path twists the normal along `u`.

## Result

- Straight 2×2×6 prism: **exact 24, valid** (was 8, invalid).
- Gently curved path: correct volume (~25) and valid.
- New regression tests assert both (the exact straight-path volume catches both the ⅓ and the invalidity).

## Out of scope (pre-existing, noted)

A path that turns sharply *toward* the profile's up-vector (e.g. a 90° quarter circle in the plane of the profile normal) still degenerates to ~0 volume — but **basic `sweep` does too**, so that's a separate pre-existing sweep degeneracy, not a rail issue. This fix neither introduces nor resolves it.

This unblocks adding non-planar profile support to `sweep_smooth` (the rail bug was the blocker) as a future follow-up.

Full `brepkit-operations` suite green (43 sweep tests); clippy `-D warnings` clean; layer boundaries valid.

### DIFF
--- a/crates/operations/src/sweep.rs
+++ b/crates/operations/src/sweep.rs
@@ -991,6 +991,13 @@ pub fn sweep_smooth(
     let degree_u = (num_rings - 1).min(3);
     let degree_v = 1;
 
+    // Rail edges, one per profile vertex (the swept-vertex path from the first
+    // to the last ring), shared between the two adjacent side faces so the shell
+    // is manifold.
+    let rail_edges: Vec<_> = (0..n)
+        .map(|i| topo.add_edge(Edge::new(first_ring[i], last_ring[i], EdgeCurve::Line)))
+        .collect();
+
     for i in 0..n {
         let next_i = (i + 1) % n;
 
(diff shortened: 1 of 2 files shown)


andymai:
Thanks — addressed all three:

1. **`path_dir` / probe location mismatch** — now both `outward` and the surface-normal probe are taken at the start ring, **mid-edge** (`v=0.5`): `path_dir` is the difference of the edge *midpoints* between ring 0 and ring 1, crossed with the edge tangent, matching `surface.normal(0.0, 0.5)`.
2. **Zero `outward` when edge ∥ path** — added a fallback: if `edge_dir × path_dir` is below tolerance, `outward` falls back to the radial direction from the ring centroid (`mid_edge − centroid`), so the reversal decision is never made on a zero vector.
3. **Curved-path assertion too broad** — tightened from `(22, 30)` to `(24, 26)` around the measured ~25, so it no longer passes on a partial fix.

Full `brepkit-operations` suite green (43 sweep tests), clippy `-D warnings` clean, layer boundaries valid.

PR #988, opened 2026-06-23
feat(operations): non-planar profiles in smooth, options, and multi-section sweeps

andymai:
## What

Extends non-planar profile support (loft [#974](https://github.com/andymai/brepkit/pull/974), sweep/pipe [#976](https://github.com/andymai/brepkit/pull/976)) to **three of the four advanced sweep variants**, using the shared `crate::cap` boundary-fill caps. Unblocked by the [sweep_smooth rail fix](https://github.com/andymai/brepkit/pull/981) and [sweep auto-orientation](https://github.com/andymai/brepkit/pull/985).

## Changes

- **`sweep_smooth`, `sweep_with_options`** — drop the planar gate, derive the frame up-hint from the section boundary's Newell normal, and route the end caps through `cap::build_cap_face` (planar ring → exact `Plane`; non-planar 4-edge ring → bilinear patch whose iso-curves are the ring chords). This is the identical conversion already shipped for the basic `sweep`/`pipe`.
- **`multi_section_sweep`** — it places each section perpendicular to the spine (`copy_face` + `transform_face`) and then **lofts** them, and `loft` already handles non-planar profiles. So the only change is `profile_to_frame_matrix`: a planar section keeps its stored plane normal; a non-planar one uses its boundary's Newell normal. Planar behavior is unchanged (the existing CCW-square tests use a stored normal that equals the Newell normal).

## Scope

The **miter-corner** variant (`sweep_miter`) still requires planar profiles — its joint faces lie on the bisector plane between legs and would become non-planar, which needs separate work (a follow-up). The README's Known-Limitations and Roadmap entries are updated to reflect that only miter remains.

## Tests

A non-planar (saddle) profile swept through `sweep_smooth`, `sweep_with_options`, and `multi_section_sweep` — each produces a `validate_solid`-valid solid with the expected ~96 prism volume (the bilinear caps fill the non-planar end rings without overfill).

Full `brepkit-operations` suite green (47 sweep tests); clippy `-D warnings` clean; layer boundaries valid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ A few areas are still maturing. Worth knowing before you build on them:
 
 - **Boolean fallback.** Most booleans run on an exact path that preserves analytic and NURBS surfaces. Hard configurations fall back to a mesh-based boolean: coincident-face contact, coaxial analytic surfaces, razor-thin geometry, or very high face counts. The fallback returns a usable, non-degenerate solid, but it tessellates the curved faces and is not guaranteed watertight.
 - **Torus booleans.** Box-with-torus and coaxial-torus cases work and give correct volumes. General torus-to-torus and torus-with-other-surface intersections have known gaps and may fall back to meshing.
-- **Non-planar profiles.** Loft, sweep, and pipe accept profiles with non-planar surfaces, and close non-planar section boundaries with bilinear caps for four-sided rings (boundaries with more than four edges, or holes on a non-planar section, are not yet supported). Revolve accepts non-planar profile surfaces; a full revolution takes any boundary, but a partial revolution still requires a planar boundary for its caps. The advanced sweep variants (smooth, scaled/guided, miter, multi-section) still require planar profiles.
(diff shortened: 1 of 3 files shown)


andymai:
Thanks — addressed all four; they were all stale doc comments (no code changes):

1. **`sweep_smooth`** — `# Errors` no longer claims it rejects non-planar/holed profiles; now documents the boundary-fill cap behavior and the real cap limitations (non-planar caps only for 4-edge rings; holes only on planar caps).
2. **`sweep_guided`** — dropped "a non-planar profile" from the propagated-error example.
3. **`profile_to_frame_matrix`** — doc now says "a profile face" and notes the planar→stored-normal / non-planar→Newell-normal behavior.
4. **`multi_section_sweep`** — `# Errors` no longer lists "a non-planar profile" as invalid; notes sections may be planar or non-planar (joined by `loft`).

Full `brepkit-operations` suite green (47 sweep tests), clippy `-D warnings` clean, layer boundaries valid.

PR #990, opened 2026-06-24
perf(algo): make Cut producing many holes near-linear (#987)

### LINKED ISSUE #987: perf: Cut producing a face with many holes is O(N²) (honeycomb / perforated panel)
## Summary

A `Cut` whose result has **many holes in a single face** is **O(N²)** in the hole count. Subtracting N disjoint prisms that pass through a slab — i.e. a perforated panel / honeycomb wall — turns the slab's top and bottom faces into one face with N inner wires, and the GFA cost grows quadratically. This is independent of #982 (that fixed `fuseAll`; this is the `cut` / general-fuse path).

Surfaced via gridfinity's honeycomb ("eco") bin walls, which cut a compound of ~300+ hex prisms from the wall and take ~15s.

## Reproduction

Cut a grid of N disjoint prisms out of a slab (each prism pierces the slab, leaving a hole), as one merged multi-region tool — what a compound cutter lowers to:

```rust
// crates/operations, release build
fn slab_with_n_holes(n_grid: usize) -> std::time::Duration {
    use brepkit_math::mat::Mat4;
    let pitch = 2.4;
    let mut topo = Topology::new();
    let span = n_grid as f64 * pitch;
    let slab = make_box(&mut topo, span, span, 2.0).unwrap();          // thin panel
    transform_solid(&mut topo, slab, &Mat4::translation(0.0, 0.0, 1.0)).unwrap();

    let mut tools = Vec::new();
    for j in 0..n_grid {
        for i in 0..n_grid {
            let h = make_box(&mut topo, 1.0, 1.0, 4.0).unwrap();        // pierces the slab
            let x = i as f64 * pitch;
            let y = j as f64 * pitch;
            transform_solid(&mut topo, h, &Mat4::translation(x, y, 0.0)).unwrap();
            tools.push(h);
        }
    }
    let tool = merge_disjoint_solids(&mut topo, &tools).unwrap();       // disjoint multi-region tool
    let t = std::time::Instant::now();
    boolean(&mut topo, BooleanOp::Cut, slab, tool).unwrap();
    t.elapsed()
}
```

## Measurements (release)

`cut(slab, merged N-prism tool)` — N = holes:

| holes (N) | time | growth vs prev |
|----------:|-----:|---------------:|
|  36 |  0.37s | — |
| 100 |  1.6s  | 2.8× holes → 4.3× time |
| 196 |  5.1s  | |
| 324 | 12.8s  | 9× holes → **34× time** |

GFA phase breakdown at N=324 (`gfa::boolean_with_tolerance`):

```
store=4ms   pave=3.67s   build(split+classify)=9.49s   bop+export=65ms
```

Both **PaveFiller** and **Builder** are super-linear; `build` dominates. `store`/`bop`/`export` stay negligible.

## Why decomposition doesn't help

Cutting the prisms in spatial batches instead of one merged tool is **worse**, because the *result* slab carries N holes regardless and each batch reprocesses the increasingly-holed blank:

| strategy (N=324) | time |
|---|---:|
| one merged cut | 13.4s |
| batches of 64 | 27.1s |
| batches of 36 | 27.5s |
| batches of 16 | 32.0s |

So the cost is driven by the **hole count on the result face**, not by how the tool is grouped — it can't be worked around at the call site.

## Likely root cause

The slab's planar top/bottom face ends up with N inner wires. The per-face work that's quadratic in inner-wire count is the suspect — wire nesting / containment ordering, pave-block construction on that face, and/or per-fragment point-in-face classification not spatially indexed over the holes. A spatial index (BVH/grid) over a face's inner wires for nesting + classification would likely bring this to ~O(N log N).

## Context

- brepkit-wasm 2.119.2, via gridfinity-layout-tool's honeycomb wall pattern (brepjs `cutAllBisect(bin, hexCompound)`).
- Distinct from #982. Not a correctness bug — output is correct, just slow.

andymai:
## Summary

Fixes #987 — a `Cut` whose result has many holes in a single face (a perforated / honeycomb panel: N disjoint prisms cut through a slab) was **O(N²)** in the hole count. Profiling found **five independent quadratic hot paths**, each the same shape — an unindexed *"against all others"* scan, or per-face geometry rebuilt on every query. Each is replaced with a spatial index or a cheap broad-phase bound. Every fix is **behavior-preserving**: face count, volume, and manifoldness are bit-identical at every N.

**N=324 holes: 2.62s → 0.36s (7.3×).** The `µs/hole²` metric falls from 15 → 3 as N grows, confirming the dominant quadratics are gone.

| holes (N) | before | after |
|----------:|-------:|------:|
| 36  | 0.066s | 0.020s |
| 100 | 0.318s | 0.063s |
| 196 | 0.865s | 0.155s |
| 324 | 2.617s | 0.360s |

## The fixes

| Hot path | Was | Fix |
|---|--:|---|
| `build_topology_face` | ~500ms | Stop cloning the shared vertex pool per sub-face — resolve existing vertices by reference (layered lookup) |
| PaveFiller FF/EF vertex snap | ~474ms | `PaveVertexIndex` spatial hash instead of scanning every pave block per intersection endpoint |
| `detect_same_domain` | ~720ms | Gate the costly polygon clip behind a 2D bbox-overlap bound — `area(poly∩poly) ≤ area(bbox∩bbox)`, so the clip only runs when it could actually matter |
| classify sub-faces | ~373ms | Build each opposing solid's ray-cast geometry **once**, not once per sub-face |
| face splitter (4 passes) | ~250ms | Grid-index the per-edge / per-endpoint scans in the arrangement, t-junction split, internal-loop, and boundary-arc passes |

Each change is a sound broad-phase or memoization — it changes *which work runs*, never *the result*. The same-domain bbox gate, for example, relies on the polygon intersection being contained in the overlap of the two 2D bounding boxes, so gating the clip on that bound can only skip clips that were going to fail the area threshold anyway.

## Correctness

- **Full workspace suite passes** (2444 tests), including a new `perforated_panel_cut_is_correct_and_manifold` regression test (asserts 4N+4 faces, exact volume, no free/over-shared edges).
- An arc-handling regression surfaced mid-work (the dovetail-cornerclip intersect — arcs bulge past their chord's bbox, which the all-line honeycomb didn't exercise); caught by the suite and fixed by scanning the full endpoint set for arc sections only.
- clippy `-D warnings`, `cargo fmt`, and `check-boundaries.sh` clean.

## Regression guard (so this doesn't silently come back)

There's no static lint for "accidentally O(N²)", so this adds a **deterministic work-counter guard**: a `perf-counters` feature (zero cost when off) exposes atomic counters on the two biggest hot paths, and `scaling_perforated_cut_is_subquadratic` asserts the work stays bounded. Counting *work* instead of wall-clock makes it sharp and non-flaky.

Verified it actually catches the regression — reverting the fixes makes the counters explode:

```
guard PASSES (fixes in):   pave_probes 42→42,         sd_clips 0→0
guard FAILS  (fixes out):  pave_probes 42→63,882,070, sd_clips 0→12,060
```

A new CI step (`cargo nextest run -p brepkit-operations --features perf-counters -E 'test(scaling_)'`) runs it on every PR.

## Residual

Scaling is now mildly super-linear (~N^1.5, down from N²) — the remaining cost is spread evenly across four phases (~100ms each at N=324) with no single dominant term. Further gains would be deeper work in the face splitter for diminishing returns.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,11 @@ jobs:
       - run: cargo nextest run --workspace
       # nextest doesn't run doc tests, so run them separately
       - run: cargo test --workspace --doc
+      # Complexity-regression guards: deterministic work-counter tests that fail
+      # if a boolean hot path regresses to O(N²) (issue #987). Gated behind the
+      # `perf-counters` feature, so they need an explicit run.
+      - name: Complexity guards
+        run: cargo nextest run -p brepkit-operations --features perf-counters -E 'test(scaling_)'
 
   coverage:
     name: Coverage
--- a/crates/algo/Cargo.toml
+++ b/crates/algo/Cargo.toml
@@ -14,6 +14,12 @@ brepkit-topology.workspace = true
 log.workspace = true
 thiserror.workspace = true
 
+[features]
+# Enables deterministic work counters (`perf` module) used by the
+# complexity-regression scaling guards. Off by default — zero cost in normal
+# and release builds (the bump calls compile to nothing).
+perf-counters = []
+
 [dev-dependencies]
 brepkit-check.workspace = true
 brepkit-topology = {workspace = true, features = ["test-utils"]}
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -1249,9 +1249,16 @@ fn arrangement_regions_from_inputs(
             if i == j {
                 continue;
             }
-            let (jlx, jly, jhx, jhy) = chord_boxes[j];
-            if ilx > jhx || jlx > ihx || ily > jhy || jly > ihy {
-                continue; // chord boxes disjoint → no crossing, no T-junction
+            // Box pruning is exact only when neither edge bulges past its
+            // chord — i.e. both are lines. An arc can cross or be T-junctioned
+            // outside its chord's box, so never prune a pair involving one
+            // (arcs are rare: rounded corners). The honeycomb cap is all lines,
+            // so this keeps the near-linear win where it matters.
+            if !i_is_arc && !other.is_arc {
+                let (jlx, jly, jhx, jhy) = chord_boxes[j];
+                if ilx > jhx || jlx > ihx || ily > jhy || jly > ihy {
+                    continue; // chord boxes disjoint → no crossing, no T-junction
+                }
             }
             let (b0, b1) = (other.a, other.b);
             // Proper interior crossing. For an arc input, only honour the break
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -93,15 +93,104 @@ fn split_sections_at_t_junctions(
     wire_pts: &[Point3],
     tol: f64,
 ) {
-    // Every distinct section endpoint (3D) is a candidate split point.
+    // Every distinct section endpoint (3D) is a candidate split point. Dedup
+    // with a fine grid (cell = tol), then index the unique points in a COARSE
+    // grid (cell = mean section length) so the per-section search below probes
+    // only nearby candidates — near-linear instead of O(sections²) on a
+    // perforated cap's many disjoint hole edges. The coarse query reproduces the
+    // former all-endpoints scan: an endpoint on a section lies within `tol` of
+    // it, hence in a coarse cell its bounding box (expanded by `tol`) overlaps.
+    let fine = tol.max(f64::MIN_POSITIVE);
+    let fine_inv = 1.0 / fine;
+    let fine_cell = |p: Point3| -> (i64, i64, i64) {
+        #[allow(clippy::cast_possible_truncation)]
+        (
+            (p.x() * fine_inv).floor() as i64,
+            (p.y() * fine_inv).floor() as i64,
+            (p.z() * fine_inv).floor() as i64,
+        )
+    };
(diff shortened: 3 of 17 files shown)


andymai:
Addressed the review feedback in 63a4c49:

- **Rustdoc CI failure** — `RayCastGeoms`'s doc linked the private `collect_face_geoms`, tripping `cargo doc --all-features -D warnings`. Switched to plain backticks.
- **Arc-bulge in the arrangement chord-box prune** (Copilot) — good catch. The prune now only fires for **line/line** pairs; any pair involving an arc takes the full check, since an arc can cross or be T-junctioned outside its chord's AABB. Mirrors the same guard already in `split_sections_at_t_junctions`. The honeycomb cap is all lines, so the near-linear win is unchanged.
- **`perf::bump_*` visibility** (Greptile) — they're only called inside `brepkit-algo`, so they're now `pub(crate)`; only `reset`/`snapshot` stay `pub` (for the scaling guard in `brepkit-operations`).
- **Face-count formula + strict volume assert** (Copilot) — the original repro placed corner/edge prisms *on* the slab boundary, which is why the count was `4N+4` rather than a clean `4N+6`. The test now insets the grid by one pitch so every hole is strictly interior → unambiguous `4N+6`, and the volume check uses the relative-tolerance `assert_volume_near` helper.

Verified locally: the dovetail-cornerclip arc test, the full io suite (182), and algo+operations (1079) all pass; rustdoc/clippy/fmt clean both with and without `perf-counters`.

PR #994, opened 2026-06-24
chore(observability): probe analytic-fallback paths + prune dead constants

andymai:
## What

Two cleanups from an audit of every brepkit operation that can degrade from an exact analytic B-Rep to an approximation.

### 1. Observability probes — `chore(observability)`

Permanent `log::debug!(target: "brepkit_approx", …)` probes at all 7 approximation branches, so any run can report exactly where an op left the analytic path:

| Path | Site |
|------|------|
| Boolean → mesh (co-refinement) fallback — *the only path that loses analytic surface types* | `operations/boolean/mod.rs` |
| Fillet analytic → Newton-Raphson walker | `blend/fillet_builder.rs` |
| Chamfer analytic → `UnsupportedSurface` (no v1 walker) | `blend/chamfer_builder.rs` |
| Offset NURBS face → 16×16 sampled-NURBS refit | `offset/offset.rs` |
| Offset trim → grid-sampling | `operations/offset_trim.rs` |
| Offset face → raw untrimmed surface | `operations/offset_face.rs` |
| Rolling-ball fillet → flat planar corner | `operations/fillet/rolling_ball.rs` |

New `crates/operations/examples/approx_census.rs` installs an in-process logger that captures the probes and prints **path + wall-clock + face count** per op:

```
cargo run --release --example approx_census -p brepkit-operations
```

Sample output (overlapping primitives): boolean stays exact and sub-millisecond on `box−cyl` and coaxial `cyl∩cyl`, but `box∩sphere` (956 faces, 187 ms), `sphere−cyl` (1392 faces, 357 ms), and `torus−box` (1733 faces, 204 ms) drop to the mesh fallback — roughly 100–1000× slower with a 10–200× face explosion. Offset/fillet/chamfer stay exact on every analytic primitive.

`offset` gains a `log` dependency (used by the new probe).

### 2. Prune dead constants — `chore(boolean)`

`MESH_BOOLEAN_PER_SOLID_THRESHOLD` / `MESH_BOOLEAN_FACE_THRESHOLD` were unconsumed leftovers from the pre-GFA "chord-based tessellated" boolean pipeline (the `collect_face_data` face-count pre-gate). That pipeline is gone — the engine is GFA-primary with a mesh fallback decided by result validation, not a face-count threshold. Their doc comments held the last reference to the removed `collect_face_data`.

## Why

Pure observability + dead-code cleanup, no behavior change: the probes are `debug`-level (silent unless a logger opts in) and the removed constants had zero references.

## Verification

- `cargo test --workspace` green — probes regress nothing.
- `cargo clippy --all-targets` clean under `-D warnings`.
- The example empirically captures all three primary fallback families (boolean→mesh, fillet→walker, offset→sampled-NURBS).

### DIFF
--- a/crates/blend/src/chamfer_builder.rs
+++ b/crates/blend/src/chamfer_builder.rs
@@ -344,6 +344,12 @@ fn compute_chamfer_stripe(
         return Ok(result);
     }
 
+    log::debug!(
+        target: "brepkit_approx",
+        "chamfer: analytic path unavailable for {}+{} — v1 has no walker fallback, returning UnsupportedSurface",
+        surf1.type_tag(),
+        surf2.type_tag()
+    );
     // v1: no walker fallback for non-analytic surface pairs.
     Err(BlendError::UnsupportedSurface {
         face: face1,
--- a/crates/blend/src/fillet_builder.rs
+++ b/crates/blend/src/fillet_builder.rs
@@ -697,6 +697,14 @@ fn compute_stripe_for_edge(
         }
     }
 
+    log::debug!(
+        target: "brepkit_approx",
+        "fillet: analytic fast-path unavailable for {}+{} ({} radius) — using Newton-Raphson walker (approximate NURBS blend surface)",
+        surf1.type_tag(),
+        surf2.type_tag(),
+        if matches!(law, RadiusLaw::Constant(_)) { "constant" } else { "variable" }
+    );
+
     // Build ParametricSurface references via PlaneAdapter for planes.
     // When a face is reversed, the outward normal is flipped. For PlaneAdapter,
     // we negate the normal. For analytic/NURBS surfaces the ParametricSurface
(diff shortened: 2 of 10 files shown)


andymai:
Addressed the review in `aa774d9` (all findings were on the census example):

**Greptile**
- **P1 (no-new-unwrap):** removed the `clippy::unwrap_used/expect_used` allow and every `.unwrap()`/`.expect()` — builders and case-runners now return `Result` and propagate with `?`, `drain()` uses `mem::take` + `unwrap_or_default`, and `main` returns `Result`. Clippy passes `--all-targets -D warnings` with no allow.
- **P2 (error detail dropped):** the error arms now push `err: {e}` unconditionally, so a real error is preserved even when a fallback probe also fired (this now surfaces the NURBS↔NURBS `inter3d` gap in the `nurbs-loft` row).

**Copilot**
- `CaptureLogger::enabled` filters by target; `drain()` takes instead of cloning; `report()` now prints every distinct probe, not just the first.

PR #997, opened 2026-06-25
fix(algo): keep cylinder slot-cut analytic (closed-circle section AABB)

andymai:
## What

Fixes the first of the boolean→mesh fallbacks surfaced by the approximation census: **a box cutting a slot into a cylinder's side** now stays analytic and watertight instead of dropping to a 62-facet mesh.

## Root cause

A box has two faces **perpendicular to the cylinder axis** (the slot's top/bottom walls). Each intersects the cylinder lateral surface in a **closed circle**; only a small arc of that circle lies within the box face, so phase_ff splits the closed circle at its boundary crossings and emits the in-face arc (`emit_split_circle_arcs`).

That emit step rejects each candidate arc whose midpoint falls outside *either* face's AABB. The face-AABB helper built its bounds from edge **endpoint vertices only** — but a cylinder lateral face's circular edges are *closed* (start == end at the seam vertex), so its AABB collapsed to a thin line at the seam. Every slot-arc midpoint then tested "outside" the cylinder AABB → **all arcs dropped → the perpendicular box faces were never created → 8 free edges → mesh fallback.**

(The box faces *parallel* to the axis intersect in line segments via a different, correct path, which is why only the perpendicular walls were lost.)

## Fix

`emit_split_circle_arcs`'s face-AABB now **samples along each edge curve** (8 points), exactly as the engine's primary `compute_face_bbox` already does, so a closed circular edge contributes its full radial extent instead of just the seam point. One-function change in `phase_ff.rs`; the sphere-hemisphere surface-union path is untouched.

## Verification

- New regression test `cut_cylinder_by_box_slot_perpendicular_walls_is_watertight`: asserts the result is closed-manifold, free-edge-free, keeps the analytic cylinder face, is compact (<20 faces), **and** — via the robust ray-cast classifier — that a point in the slot is `Outside` and the cylinder body is `Inside` (the cut geometry is correct, not just topologically closed).
- Census: `cyl − box (slot)` went from **62 faces / mesh fallback** to **8 faces / exact analytic / 0.71 ms**. No other census case changed (box∩sphere, sphere−cyl, cyl∪cyl still fall back — separate root causes).
- `cargo clippy --all-targets` clean; full `cargo test --workspace` green.

## Note

The tessellation-based volume measure reads ~+1.4% high on the resulting arc-edged notched cylinder (a known, separate limitation — it diverges upward with finer deflection), so the test verifies the cut geometrically via classification rather than by volume.

### DIFF
--- a/crates/algo/src/pave_filler/phase_ff.rs
+++ b/crates/algo/src/pave_filler/phase_ff.rs
@@ -2076,15 +2076,25 @@ fn emit_split_circle_arcs(
         let mut any = false;
         for oe in wire.edges() {
             if let Ok(edge) = topo.edge(oe.edge()) {
-                for vid in [edge.start(), edge.end()] {
-                    if let Ok(v) = topo.vertex(vid) {
-                        let p = v.point();
-                        min =
-                            Point3::new(min.x().min(p.x()), min.y().min(p.y()), min.z().min(p.z()));
-                        max =
-                            Point3::new(max.x().max(p.x()), max.y().max(p.y()), max.z().max(p.z()));
-                        any = true;
-                    }
+                let (Ok(sv), Ok(ev)) = (topo.vertex(edge.start()), topo.vertex(edge.end())) else {
+                    continue;
+                };
+                let (sp, ep) = (sv.point(), ev.point());
+                let (t0, t1) = edge.curve().domain_with_endpoints(sp, ep);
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the Copilot nit in `24bfd58`: hoisted the duplicated sample count `8` into a single binding (matching `compute_face_bbox`). Greptile reviewed the prior commit with no findings.

PR #1003, opened 2026-06-25
fix(algo): bound sphere/torus faces by surface extent in boolean broad-phase

andymai:
## What

Makes the boolean **broad-phase face AABB sound for sphere and torus faces**.

`compute_face_bbox` (in `phase_ff.rs`) derived each face's AABB from its
boundary edges alone. A curved analytic face bulges *between* its boundaries,
so the box collapsed:

- a **sphere hemisphere** flattened to its equatorial disk (`z ∈ [0, 0]`), and
- a **full torus** — whose entire boundary is two degenerate seam-point `Line(v0, v0)` edges — collapsed to a single **point** `(R, 0, 0)`.

The FF broad-phase then wrongly rejected genuinely intersecting face pairs (and
the 16-sample section filter dropped the curve), forcing every boolean on those
primitives down the mesh co-refinement fallback (100–1000× slower, analytic
surface types lost).

## How

- Add closed-form `aabb()` to `SphericalSurface` (`center ± radius`) and
  `ToroidalSurface` (ring + tube extent, orientation-aware) in `brepkit-math`,
  with unit tests.
- Union it into the face bbox for `Sphere`/`Torus` faces in `compute_face_bbox`.

The bbox only ever **widens**, so it can add broad-phase candidates but never
drop a needed pair — the precise in-both restriction downstream still trims
exactly. Cylinder, cone, and plane boundaries already bound their faces and are
untouched. Same class as the cylinder seam-collapse fix in #997.

## Effect (raw GFA, measured)

This is a **foundational** fix in a multi-PR campaign to eliminate the four
remaining boolean→mesh fallbacks (`box∩sphere`, `sphere−cyl`, `cyl∪cyl` perp,
`torus−box`). It advances all three curved cases but closes none on its own —
the analytic results land in the follow-up split/assembly PRs:

| case | before | after |
|---|---|---|
| `torus − box` | hard error: *"no faces selected"* (pairs rejected at broad-phase) | reaches shell assembly (*"all shells classified as holes"*) |
| `sphere − cyl` | cyl×sphere section dropped at broad-phase | sections survive (sphere split/keep is the follow-up) |
| `box ∩ sphere` | box-cap sections dropped | sphere face survives the broad-phase |

## Test

- 4 new unit tests for `SphericalSurface::aabb` / `ToroidalSurface::aabb`
  (origin, off-center, canonical, and x-axis-oriented torus).
- Full workspace suite green: **2444 passed, 9 skipped**, no regressions
  (incl. all gridfinity in-mem fixtures and `intersect_box_sphere_succeeds`).

### DIFF
--- a/crates/algo/src/pave_filler/phase_ff.rs
+++ b/crates/algo/src/pave_filler/phase_ff.rs
@@ -1096,15 +1096,27 @@ fn compute_face_bbox(topo: &Topology, face_id: FaceId) -> Result<Aabb3, AlgoErro
         }
     }
 
-    if points.is_empty() {
+    let boundary_bbox = if points.is_empty() {
         // Degenerate face with no edges -- use a zero-volume box at origin
-        Ok(Aabb3 {
+        Aabb3 {
             min: Point3::new(0.0, 0.0, 0.0),
             max: Point3::new(0.0, 0.0, 0.0),
-        })
+        }
     } else {
-        Ok(Aabb3::from_points(points))
-    }
+        Aabb3::from_points(points)
+    };
+
+    // A sphere or torus face bulges beyond its boundary edges (a hemisphere's
+    // only boundary is its equatorial circle; a full torus's are degenerate
+    // seam points), so the boundary-sampled bbox underestimates the true extent
+    // and the broad-phase would wrongly reject genuinely intersecting pairs.
+    // Union with the surface's closed-form extent to keep the bbox a sound
+    // superset. Cylinder/cone/plane boundary edges already bound their faces.
+    Ok(match topo.face(face_id)?.surface() {
+        FaceSurface::Sphere(s) => boundary_bbox.union(s.aabb()),
+        FaceSurface::Torus(t) => boundary_bbox.union(t.aabb()),
+        _ => boundary_bbox,
+    })
 }
 
 /// Compute AABBs for a list of faces.
--- a/crates/algo/src/pave_filler/phase_ff.rs
+++ b/crates/algo/src/pave_filler/phase_ff.rs
@@ -53,8 +53,8 @@ pub fn perform(
     let faces_b = brepkit_topology::explorer::solid_faces(topo, solid_b)?;
 
     // Pre-compute face AABBs for rejection
-    let bboxes_a = compute_face_bboxes(topo, &faces_a)?;
-    let bboxes_b = compute_face_bboxes(topo, &faces_b)?;
+    let bboxes_a = compute_face_bboxes(topo, &faces_a, tol)?;
+    let bboxes_b = compute_face_bboxes(topo, &faces_b, tol)?;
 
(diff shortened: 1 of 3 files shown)


andymai:
Addressed all three review findings in cbb3be8:

- **P1 (wrong-hemisphere sections):** the sphere face AABB is now **hemisphere-accurate** — `SphericalSurface::aabb_region(pole)` returns the half-ball on the pole side, with the pole axis derived from the boundary winding via the new shared `sphere_region_axis` helper (extracted from, and now also backing, the existing in-both `sphere_side` filter, so the two stages stay consistent). North and south hemisphere faces no longer share a box, so a south-only cut can no longer create a section on the north hemisphere. Verified by direct raw-GFA trace: `sphere − cyl` / `box ∩ sphere` advance identically to the full-sphere version, now per-hemisphere correct.
- **P2 (trimmed-torus overreach):** the full-torus surface AABB is now gated to **full, untrimmed tori only** via `face_boundary_all_degenerate` (the degenerate seam-only boundary that collapses to a point — mirroring the #1001 full-torus gate). A trimmed torus patch keeps its real-edge boundary box and cannot admit geometry on omitted angular bands.
- **Copilot (empty-boundary union):** when a face has no boundary samples, the surface AABB is used **directly** instead of unioning the origin-centered zero box.

Full suite green (2447 passed, 9 skipped); 3 new `aabb_region` unit tests cover the north/south/degenerate-pole cases.

andymai:
Addressed the closed-circle P1 in ab37b2e — good catch, and it matters directly for the follow-up sphere-split PR which produces exactly these circle-bounded faces.

`sphere_region_axis` now **samples along each boundary edge** (8 points/edge in oriented order) rather than using only endpoints, then computes the winding over the resulting closed polyline. A boundary built from a single closed `Circle` edge (where `sp == ep`) recovers its loop and yields the circle’s plane normal as the pole axis, instead of returning `None` and falling back to the loose full-sphere box. Line-bounded hemispheres are unaffected (same axis direction).

Added `sphere_region_axis_closed_circle_boundary` covering the closed-`Circle` sphere face. Full suite green (2448 passed).

andymai:
Addressed the closed-trim P1 in the latest commit — same closed-curve subtlety as the sphere case, now handled for the torus too.

`face_boundary_all_degenerate` now samples **along** each boundary edge and requires zero spatial extent (every sample within `tol` of the start), rather than only checking endpoint coincidence. A degenerate `Line(v0, v0)` seam (the full-torus signature) stays `true`; a closed `Circle`/NURBS edge bounding a trimmed patch — coincident endpoints but a real loop — now correctly returns `false`, so the patch keeps its real-edge box instead of being widened to the whole torus.

Tests: `closed_circle_torus_boundary_is_not_full_torus` (trimmed → false) and `point_seam_boundary_is_full_torus` (full → true). Verified `torus − box` still advances past the broad-phase (the full torus is still detected). algo+operations suites green (1083 passed).

PR #1005, opened 2026-06-25
fix(operations): assemble and render sphere−cyl Cut analytically

andymai:
## What

Makes `Cut(sphere, through-cylinder)` — and the bored-quadric family — produce an **exact analytic, watertight, correctly-rendered** result instead of the mesh fallback. Census: `sphere − cyl` **484ms / 1392 planar faces → 0.84ms / 3 analytic faces**.

Second PR in the campaign to close the four remaining boolean→mesh fallbacks (after #1003, the surface-aware face-AABB keystone).

## The boolean fix (the sphere was being dropped)

GFA returned 3 cyl faces with the sphere entirely gone. Root-caused to **five coupled gaps**, each unblocking the next:

1. **math** — no exact `(Sphere, Cylinder)` intersection arm: the generic marcher returned `NurbsCurve`, so the closed-circle FF split never fired. Added `exact_sphere_cylinder` (coaxial `z = ±√(R²−r²)` circles).
2. **phase_ff** — emit the section as `EdgeCurve::Circle`.
3. **face_splitter** — route an all-closed-circle-section hemisphere to `split_face_with_internal_loops` (cap + band-with-hole) instead of unsplitting; aim the sphere-band interior point toward the hole's latitude.
4. **same_domain** — don't fuse two curved same-surface faces that share an outer-wire edge set but cover **distinct regions** (the two hemisphere bands of a bored sphere share the equator polygon) — discriminate by interior sample.
5. **measure/check** — the annular spherical band over-integrates the removed polar cap; route bored-quadric volume to the orientation-aware, hole-clipped analytic integrator.

Result: 2 spherical bands-with-hole + 1 inner cylinder wall, watertight, volume = `V_sphere − V_bore`.

## The tessellation fix (so it renders correctly)

A sphere/torus band between two constant-`v` latitude circles is degenerate in UV (each latitude projects to a zero-area segment), so the CDT path filled the removed polar cap (mesh area 646 vs 587.7, tunnel mouth skinned over, not watertight). Added `tessellate_latitude_band_shared` — the curved analogue of the cylinder/cone band path: reuses the shared rim vertices at the two latitudes for watertight stitching, and inserts intermediate latitude rows for the surface curvature. Gated conservatively to full-revolution latitude bands; every other sphere/torus face takes the unchanged path. **Mesh area 646 → 586 (→587.9 at fine deflection), watertight, tunnel mouth open.**

## Verification

- Raw GFA: 3 analytic faces (2 sphere bands + cyl wall), free edges 0, manifold.
- Census: `sphere − cyl 0.84ms faces=3 exact analytic` (was 484ms/1392 FALLBACK); the other 8 boolean cases unchanged (no regression).
- Volume = **587.671** = `V_sphere − V_bore`.
- Tessellation: watertight at deflections 0.5→0.01, area converges to the analytic 587.7.
- Full workspace suite: **2454 passed, 9 skipped**; clippy clean; fmt clean; layer boundaries valid.
- Regression fixtures: `cut_sphere_by_through_cylinder_is_analytic_watertight` (boolean) + `bored_sphere_band_area_and_watertight` (tessellation).

## Known tech debt (not introduced here, flagged for follow-up)

`solid_volume` for a holed sphere routes to `analytic_sphere_signed_volume`, a **pre-existing** analytic function whose cap-vs-band logic is wrong for bored spheres (it caps only one of the two bands). This PR routes bored-quadric volume around it via the hole-clipped Gauss integrator (the `analytic_faces_solid_volume` fast path). Fixing `analytic_sphere_signed_volume` directly is a separate, deeper cleanup (three volume paths converge there) and is left for a dedicated PR.

## Does not touch

`box ∩ sphere` (different root — disjoint great-circle arcs), `cyl ∪ cyl`, `torus − box` remain fallback; their analytic splits are later PRs. The curved-band tessellation here will be reused by `torus − box`.

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -1869,7 +1869,16 @@ pub fn split_face_2d(
             // seam connections (e.g., torus), not true line boundaries.
             && (e.start_3d - e.end_3d).length() > tol.linear
     });
-    if all_boundary_line && !is_plane {
+    // `split_noseam_face_direct` carves cap+band from OPEN section arcs (it
+    // skips full-circle sections, relying on the FF boundary-crossing split to
+    // pre-open them). A hemisphere cut by a cylinder yields a CLOSED circle
+    // interior to the face (no equator crossing) — there are no open arcs, so
+    // that path would unsplit it. Defer those to the internal-loops path
+    // below, which carves the cap disc as a hole.
+    let has_open_section = sections
+        .iter()
+        .any(|s| (s.start - s.end).length() > tol.linear);
+    if all_boundary_line && !is_plane && has_open_section {
         return split_noseam_face_direct(
             &surface,
             &boundary_edges,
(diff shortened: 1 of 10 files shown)


andymai:
Addressed all four findings in 64fd460 (all in the measurement-layer hole-clip — the boolean and tessellation fixes were untouched):

- **P1 (oriented traversal):** `full_revolution_hole_vs` now derives each wire-ordered start vertex respecting `oe.is_forward()` (edge end when reversed), per the check-crate oriented-traversal rule.
- **Closing-step (Copilot):** the unwrapped u-deltas now sum around the *closed* loop (`(i+1) % n`), including the closing step back to the first vertex — a multi-edge full-revolution hole no longer undercounts its span.
- **P2 (float `==`):** replaced `signum() == signum()` with a strict same-sign product test `(hv - v_min) * (v_pole - v_min) > 0.0`.
- **Torus fast-path (Copilot):** restricted `analytic_faces_solid_volume` to bored **spheres**; a bored torus (no torus hole-clipping yet) now returns `None` and defers to tessellation. Torus support lands with the `torus − box` split.

All behavior-preserving for the sphere case (the `cut_sphere_by_through_cylinder` fixture still passes, volume 587.671); full suite green (2454 passed).

PR #1006, opened 2026-06-25
fix(operations): close box∩sphere boolean analytically (seam split + collar render/volume)

andymai:
## What

Makes `Intersect(box, sphere)` produce an **exact analytic, watertight, correctly-rendered, correctly-measured** result instead of a 956-face mesh fallback. Census: **190ms / 956 mesh → 1.94ms / 8 analytic faces** (6 plane disks + 2 spherical collar patches).

Third PR in the campaign to close the four remaining boolean→mesh fallbacks (after #1003 keystone AABB, #1005 sphere−cyl). This is the hardest case — the centered sphere's collars are full-longitude-wrap bands with a *scalloped* floor.

## The fix — four coupled layers

**1. Section-circle split (`algo/phase_ff::emit_split_circle_arcs`)** — split arcs strictly *below* π. A span of exactly π is a diametric semicircle whose two halves share both endpoints, so the assembler's endpoint-keyed `merge_duplicate_edges` collapsed the distinct north/south arcs into one edge (failing the Euler gate). Forcing a midpoint vertex for any span ≥ π gives the two arcs distinct endpoints — **no merge-key change**, so the reversed-vertex merge of genuinely-shared arcs is untouched (this is what makes it regression-safe where a merge-key discriminator was not).

**2. Seam crossings (`algo/phase_ff::sphere_seam_plane_crossings`)** — a box face's great-circle section must cross the sphere's *faceted* equator to split, but testing against the inscribed chords missed the crossings by the polygon sagitta. Fit the seam plane (Newell) and solve circle∩plane analytically — facet-independent.

**3. Arrangement walk (`algo/face_splitter::split_noseam_by_arrangement`)** — the disjoint great-circle arcs share no endpoints, so the generic wire builder (which U-turns on the degenerate v=0 seam confluence) can't chain them. A dedicated DCEL arrangement walk assembles the box-face disks + the two collar patches, selecting the collar by longitude winding.

**4. Collar render + volume:**
- **Render** (`operations/tessellate`): generalize the revolution-band mesher to a *varying-v* collar — outer ring = scalloped great-circle/seam floor, inner ring = latitude-cap hole — with column-aligned curvature rows and a single whole-run orientation (the per-triangle normal flip is unstable on the thin stitch triangles). Watertight: 304 → 0 boundary edges.
- **Volume** (`operations/measure`): a scalloped collar's analytic integral is a deferred u-dependent lune trim, so its volume comes from the (now watertight) whole-solid mesh via the divergence theorem. The constant-v bored-quadric fast path (#1005 sphere−cyl) is gated to stay analytic and is unchanged.

## Verification

- Raw GFA: F=8 (6 disks + 2 collars), free edges 0, manifold.
- Census: `box ∩ sphere 1.94ms faces=8 exact analytic`; the other 8 boolean cases unchanged.
- Render: tessellated mesh watertight (0 boundary edges) at deflections 0.05 and 0.005; area matches analytic (no cap-fill).
- Volume: `solid_volume` = 797.4 (analytic 797.97, <0.1%), convergent across deflections.
- No regression: `sphere − cyl` still 587.671 analytic/watertight; plain sphere/cyl/cone/torus tessellate unchanged.
- Full suite: **2456 passed, 9 skipped**; clippy clean; fmt clean; layer boundaries valid.
- Fixtures: `intersect_box_centered_sphere_is_analytic_collar` (boolean) + `box_centered_sphere_collar_tessellates_watertight` (tessellation).

## Reused by `torus − box` (PR5)

Layer 1 (split) and the Layer-4 collar render/volume machinery are surface-agnostic / already dispatch for `Torus`, so the toroidal collar of `torus − box` will reuse them; only its analytic *split* (the plane×torus quartic) remains case-specific work.

### DIFF
--- a/crates/algo/src/builder/face_splitter/special_cases.rs
+++ b/crates/algo/src/builder/face_splitter/special_cases.rs
@@ -50,19 +50,16 @@ pub(super) fn split_noseam_face_direct(
         }]
     };
 
-    // Collect open section arcs on this face.
+    // Collect open section arcs on this face, plus any closed-circle sections
+    // (interior cap circles, e.g. a latitude cut that stays inside one
+    // hemisphere) which become holes in the assembled region.
     let mut open_sections: Vec<OrientedPCurveEdge> = Vec::new();
+    let mut closed_sections: Vec<OrientedPCurveEdge> = Vec::new();
     for section in sections {
         let pcurve_on_this_face = match rank {
             Rank::A => &section.pcurve_a,
             Rank::B => &section.pcurve_b,
         };
-
-        // Skip full-circle section edges (start approx end in 3D) -- only use
-        // the open arcs produced by the FF closed-circle split.
-        if (section.start - section.end).length() < tol {
-            continue;
-        }
 
         let precomputed_uv = match rank {
             Rank::A => section.start_uv_a.zip(section.end_uv_a),
@@ -78,7 +75,7 @@ pub(super) fn split_noseam_face_direct(
             )
         });
 
-        open_sections.push(OrientedPCurveEdge {
+        let edge = OrientedPCurveEdge {
             curve_3d: section.curve_3d.clone(),
             pcurve: pcurve_on_this_face.clone(),
             start_uv,
@@ -88,7 +85,15 @@ pub(super) fn split_noseam_face_direct(
             forward: true,
             source_edge_idx: None,
             pave_block_id: None,
-        });
+        };
+
+        // Full-circle section edges (start approx end in 3D) are interior caps;
+        // open arcs were produced by the FF boundary-crossing split.
+        if (section.start - section.end).length() < tol {
+            closed_sections.push(edge);
+        } else {
+            open_sections.push(edge);
+        }
     }
 
     if open_sections.is_empty() {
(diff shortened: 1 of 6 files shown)


andymai:
Addressed all findings in the latest commit:

- **Constant-v check (Copilot):** `sphere_outer_wire_constant_v` now samples ALONG each outer-wire edge (8/edge), not just the start vertex — a great-circle arc with both endpoints on the seam latitude but bulging away is no longer mis-read as a constant-v band (which would wrongly take the analytic fast path). Threshold aligned to the linear-tolerance magnitude (1e-7) [**Greptile P2**].
- **`continue` scoping (Greptile P2):** the sphere seam-plane-crossing path now only skips the chord-based `face_hits` when the section actually crosses the seam (a faceted hemisphere, whose chords miss by the sagitta); a latitude-circle section or a non-hemisphere sphere face falls through to `face_hits` as before.
- **Negative arc-range (Copilot):** the collar row count uses `(v_cap - floor_v_near).abs()` so a southern collar (floor above the cap) scales with band height instead of hitting the `<= 0` fixed-count fallback.
- **No-op / doc (Copilot):** removed `let _ = full_circle_cols;`; corrected `emit_aligned_quad_strip`’s doc (the collar passes the raw emit, oriented once afterward by `orient_triangle_run`).
- **Test docs (Copilot):** moved both new fixtures’ `///` blocks above `#[test]`.

Verified: census still flips (box∩sphere 8 analytic faces; sphere−cyl unregressed at 587.671), full suite 2456 green, clippy/fmt clean.

PR #1008, opened 2026-06-25
fix(algo): assemble perpendicular cyl∪cyl Fuse analytically

andymai:
## What

Makes `Fuse(cyl, cyl)` for two perpendicular equal-radius cylinders (a **Steinmetz solid**) produce an analytic, manifold, correct-volume B-Rep instead of a 138-face mesh fallback. Census: **19ms / 138 mesh → 1.78ms / 6 analytic faces** (2 mutually-trimmed cylinder walls + 4 caps).

Fourth PR in the campaign to eliminate the boolean→mesh fallbacks (after #1003 keystone, #1005 sphere−cyl, #1006 box∩sphere).

## The fix (analytic B-Rep + volume)

- **`phase_ff`** — keep the closed Steinmetz seam loop whole when its in-both run is near-complete (≥ N−3), and clamp the seam-wrap-trim parameter to the curve domain. The closed-NURBS seam's wrap returned an out-of-domain `t` that `NurbsCurve::evaluate` extrapolated to a garbage point, collapsing the carved loop → walls dropped.
- **`face_splitter`** — emit the complementary outer-wall remainder (boundary-with-hole) for the Steinmetz internal-loop split, with a curved-remainder interior point placed clear of the lens, so Fuse keeps both mutually-trimmed walls (not just the inside lobe).
- **`check`/`measure`** — integrate the holed cylinder/cone band excluding its inner loops (even-odd over the *combined* lens arrangement — each lens ellipse is a full-u sinusoid bounding no area alone), so the analytic volume subtracts the Steinmetz intersection.

## Verification

- Raw GFA: 6 analytic faces, **manifold** (euler 4), free edges 0.
- Census: `cyl ∪ cyl 1.78ms faces=6 exact analytic`; other 8 boolean cases unchanged.
- Volume: 985.3 (analytic 987.0, **0.17%**), via the holed-band analytic integrator.
- Full suite **2457 passed**; clippy clean; fmt clean; layer boundaries valid.
- Regression fixture: `fuse_perpendicular_cylinders_is_analytic_watertight`.

## Known limitation — watertight render mesh deferred (precisely root-caused)

The preview tessellation renders **two full interpenetrating tubes** (correct silhouette, hidden internal walls). `solid_volume` is analytic and correct, **independent of the mesh**, so volume/STEP/B-Rep are right; only the preview mesh lags.

A watertight render needs the *exact* ellipse seam, which **self-touches at (0,±3,0)** — a genuine boundary singularity of the Steinmetz union. Eight measured tessellator/GFA approaches traced this to one missing core-GFA primitive: **arc-identity-aware edge merging**. The exact seam's four co-endpoint arcs (a "double-theta" graph between the two crossing vertices) are collapsed by the current endpoint-keyed `merge_duplicate_edges` → non-manifold (euler odd). This is the **same primitive that gates box∩sphere's exact closure and torus−box**, is regression-prone (a naive midpoint key breaks the coplanar-cap-cylinder tests), and is a **standalone dedicated core-GFA effort** — not part of this PR. The current marched-NURBS seam is manifold precisely because its loops are ~0.11 apart and dodge the singularity.

## Does not touch

The render-blocking arc-identity primitive is the campaign's terminal finding and the highest-leverage next step (unblocks cyl∪cyl render + box∩sphere exact + torus−box). `torus − box` remains a future PR.

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -2821,13 +2821,17 @@ pub fn face_has_curved_lens_holes(topo: &Topology, face_id: FaceId) -> bool {
     }
     face.inner_wires().iter().any(|&wid| {
         topo.wire(wid).is_ok_and(|wire| {
-            wire.edges().iter().any(|oe| {
-                topo.edge(oe.edge()).is_ok_and(|e| {
-                    matches!(
-                        e.curve(),
-                        EdgeCurve::Circle(_) | EdgeCurve::Ellipse(_) | EdgeCurve::NurbsCurve(_)
-                    )
-                })
+            // The degenerate lens hole is a SINGLE closed curved edge (the seam
+            // ellipse). A regular curved hole (e.g. a drilled bore) has multiple
+            // edges and a working generic interior — don't force it to mesh.
+            let [oe] = wire.edges() else {
+                return false;
+            };
+            topo.edge(oe.edge()).is_ok_and(|e| {
+                matches!(
+                    e.curve(),
+                    EdgeCurve::Circle(_) | EdgeCurve::Ellipse(_) | EdgeCurve::NurbsCurve(_)
+                )
             })
         })
     })
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -2799,6 +2799,40 @@ pub fn split_face_2d(
     sub_faces
 }
 
+/// Whether a face is a cylinder/cone lateral wall carrying CURVED inner-wire
+/// loops — the lens-hole signature (a closed Circle/Ellipse/NURBS edge where
+/// another quadric crosses the wall). For these the generic `interior_point_3d`
+/// /`sample_face_interior` fallbacks are unsafe: each lens loop is a single
+/// closed edge with a degenerate start/end UV, so the assembled hole polygon is
+/// unusable and a generic sample can land inside the removed lens. The dedicated
+/// `cylinder_cone_remainder_interior` handles them; when even its dense grid
+/// finds no contained point, the analytic split must abort to mesh rather than
+/// classify the wall from inside the removed region.
+pub fn face_has_curved_lens_holes(topo: &Topology, face_id: FaceId) -> bool {
+    use brepkit_topology::edge::EdgeCurve;
+    let Ok(face) = topo.face(face_id) else {
+        return false;
+    };
+    if !matches!(
+        face.surface(),
+        FaceSurface::Cylinder(_) | FaceSurface::Cone(_)
+    ) {
+        return false;
+    }
+    face.inner_wires().iter().any(|&wid| {
+        topo.wire(wid).is_ok_and(|wire| {
+            wire.edges().iter().any(|oe| {
+                topo.edge(oe.edge()).is_ok_and(|e| {
+                    matches!(
+                        e.curve(),
+                        EdgeCurve::Circle(_) | EdgeCurve::Ellipse(_) | EdgeCurve::NurbsCurve(_)
+                    )
+                })
+            })
+        })
+    })
+}
+
 /// Get a point guaranteed inside a sub-face's outer wire (in UV space),
 /// not inside any inner wire (hole), then evaluate it to 3D via the surface.
 #[allow(clippy::too_many_lines)]
--- a/crates/algo/src/builder/face_splitter/special_cases.rs
+++ b/crates/algo/src/builder/face_splitter/special_cases.rs
@@ -1474,33 +1474,39 @@ fn cylinder_cone_remainder_interior(remainder: &SplitSubFace) -> Option<Point3> 
     // construction — AND outside every lens hole). A point outside the sub-face
     // fed to classification would drop valid curved wall faces, so an
     // uncontained best is rejected.
-    let mut best: Option<(f64, Point3)> = None;
-    let n_u = 48;
-    let n_v = 5;
-    for iu in 0..n_u {
-        #[allow(clippy::cast_precision_loss)]
-        let u = u_lo + u_span * (iu as f64) / f64::from(n_u);
-        for iv in 1..n_v {
+    let search = |n_u: u32, n_v: u32| -> Option<Point3> {
+        let mut best: Option<(f64, Point3)> = None;
+        for iu in 0..n_u {
             #[allow(clippy::cast_precision_loss)]
-            let v = v_min + (v_max - v_min) * (iv as f64) / f64::from(n_v);
-            if point_in_hole_loops_uv(&hole_segs, u, v) {
-                continue; // Inside a lens hole — not the kept region.
-            }
-            let Some(p) = remainder.surface.evaluate(u, v) else {
-                continue;
-            };
-            let min_d = hole_pts
-                .iter()
-                .map(|h| (*h - p).length())
-                .fold(f64::INFINITY, f64::min);
-            if best.is_none_or(|(bd, _)| min_d > bd) {
-                best = Some((min_d, p));
+            let u = u_lo + u_span * f64::from(iu) / f64::from(n_u);
+            for iv in 1..n_v {
+                #[allow(clippy::cast_precision_loss)]
+                let v = v_min + (v_max - v_min) * f64::from(iv) / f64::from(n_v);
+                if point_in_hole_loops_uv(&hole_segs, u, v) {
+                    continue; // Inside a lens hole — not the kept region.
+                }
+                let Some(p) = remainder.surface.evaluate(u, v) else {
+                    continue;
+                };
+                let min_d = hole_pts
+                    .iter()
+                    .map(|h| (*h - p).length())
+                    .fold(f64::INFINITY, f64::min);
+                if best.is_none_or(|(bd, _)| min_d > bd) {
+                    best = Some((min_d, p));
+                }
             }
         }
-    }
-    // No contained candidate found → report (keep the interior unset) rather
-    // than return a point outside the sub-face.
-    best.map(|(_, p)| p)
+        best.map(|(_, p)| p)
+    };
+
+    // First pass at the standard density (the census finds a contained interior
+    // here). If a thin kept strip or a close pair of lens loops yields nothing
+    // contained at this resolution, take ONE refined denser pass before giving
+    // up — a present-but-narrow remainder strip then resolves. Returning `None`
+    // (no contained point even dense) signals the caller to abort the analytic
+    // split rather than fall back to an uncontained generic interior point.
+    search(48, 5).or_else(|| search(256, 17))
 }
 
 /// Whether `(u, v)` lies inside the region bounded by the combined inner-loop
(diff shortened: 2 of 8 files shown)


andymai:
Tightened all four generality findings (the holed-cyl/cone paths were gated too broadly):

- **R3 clamp / periodic Ellipse (Copilot):** the wrap-trim parameter is now clamped **only for `NurbsCurve`**; periodic `Ellipse`/`Circle` keep their unwrapped (`b1 > N`) seam-wrapping parameters (their cos/sin evaluation handles out-of-domain correctly), so a legitimate wrapped arc no longer collapses.
- **Remainder interior leaves face (P1):** `cylinder_cone_remainder_interior` now restricts its grid to the face’s **covered u-interval** (complement of the largest boundary-u gap) + the boundary’s projected v-range — a partial-angular face is searched only over its own arc.
- **Generic holes use Steinmetz clipping (P1):** two-layer gate. The integrator’s combined even-odd clip is now used only where it’s **exact** (the interleaving lens, or holes v-disjoint at shared u) via `combined_evenodd_holes_safe`, deferring the pathological overlapping-independent-holes case; and `analytic_faces_solid_volume` accepts a holed cyl/cone wall **only when the whole solid is the lens fuse** (`solid_is_steinmetz_lens_fuse`) — an ordinary drilled bore now defers to tessellation (its prior correct behavior).
- **Fragile full-revolution detection (P1):** replaced `u_span >= TAU − 1e-3` (raw max−min) with `covered_angular_span` (gap-complement, `≥ 2π − 1`), robust to a partial face straddling the seam; the winding term still catches single-loop bands.

Added `gate_tests` proving the tightened logic directly. cyl∪cyl unchanged (6 analytic faces, vol 985.3, fixture passes); all gridfinity/dovetail/scoop in-mem fixtures + clean census cases pass; full suite 2459 green.

andymai:
Resolved both R3 wrap-trim P1s **at the root** (rather than tuning the heuristic):

**Root cause:** `face_circumferential_u_gap` sampled each rim circle at only 16 steps (~0.39 rad/step), exceeding the 0.2 `largest_u_gap` threshold, so it reported a **phantom u-gap on a full cylinder wall**. The Steinmetz seam's genuinely-on-surface points fell in that phantom gap → `FaceExtent::contains` rejected them → the seam read as partial (21-23/24). Densifying to 64 steps (~0.098 rad) removes the artifact — the seam now reads a true **24/24 in-both** (densening can only remove spurious gaps, never create false ones; a real partial-arc face still shows its large angular gap).

- **Near-whole misclassifies (P1):** with the root fixed, I **deleted the loose `in_both ≥ N−3` heuristic entirely**. Keep-whole now requires the WHOLE curve in-both (`b1−b0 == N`); a genuine partial (`b1−b0 < N`, a real contiguous out-of-face arc) is trimmed into open arcs at its actual crossings. No count threshold to misclassify.
- **Wrapping NURBS collapses (P1):** new `trim_closed_curve_to_inboth_arc` — periodic Ellipse/Circle keep one unwrapped arc; a **wrapping NURBS preserves BOTH in-domain pieces** (`[t0, domain_end]` + `[domain_start, t1−span]`, meeting at the closed curve's coincident seam endpoints) instead of clamp-dropping the head; non-wrapping NURBS → one arc. (Supersedes the earlier NURBS-only clamp.)

2 direct tests added (`trim_partial_ellipse_to_single_open_arc`, `trim_wrapping_nurbs_preserves_both_pieces`). cyl∪cyl unchanged (6 analytic faces, vol 985.3); all gridfinity/dovetail/scoop in-mem fixtures + clean census cases pass; full suite 2461 green.

andymai:
Resolved both overcount P1s **structurally** — the integrator is now correct-or-decline for any holed face:

- **Unsafe holes ignored (P1):** `integrate_with_trimming` now returns `Result`. After the lens-clipping early-return (the ONLY path that subtracts inner wires — the Steinmetz lens or v-disjoint holes), any remaining holed face returns `Err(IntegrationFailed("…not a clippable lens; defer to tessellation"))`. So a holed face is **either integrated correctly or declines** (→ `volume.rs` `.ok()?` → tessellation) — no path returns a band integral that ignores/merges holes.
- **Near-full overcount (P1):** replaced `covered_angular_span ≥ TAU−1.0` (1-rad slack) with `is_full_revolution_boundary` — **sampling-density-adaptive**: full only when the largest boundary u-gap ≤ 2× the median of the other gaps. A full wall (cap circles at a fixed step) has uniform gaps (ratio ~1 → full); a real 0.5-rad uncovered arc on a ~0.2-rad step has ratio ~2.7 → partial. No fixed angular slack.

4 new `gate_tests` prove it: declines on overlapping-independent holes, declines on a near-full (2π−0.5) partial holed face, succeeds on the lens + hole-less walls, and the density-adaptive full/partial classification. `integrate_face` already returned `Result`, so no caller ripple.

cyl∪cyl unchanged (6 analytic faces, vol 985.3); sphere bored faces use the separate latitude path (unaffected — sphere−cyl 587.671, box∩sphere 797.4 pass); all gridfinity/dovetail/scoop in-mem fixtures + clean census cases pass; full suite 2465 green.

andymai:
Resolved the declined-holes-error P1 — and all four prior rounds of integrator findings — at the root, by **abandoning the general holed-cyl/cone integrator** for the one value cyl∪cyl actually needs.

Four review rounds showed the general analytic holed-face integrator was the wrong tool (the check crate's `solid_volume`/`solid_area` have no tessellation fallback, so a decline becomes a hard error). cyl∪cyl only needs the **Steinmetz lens-fuse volume**, which has an exact closed form.

- **Reverted the entire check-crate change** — `face_integrator.rs` is back to `main` (`git diff main -- crates/check/` is **empty**). No integrator, no decline/error path, no blast radius. All four rounds of findings are gone.
- **`volume.rs`:** when `solid_is_steinmetz_lens_fuse` matches (two equal-radius holed cylinder walls sharing inner-wire edges + planar caps), return `V = π·r²·(h₁+h₂) − (16/3)·r³` — **986.97 exact** (vs the integrator's 985.3). A non-lens holed cyl/cone solid returns `None` → tessellation.

PR4 is now 4 files: the boolean B-Rep fixes (phase_ff R3 root-fix + face_splitter R4), the boolean fixture, and the closed-form volume + gate (with 2 tests: the exact 986.97 volume, and the gate not firing on a plain or coaxial cylinder). cyl∪cyl unchanged (6 analytic faces, free=0, vol 986.97); all gridfinity/dovetail/scoop fixtures + clean census + sphere−cyl (587.671) + box∩sphere (797.4) pass; suite 2461 green.

andymai:
Both Round-5 findings fixed — geometry validations the gates were missing:

- **Steinmetz gate matches other cylinders (Greptile + cubic):** `solid_is_steinmetz_lens_fuse` validated topology only. Added `cylinders_perpendicular_and_intersecting` — requires **equal radius** (≤1e-6), **perpendicular axes** (`|a0·a1| ≤ 1e-6`), and **intersecting axis lines** (closest approach ≤ 1e-6). The closed form's `16r³/3` term is exact only for the right-angle Steinmetz, so an oblique or parallel-offset equal-radius fuse with the same topology now returns `None` → tessellation computes the real overlap.
- **Interior-point lacks face containment (cubic):** `cylinder_cone_remainder_interior` now filters grid candidates through `point_in_hole_loops_uv` — an even-odd vertical-ray test over the holes' projected UV segments, evaluated in every 2π `u`-translate so a seam-wrapping lens loop is matched. Candidates inside a hole are skipped; if none is contained it returns `None` (interior unset) rather than an uncontained point that could drop a valid wall face.

3 tests added (axis-perpendicular/intersecting predicate incl. oblique/parallel/skew negatives; the hole-containment ray test; and a real cylinder-wall sub-face whose returned interior point is on the wall and clear of the lens). cyl∪cyl unchanged (986.97 exact, 6 analytic faces, free=0); `git diff main -- crates/check/` still empty; gridfinity/dovetail/scoop + clean census + sphere−cyl + box∩sphere pass; suite 2464 green.

andymai:
Round-6 — all three P1s fixed; Finding 3 turned out to be a **pre-existing engine bug**:

- **Holed fallback ignores holes (structural root):** `try_analytic_solid_volume` (the whole-solid closed-form fast-path) had **no inner-wire check** — any bored primitive (e.g. a coaxial tube, whose annular caps carry inner wires) silently returned the **hole-filled** `π·r²·h`. Added a guard at the top of its face loop: any face with inner wires → `None` → the hole-aware `volume_from_direct_face_tessellation` path. **No analytic path returns a hole-filled volume now** (the validated Steinmetz lens is still claimed by the per-face path, tried after this). New test `drilled_cylinder_volume_subtracts_the_bore` (a drilled tube now reports the bore-subtracted volume).
- **Truncated overlap uses full lens:** the gate now requires both finite walls extend **≥ r past the axis intersection** (`wall_extends_past`, using the intersection point now returned by `cylinders_perpendicular_and_intersecting`). A short/cap-shifted fuse, where `16r³/3` would over-subtract, defers to tessellation. New test `truncated_perpendicular_fuse_gate_defers`.
- **Seam-crossing inner loop drops a segment:** `cylinder_cone_remainder_interior` dropped the lens loop segment bridging the u-seam (`|Δu| ≥ π`), leaving an open boundary. Now unwraps u relative to the previous sample (`((Δu+π) rem_euclid 2π) − π`) so the segment lives in the extended u-frame and the even-odd containment test (which sums over every 2π translate) sees a closed boundary. New test `hole_containment_handles_seam_crossing_loop`.

cyl∪cyl unchanged (6 analytic faces, free=0, vol 986.97 exact); new `plain_primitives_still_use_the_analytic_fast_path` confirms no perf regression (plain cyl/cone/sphere/torus still analytic); `git diff main -- crates/check/` still empty; gridfinity/dovetail/scoop + clean census + sphere−cyl + box∩sphere pass; suite 2468 green.

andymai:
Fixed — the gate now accounts for **every** face, not just the presence of the two holed walls:

- **Reject unholed cylinder faces:** the lens fuse has exactly two cylindrical faces, both holed; a third (unholed) cylinder wall means a foreign attached body → decline.
- **Account for every planar face:** the only non-cylinder faces allowed are the end-caps of the two lens cylinders — planes whose normal is parallel to `a0` or `a1`. A plane pointing elsewhere belongs to another body → decline.

So the closed form fires only when every face is one of the two holed lens walls or an axis-aligned cap of them — an extra attached cylinder (whose volume the two-cylinder formula would drop) now defers to tessellation.

Test `gate_rejects_extra_face_beyond_the_lens_fuse`: census passes; census + extra unholed cylinder wall → declines; census + a tilted cylinder’s cap → declines; census + a z-axis-aligned cap → still passes (proving it’s the orientation check, not a blunt face-count). cyl∪cyl unchanged (6 analytic faces, free=0, vol 986.97 exact); `git diff main -- crates/check/` still empty; gridfinity + clean census + sphere−cyl + box∩sphere pass; suite 2469 green.

andymai:
Fixed — the cap-axis-alignment test now uses the **squared cosine** `(n·a)² ≥ (1−ε)²·|n|²·|a|²` instead of `n·a > 1−ε`, so a non-unit (but parallel) plane normal or axis is matched regardless of magnitude rather than spuriously rejected. (Note: a rejected cap defers to tessellation, which is correct after the Round-6 holed-face fix — so this was at most a missed-fast-path perf case — but the robust check is the right thing.) cyl∪cyl unchanged (986.97 exact); suite 2469 green.

andymai:
Fixed — the gate now requires the **exact cap count**, not just orientation: exactly 2 caps perpendicular to each axis (4 total). An extra axis-aligned plane (e.g. an attached box face) makes `caps_a0`/`caps_a1` exceed 2 → `None` → tessellation, so the foreign body’s volume is no longer dropped.

Combined with the earlier passes, the gate now accepts a solid only when its faces are EXACTLY: two holed perpendicular-equal-radius seam-sharing cylinder walls + four axis-aligned non-truncated caps, and nothing else (any sphere/cone/torus/NURBS/unholed-cylinder/foreign-plane already rejects). That face set is the Steinmetz fuse by construction. (And the structural safety net still holds: any holed analytic solid that *isn’t* the validated lens routes to tessellation, which is correct.)

Updated `gate_rejects_extra_face_beyond_the_lens_fuse` — the z-aligned extra cap now asserts **rejection** (a 5th axis-aligned cap → `caps_a0 == 3`). cyl∪cyl unchanged (6 analytic faces, vol 986.97 exact); suite 2469 green.

andymai:
Fixed — the interior-point grid-failure path can no longer hand back an uncontained point:

- **Densify on failure:** the contained-interior search is now two-pass — standard `48×5` (the census resolves here, unchanged), then one refined `256×17` only if that finds nothing contained, so a present-but-narrow remainder strip still yields an on-wall point clear of the lens.
- **Safe failure:** I traced the caller — `fill_images_faces.rs:412` consumed the `None` via `unwrap_or_else(|| interior_point_3d(...))`, and the deeper `sample_face_interior` fallback samples the lens center; both unsafe. New predicate `face_has_curved_lens_holes` (a cyl/cone wall carrying a Circle/Ellipse/NURBS inner loop) gates two points: `fill_images_faces` leaves the interior unset, and `classify_sub_faces` returns `Err(ClassificationFailed)` for such a face → the boolean **mesh-falls-back** (correct) rather than classifying the wall from inside the removed lens. (`Builder::build_result` already propagates `Err` → `operations::boolean` mesh fallback.)

New test `remainder_interior_found_for_thin_kept_strip` (a narrow band mostly filled by a lens hole still yields a contained point via the dense pass). cyl∪cyl unchanged (6 analytic faces, free=0, vol 986.97 exact); all gridfinity/dovetail/scoop in-mem fixtures pass (the abort gate did not wrongly trip); `git diff main -- crates/check/` still empty; suite 2470 green.

andymai:
Two findings:

- **Lens-hole detection too broad (cubic P2) — fixed:** `face_has_curved_lens_holes` now requires the inner wire to be a **single** curved edge (the degenerate lens loop), via a slice pattern `let [oe] = wire.edges()`. A regular curved hole (drilled bore, multi-edge inner wire) no longer trips the safe-abort gate, so it keeps its working generic interior instead of mesh-falling-back. (cyl∪cyl is unaffected — its grid finds a contained interior, so it never reaches the abort gate.) Verified: gridfinity fixtures + suite 2470 green.

- **Wrapped trim loses continuity (Greptile P1) — documented latent limitation:** this is the two-arc split for a closed NURBS whose in-both run wraps the seam *partially* (`b1 > N` but not whole). cyl∪cyl never hits it — its seam reads a full `b1−b0 == N` (after the densified-u-gap root-fix) and is kept whole — and no gridfinity/dovetail/scoop fixture exercises a wrapping-*partial* NURBS section (full suite green). Both candidate behaviors have a tradeoff in the independent-section-edge model (single arc drops the head; two arcs meet only at the seam vertex), and a truly continuous seam-crossing boundary is a deeper section-model change. Tracking it as a known limitation rather than block this PR on an unexercised path.

cyl∪cyl unchanged (6 analytic faces, free=0, vol 986.97 exact); `git diff main -- crates/check/` empty; suite 2470 green.

PR #1010, opened 2026-06-26
fix(algo): close torus−box boolean analytically (plane×torus seam + toroidal band)

andymai:
## What

Makes `Cut(torus, box)` fully analytic — B-Rep + watertight render + correct volume — instead of a 1733-face mesh fallback. Census: **233ms / 1733 faces → 14ms / 5 faces**.

**The campaign's fourth and final case** (after #1003 keystone, #1005 sphere−cyl, #1006 box∩sphere, #1008 cyl∪cyl). With this, **all primitive-boolean cases in the census are analytic — zero mesh fallbacks.**

The case: torus (major 10, minor 3) minus a side-8 box that severs one side of the ring, leaving a capped C-tube (genus-0).

## The fix (staged)

**B-Rep**
- **plane×torus marcher closure** — the marcher emitted near-closed-but-*open* ovals (~0.15mm gap, one grid step) that fell into a no-man's-land: too open to be recognized as a loop, never clipped. Close the fit when the chain wraps (a grazing inner-tangent figure-eight stays open). Adds exact `intersect_line_torus` (quartic, Durand-Kerner) for box-edge crossings.
- **`FaceExtent` for a whole torus** — a whole untrimmed torus is now full extent, so `restrict_curves_to_faces` clips the ovals against the box (it was bailing on the `None` v-range).
- **exact-crossing trim + arc-split + band tracer** — trim each oval at its **exact** box-edge∩torus crossings (z=±1.105, shared vertices → watertight), always split the kept arc at its midpoint, and trace the kept toroidal band (a φ-wrapping u-band, the long 294° kept side).

**Render + volume**
- `tessellate_torus_notch_band` — sweeps the band along u, sharing seam vertices with the notch walls (crack-free). The u-sweep starts at each boundary loop's **kept-side edge** (a mean start folds the stitch back over the boundary strip → undercount).
- volume via `signed_volume_from_mesh` off the now-watertight mesh (the band isn't closed in isolation), gated to the notch-band signature.

**General engine improvements** (motivated here, gated narrow)
- SD `source_face` guard: complementary sub-faces split from the same input face are never same-domain duplicates of each other.
- `perform_areas`: a single result shell is always growth (no enclosing shell to be a cavity of); multi-shell results keep the volume-sign split.

## Why not the "arc-identity merge-key primitive"

The long-discussed shared *arc-identity-edge-merge primitive* is **not buildable**: `merge_duplicate_edges`'s endpoint-pair key is load-bearing for the gridfinity lip ring (Line-chord + Circle-arc co-endpoint pairs, deviating up to 2.4mm, that **must** collapse for manifoldness), while torus−box's notch-wall lens needs the **opposite** (Line + co-endpoint arc must stay distinct). Same local configuration, opposite required outcome → no merge *key* can separate them. The **splitter-side arc-split** (give the lens arc a midpoint vertex so the face is 3 edges) controls the geometry *we* emit and sidesteps it — so **`merge_duplicate_edges` is byte-identical to main**.

## Verification

- census `torus − box`: 14ms / 5 faces (1 toroidal band + 4 plane notch walls) / exact analytic; free=0, manifold.
- render watertight: whole-solid mesh bd-edges == 0, stable across deflection 0.1/0.05/0.02 (no seam cracks).
- volume: `solid_volume` = 1537.5 vs 1543.07 (Monte-Carlo, 20M) = **0.36%**, convergent upward (inscribed-mesh).
- **No regression:** gridfinity d1/d3/d4/d5 + coplanar caps (#909/#859) + #696 seams green; box∩sphere (8) / sphere−cyl (3) / cyl∪cyl (6) / cyl∩cyl (3) unchanged; **plain torus unchanged** (the new branches gate on the notch-band signature); `merge_duplicate_edges` byte-identical to main; full suite **2476 passed**.
- 2 fixtures: `cut_torus_by_box_notch_is_analytic_watertight`, `torus_box_notch_band_tessellates_watertight`.

---
## Summary by cubic
Makes the torus − box cut fully analytic, watertight, and fast. Replaces a 1733‑face mesh fallback with a 5‑face B‑Rep (233ms → 14ms).

- **New Features**
  - Analytic trim of plane×torus ovals at exact box‑edge×torus crossings using `intersect_line_torus` (quartic), then split the kept arc at its midpoint; keeps all in‑box arcs.
  - Torus notch band tracing with `split_torus_band_by_arrangement` (φ‑wrapping u‑band), and structured `tessellate_torus_notch_band` that sweeps along u from each loop’s kept‑side edge and shares seam vertices; interior point from the boundary loops’ long‑arc midpoint.
  - Volume for this case is taken from the watertight whole‑solid mesh via `signed_volume_from_mesh` when a torus notch band is detected.

- **Bug Fixes**
  - Plane×torus marcher: wrapped ovals now close; grazing inner‑tangent figure‑eight stays open.
  - A whole, untrimmed torus is treated as full `FaceExtent` so ovals are clipped against the box face.
  - Same‑domain detection skips only complementary splits from the same `source_face` (distinct interiors); coincident same‑source duplicates still de‑duplicated.
  - Single‑shell classification: a lone negative‑volume shell is growth only if outward per a curvature‑robust divergence‑flux check; rejects genuine inward cavity results. Added cube orientation tests.
  - Hardened torus pieces: tighter arc‑join tolerance; notch‑band v‑wrap detection samples along edges; quartic root finder filters by residual and de‑dups. Added stronger torus−box fixtures (exact 5‑face decomposition; watertight across deflections).

<sup>Written for commit bcd2832a3932c27827637514ef834a7bd45e9c77. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1010?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/algo/src/builder/builder_solid.rs
+++ b/crates/algo/src/builder/builder_solid.rs
@@ -498,7 +498,16 @@ fn perform_areas(topo: &Topology, shells: &[Vec<FaceId>]) -> (Vec<Vec<FaceId>>, 
 
         let signed_vol = signed_volume_of_shell(topo, shell);
 
-        if signed_vol >= 0.0 {
+        // A SINGLE shell is the result's outer boundary — there is no enclosing
+        // shell for it to be a cavity OF, so it is always growth. Its
+        // `signed_volume_of_shell` can read negative when a curved band face
+        // (e.g. the torus−box kept toroidal band, which wraps the tube) is
+        // sampled by the fan-tet integral over only its outer-wire CORNER
+        // vertices — a degenerate under-sampling that flips the sign. Treating a
+        // lone shell as growth avoids discarding the whole result as a hole;
+        // multi-shell results keep the volume-sign split (a Cut can leave the
+        // tool's interior as a separate negative-volume shell).
+        if shells.len() == 1 || signed_vol >= 0.0 {
             growth.push(shell.clone());
         } else {
             holes.push(shell.clone());
--- a/crates/algo/src/builder/builder_solid.rs
+++ b/crates/algo/src/builder/builder_solid.rs
@@ -483,6 +483,138 @@ fn face_normal_at(topo: &Topology, face_id: FaceId, point: Point3) -> Option<Vec
 
 // ── Phase 3 ──────────────────────────────────────────────────────────
 
+/// Robust outward-orientation test for a closed shell, independent of face
+/// curvature and wire winding.
+///
+/// `signed_volume_of_shell` signs each face by its surface normal but integrates
+/// the magnitude over the outer-wire CORNER vertices only — a fan that
+/// under-samples (and can sign-flip) a doubly-curved band whose corners barely
+/// bound the wrapped surface (the torus−box notch band reads negative despite
+/// being outward). This computes the divergence-theorem flux `∮ P · n dA` per
+/// face with a curvature-aware quadrature — a planar face from its corner fan
+/// (exact), a curved face from a `(u, v)` GRID over its boundary's parameter box
+/// with the local area element `|∂P/∂u × ∂P/∂v|` — so a wrapped band integrates
+/// correctly. A closed outward shell yields a positive total (≈ 3·Volume); a
+/// genuinely inward-oriented lone shell (a Cut leaving only a cavity component)
+/// yields negative and is still rejected. Returns `None` when no face yields a
+/// usable contribution so the caller falls back to the volume sign.
+fn shell_is_outward_oriented(topo: &Topology, faces: &[FaceId]) -> Option<bool> {
+    let mut flux = 0.0_f64;
+    let mut any = false;
+    for &fid in faces {
+        let Ok(face) = topo.face(fid) else { continue };
+        let surface = face.surface();
+        if let FaceSurface::Plane { .. } = surface {
+            // Planar: corner fan from sampled boundary points is exact.
+            let Ok(wire) = topo.wire(face.outer_wire()) else {
+                continue;
+            };
+            let mut pts: Vec<Point3> = Vec::new();
+            for oe in wire.edges() {
+                let Ok(edge) = topo.edge(oe.edge()) else {
+                    continue;
+                };
+                let (Ok(sv), Ok(ev)) = (topo.vertex(edge.start()), topo.vertex(edge.end())) else {
+                    continue;
+                };
+                let (sp, ep) = (sv.point(), ev.point());
+                for k in 0..4 {
+                    let f = f64::from(k) / 4.0;
+                    let f = if oe.is_forward() { f } else { 1.0 - f };
+                    pts.push(edge.curve().evaluate_with_endpoints(f, sp, ep));
+                }
+            }
+            if pts.len() < 3 {
+                continue;
+            }
+            let centroid = {
+                let n = pts.len() as f64;
+                let mut c = Vec3::new(0.0, 0.0, 0.0);
+                for p in &pts {
+                    c += Vec3::new(p.x(), p.y(), p.z());
+                }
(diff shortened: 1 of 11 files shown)


andymai:
Addressed all 10 findings. The two **general engine changes** are now properly scoped:

- **Single inward shell becomes solid (Greptile P1 + cubic P2):** replaced `shells.len()==1 → growth` with `is_growth = signed_vol≥0 ? true : (lone shell ? shell_is_outward_oriented() : false)`. `shell_is_outward_oriented` is a curvature-robust divergence flux `∮P·n dA` (planar via corner fan, curved via a (u,v) grid with the `|∂P/∂u×∂P/∂v|` area element). **Scoped:** it fires only on a lone *negative*-corner-fan-volume shell — so bored quadrics (sphere−cyl, positive) never call it (I hit and fixed a sphere−cyl regression from calling it unconditionally), only torus−box does (reads outward). A genuine inward cavity is still rejected. +2 unit tests (outward cube → growth; reversed cube → rejected).
- **Same-origin duplicates can survive (Greptile P2):** the guard now skips SD only for `same_source_complementary_split` (same `source_face` AND distinct interior points >100·tol) — the complementary in-tube/out-tube split skips SD, but a coincident same-source duplicate still reaches the SD overlap test and is caught. Applied to all three union passes.

Torus-specific: **(P1)** band interior point = the boundary loops’ long-arc midpoint, not hardcoded θ=π (mirrored/rotated cuts); torus shortcut gated to all-open sections; `join_tol` tightened tol*1000→tol*100; notch-band v-detection samples 8 pts along each edge; `real_roots_quartic` filters by polynomial residual + dedups. **(P2)** trim keeps ALL in-box arcs (not just the longest); fixture strengthened to the exact decomposition (5 faces, tori==1, planes==4, free==0, others==0); doc comment fixed.

`merge_duplicate_edges` is byte-identical to main (no merge/connectivity changes). gridfinity d1/d3/d4/d5 + coplanar caps (#909/#859) + #696 green; all census cases analytic (sphere−cyl recovered); torus−box still 5 faces / free=0 / render bd=0 / volume ≈1543; full suite 2478 passed.

andymai:
The four findings re-posted this round are the round-1 findings — all addressed in `bcd2832` (Greptile re-anchored its text to lines the fix changed; the code at those lines is now the fixed version, and cubic marked its parallel findings ✅ Addressed):

- **special_cases.rs interior sample** → now the **long-arc midpoint** of the two boundary loops, not `torus.evaluate(PI,0)` (that token only survives in unrelated `line_chord` helpers at 2120/2142). Handles mirrored/rotated cuts.
- **phase_ff.rs secondary arcs** → now emits **every** kept in-box arc (the `arc_point_sets` loop), not just the longest — the comment at L948 documents it.
- **builder_solid.rs single shell** → **scoped**: a lone shell is growth only if outward; for a negative corner-fan volume it consults a divergence-flux orientation test (L505+ is now that helper). +2 unit tests: outward cube → growth, reversed/inward cube → rejected.
- **same_domain.rs source_face guard** → now `same_source_complementary_split` (same source **AND** distinct interior points), so a coincident same-source duplicate still reaches the SD overlap test and is caught.

`merge_duplicate_edges` is byte-identical to main; gridfinity d1/d3/d4/d5 + coplanar caps green; all census cases analytic (sphere−cyl recovered); torus−box 5 faces / free=0 / render bd=0 / volume ≈1543; suite 2478. Merging.

PR #1054, opened 2026-07-08
fix(algo): close dovetail corner-clip intersect chord/arc lens

andymai:
## Summary

Closes the deferred `dovetail_corner_clip_intersect_is_watertight` case: the baseplate corner-rounding Intersect (slab-with-pockets × rounded-rect prism) now returns a compact, watertight, fully analytic result (41 faces, both corner barrel pieces preserved) instead of falling back to a 6116-facet mesh slab that poisoned every downstream dovetail connector boolean (~11 min per tile, non-manifold STL).

Two stacked GFA roots, which only close **together** (chord-fix-only: free 1→39; arc-fix-only: free 41; both: free 0):

### 1. Coplanar chord/arc lens (`phase_ff_coplanar.rs`)

The FF-coplanar phase projects boundary edges as straight 2D segments, so a rounded-corner boundary **arc** became a **chord** section — while the true arc already existed as an FF section (the barrel×cap circle, split at the operand's diagonal seam vertex). The `has_existing_section_at` midpoint dedup can never catch this: emitted circle arcs store the full-circle bbox, whose midpoint is the circle **center**. The resulting co-endpoint chord/arc lens breaks the wire weave — the chord lands in the outer wire and the true arc is orphaned as a zero-area slit (the free edge).

Fix: `matching_arc_section_exists` — skip a **Circle** boundary edge when its exact arc (same circle, same endpoints, shared face) already exists as a section. Line boundary edges are never skipped: a co-endpoint line/arc pair can be a genuine lens with material between the curves (the torus-box in-tube case).

### 2. Reverse-twin phantom arc splits (`face_splitter`)

`split_sections_at_t_junctions` normalized arc split parameters via `domain_with_endpoints`, which always returns the **CCW** span. Sections are pushed as forward/reverse pairs — for the reverse twin the CCW span is the long complement (315° for a 45° corner arc), so a point on the circle but *outside* the arc normalized to an interior `t` (45/315 = 1/7), and the split evaluator (shorter-arc convention) minted a **phantom vertex on the true arc's interior**. That desynced the coincident caps' partitions and killed their same-domain pairing.

Fix: `find_splits_on_section_arc` — section splits use the same shorter-arc convention as `evaluate_edge_at_t` (sections are ≤π by construction). Boundary-edge splitting deliberately keeps the CCW-domain convention: boundary arcs may genuinely exceed π — switching them regressed the d1/d3/d4/d5 gridfinity lip fuses (caught by the wasm canary during development, reverted).

## Verification

- `dovetail_corner_clip_intersect_is_watertight` un-ignored and green (full operations boolean, no fallback)
- Raw-GFA guard tightened from `free<=1` to `free==0` (renamed `cornerclip_intersect_raw_gfa_stays_watertight`)
- `cargo test -p brepkit-wasm --lib gridfinity` green (27/27 — the d4/d5 lip-fuse canary)
- Full workspace suite green (77 suites; `brepkit-render` `compute_mesh_lod` SIGSEGV is pre-existing on main in this environment, unrelated)
- clippy `-D warnings` clean, boundaries clean
- Roadmap skill updated in-PR (cornerclip row → CLOSED; socket-loft row updated — that case is a different mechanism and remains deferred, ops-level state unchanged)

## Expected downstream effect

The corner-rounding intersect is shared by all baseplate generators — the dovetailKey (bnd=108), fit-offset (bnd=184/144), snapClip (nm=14/16/11) scenario failures and the >25-min dovetail suite timeout all consume its mesh-fallback slab. Tool-side re-probe follows after release.

---
## Summary by cubic
Fixes the dovetail corner-clip Intersect by removing a chord/arc lens and a reverse‑twin arc split bug. The Intersect now returns a compact, watertight analytic result (41 faces; both corner barrel pieces) instead of a slow, non‑manifold mesh fallback.

- **Bug Fixes**
  - FF‑coplanar: added `matching_arc_section_exists` to skip projecting a Circle boundary edge as a straight chord when the exact arc section already exists; compares center, radius, endpoints, ARC MIDPOINT, and requires parallel normals so same‑center different‑plane circles don’t match. Line edges are never skipped.
  - Splitter: introduced `find_splits_on_section_arc` and used the shorter‑arc convention for Circle section splits only; kept the CCW domain for boundary arcs and all ellipse sections.
  - Tests: un‑ignored `dovetail_corner_clip_intersect_is_watertight` and tightened the raw‑GFA guard to `free == 0`; both tests pass.
  - Result: no mesh fallback; analytic output stays small and watertight.

<sup>Written for commit 617db9badf44dac16c4e410d65aafc7d8679e60a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1054?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Both P1 findings addressed in 4011140:

- **Complementary Arc Match**: `matching_arc_section_exists` now also compares the arc midpoint (evaluated via `domain_with_endpoints` on the boundary edge vs the existing section's `t_range` midpoint — the `arc_exists` pattern from `phase_ff.rs`). The complementary arc between the same endpoints no longer matches, so its chord section is kept.
- **Long Ellipse Split Loss**: ellipse sections are back on the domain-based `find_splits_on_ellipse`. `find_splits_on_section_arc` is now circle-only — only circle sections carry the FF emitter's ≤π-span guarantee that makes the shorter-arc parameterization sound.

Re-verified: both cornerclip tests green, `cargo test -p brepkit-wasm --lib gridfinity` 27/27, full workspace green, clippy clean.

PR #1117, opened 2026-07-18
fix(tessellate): keep self-intersecting planar caps watertight via fan fallback

andymai:
## Summary
Fixes the gridfinity "2×2 solid with slot (stadium) cutout" export producing a non-watertight STL (8 boundary edges) — the first of the solid-cutouts parity failures. Root-caused with a faithful native repro (operands captured from the real tool cut site).

## Root cause
**Tessellation defect, not a boolean bug.** `boolean(Cut, bin, stadium)` yields a watertight, fully-analytic B-Rep (F=60, euler=2, nm=0, bd=0). Boundary edges appear only when tessellating at fine deflection (0 @ defl=0.5 → 8 @ 0.05 — the count-grows-with-refinement signature).

The offending face is the bin's top ledge plane: a weaving, simply-connected planar annulus. At one corner the inner socket-foot cone rim (circle r=1.25) and the bin's outer rounded corner (circle r≈5.99) **genuinely intersect** — the inner rim bulges ~0.11mm *past* the outer arc, so the ledge pinches through zero to negative width. The projected boundary polygon is therefore self-intersecting.

`cdt_triangulate_simple` on such a polygon: CDT recovers the crossing constraints with a Steiner vertex, then the input-index mapping **silently drops** every triangle touching it → the pinch sliver is left untriangulated → single-use mesh edges → cracks.

## Fix
Detect that CDT introduced Steiner vertices (`mapped < cdt_triangles.len()`) and fall back to `fan_triangulate`, which uses only the original boundary vertices and is manifold by construction regardless of the self-overlap. Steiner vertices arise *only* for self-intersecting boundaries, so normal planar faces are unaffected. Minimal (+13/-1 in planar.rs); no L0–L2 code touched.

## Verification
- Faithful repro: boundary_edges **8 → 0** at defl 0.1/0.05/0.01; cut B-Rep unchanged (nm=0 bd=0).
- `brepkit-operations --lib tessellate` **72/0** (incl. new `pinched_ledge_prism_is_watertight`); `brepkit-operations --lib` **766/0**; `brepkit-wasm --lib gridfinity` **27/0** (canary); `brepkit-io` green; clippy `-D warnings` clean.

## Follow-up (not in scope)
The same Steiner-drop pattern exists in the holed CDT paths (`run_planar_cdt`, `tessellate_planar_shared_with_holes`) — no current repro self-intersects an *inner* wire, so untouched. Noted in the roadmap.

---
## Summary by cubic
Fixes tessellation cracks in self-intersecting planar caps by falling back to a fan when CDT inserts Steiner vertices, keeping caps watertight. The gridfinity “2×2 solid with stadium slot” now exports a watertight mesh.

- **Bug Fixes**
  - In `cdt_triangulate_simple`, detect Steiner insertion (`mapped < cdt_triangles.len()`) and fall back to `fan_triangulate`; also fall back if CDT maps no triangles. Only applies to self-intersecting boundaries; normal faces are unchanged.
  - Adds `pinched_ledge_prism_is_watertight` regression with captured polygons; asserts 0 boundary and non‑manifold edges at deflections 0.1 and 0.05.

<sup>Written for commit 77f44b1d8a50c9c430348e48d370fc697bcb6193. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/operations/src/tessellate/planar.rs
+++ b/crates/operations/src/tessellate/planar.rs
@@ -775,12 +775,14 @@ pub(super) fn cdt_triangulate_simple(positions: &[Point3], normal: Vec3) -> Vec<
     }
 
     let mut indices = Vec::with_capacity(cdt_triangles.len() * 3);
+    let mut mapped = 0usize;
     for &(v0, v1, v2) in &cdt_triangles {
         if let (Some(&i0), Some(&i1), Some(&i2)) = (
             cdt_to_input.get(&v0),
             cdt_to_input.get(&v1),
             cdt_to_input.get(&v2),
         ) {
+            mapped += 1;
             #[allow(clippy::cast_possible_truncation)]
             {
                 indices.push(i0 as u32);
@@ -790,7 +792,17 @@ pub(super) fn cdt_triangulate_simple(positions: &[Point3], normal: Vec3) -> Vec<
         }
     }
 
-    if indices.is_empty() {
+    // A boundary constraint can only cross another constraint when the polygon
+    // self-intersects (booleans occasionally emit a planar face whose outer
+    // wire pinches through zero width — two boundary arcs that overlap by a few
+    // hundred microns). CDT recovers crossing constraints by inserting a Steiner
+    // vertex at the crossing; triangles touching it have no input-vertex mapping
+    // and would be dropped here, leaving a hole. The Steiner vertex also splits
+    // the shared boundary edges, which cracks against the neighbouring faces.
+    // Fall back to a fan, which uses only the original boundary vertices and is
+    // manifold by construction (each boundary edge used once, each diagonal
+    // twice) regardless of the self-overlap.
+    if mapped < cdt_triangles.len() || indices.is_empty() {
         return fan_triangulate(n);
     }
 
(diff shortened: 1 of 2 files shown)


andymai:
Addressed both review findings by dropping the roadmap change from this PR (force-pushed 77f44b1 — now touches only planar.rs + tests.rs):
- **Greptile P2 (SKILL.md:131, 'fix pending' after the fix lands)** and **Copilot (SKILL.md:453, duplicated row)**: the roadmap's entire 'DEFERRED but ready' section is duplicated *and diverged* on `main` already (a pre-existing issue, 322 vs 346 lines) — that's what made my added row appear twice. Rather than expand a tessellation fix into a 320-line roadmap dedup, I removed the added entry. The stadium case's closure is recorded in git history + the PR body; the pre-existing section duplication is a separate cleanup.

The code fix and regression test are unchanged. Re-verified locally: repro bd 8→0, tessellate 72/0, operations 766/0, gridfinity canary 27/0, clippy clean.

PR #1119, opened 2026-07-19
fix(algo): salvage closed-circle cap sections in the planar face splitter

andymai:
## Summary
Fixes the gridfinity **screw-base export family** (task #12): 9 of 13 scenarios go from non-watertight to clean, and the 4×4 stress case drops from **1792 → 3** boundary edges. Root-caused end-to-end with a faithful native repro captured from the real tool pipeline.

## Root cause
The screw/magnet holes are cylinder cuts into the base socket — watertight in isolation. The failure is the final **socket → bin-body `Fuse`**: the socket's top face (Plane z=5) carries the drilled-hole rims as **closed-circle** sections, coincident with the body's bottom plane. `arrangement_regions_from_inputs` retains only inputs with a non-zero UV chord, and a **closed circle has start == end** (zero-length chord) → every hole cap is dropped. The impl's single-closed / all-Line fast path only fires for a *lone* circle, so two-or-more circles (or circles mixed with the socket's corner-cone arcs) lose their caps → the drilled cylinder rims are left as free edges → GFA opens (raw `bd=4`) → operations mesh-falls-back to hundreds of all-planar faces (the 112–1792 export boundary edges).

This was mis-diagnosed twice before landing here (first as a #696 tessellation crack, then as a same-domain inner-wire drop); both were refuted with literal data. The verified root is the closed-circle drop in the planar arrangement splitter.

## Fix
`split_face_2d` now wraps `split_face_2d_impl`: on a **plane** face it peels off **genuine** cap circles (closed `Circle` sections *strictly interior* to the outline — centre inside by more than the radius, which rejects the tangent corner-cylinder remnants the pave machinery leaves at faceted junctions), splits the face by the remaining open sections through the unchanged impl, then carves each cap into its containing sub-face via the existing `split_face_with_internal_loops` (disc cap + holed remainder). A lone cap circle still routes through the impl's fast path untouched, and faces with no interior cap circle are byte-identical to before. General to the coincident-planar-interface-with-drilled-holes class, not screw-specific.

## Verification
- Faithful native repro: raw `gfa::boolean(Fuse)` **bd 4 → 0**; operations result **672 all-plane mesh-fallback → 71 analytic faces, watertight**.
- Tool re-probe (overlay): **screw family 9/13 pass** (was 0/13); 4×4 stress 1792→3; **no regression** (solid-cutouts unchanged 5 fail/23 pass).
- Foils: `algo` 166/0, `operations` 767/0 (incl. new `fuse_capping_slab_preserves_drilled_hole_caps`, verified fails-without/passes-with), `wasm gridfinity` **27/0**, `io` green, clippy `-D warnings` clean.

## Follow-ups (not in scope)
4 residual screw scenarios: 4×4 stress (3 bd — a few near-grid-boundary holes the conservative interiority guard skips), 2 lightweight variants (448 — different base structure), and the scoop+label+lip permutation (132 — separate feature roots).

---
## Summary by cubic
Fixes planar face splitting to preserve closed-circle hole caps in coincident-plane booleans, preventing free edges and mesh fallbacks. Gridfinity screw-base exports: 9/13 now pass; 4×4 stress drops boundary edges from 1792 to 3.

- **Bug Fixes**
  - Salvage interior closed `Circle` sections on plane faces: peel caps, split by open sections, then carve via `split_face_with_internal_loops`.
  - Prevent zero-UV-chord circles from being dropped by salvaging them before `arrangement_regions_from_inputs`.
  - Keep the single-circle fast path; non-planar and cap-free faces unchanged.
  - Add regression `fuse_capping_slab_preserves_drilled_hole_caps` to assert watertight fuse and cylinder preservation.

- **Refactors**
  - Build containment polygon from traversal-ordered wire points for correct interiority on reversed edges; keep `PlaneFrame` convention unchanged.
  - Add `CAP_INTERIORITY_MARGIN` (5%) and single-pass section partitioning with stored cap centers for robustness.

<sup>Written for commit bfe7c188aeae5d1eca7ba9c28762b700a5bea9af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1119?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/algo/src/builder/face_splitter/mod.rs
+++ b/crates/algo/src/builder/face_splitter/mod.rs
@@ -3175,43 +3175,57 @@ pub fn split_face_2d(
     // point with an open arc section, and their "circle" is the corner cylinder
     // tangent to the face outline). Only genuine caps are peeled off; everything
     // else stays in `sections` and reaches the impl unchanged.
-    let wire_pts = collect_wire_points(topo, face.outer_wire());
+    // The FRAME must be built from the same points (and thus the same
+    // convention) the impl uses, so the sub-face UVs and the projected circle
+    // centres agree. The containment POLYGON, by contrast, must follow wire
+    // TRAVERSAL order — `collect_wire_points` takes every edge's stored
+    // `start()` and ignores the oriented-edge flag, so a wire carrying reversed
+    // edges would come back scrambled and the containment test meaningless.
+    let frame_pts = collect_wire_points(topo, face.outer_wire());
     let owned_frame;
     let cap_frame = if let Some(f) = frame {
         f
     } else {
         let normal = extract_plane_normal(&surface);
-        owned_frame = PlaneFrame::from_plane_face(normal, &wire_pts);
+        owned_frame = PlaneFrame::from_plane_face(normal, &frame_pts);
         &owned_frame
     };
-    let outer_poly: Vec<Point2> = wire_pts.iter().map(|&p| cap_frame.project(p)).collect();
+    let outer_poly: Vec<Point2> = collect_wire_points_oriented(topo, face.outer_wire())
+        .iter()
+        .map(|&p| cap_frame.project(p))
+        .collect();
 
-    let is_cap_circle = |s: &SectionEdge| -> bool {
-        if (s.start - s.end).length() >= tol.linear {
-            return false; // not a closed loop
-        }
-        let EdgeCurve::Circle(circle) = &s.curve_3d else {
-            return false;
+    // One pass: partition into cap circles (keeping each centre so the carve
+    // step never re-derives it) and everything else.
+    let mut cap_sections: Vec<SectionEdge> = Vec::new();
+    let mut cap_centers: Vec<Point3> = Vec::new();
+    let mut rest_sections: Vec<SectionEdge> = Vec::new();
+    for s in sections {
+        let cap_center = if (s.start - s.end).length() < tol.linear
+            && outer_poly.len() >= 3
+            && let EdgeCurve::Circle(circle) = &s.curve_3d
+        {
+            let center_uv = cap_frame.project(circle.center());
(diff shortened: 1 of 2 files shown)


andymai:
Addressed all four review findings in bfe7c188:

- **Copilot — `collect_wire_points` ignores oriented-edge direction** (real correctness issue, thanks): the containment polygon is now built from **traversal-ordered** wire points via a local `collect_wire_points_oriented`. Important subtlety found while fixing it: those same points feed `PlaneFrame::from_plane_face`, and the frame **must** keep the impl's convention or the sub-face UVs and the projected circle centres stop agreeing — switching both broke the regression test immediately. So the frame stays on the unoriented points and only the polygon is reordered. The shared `collect_wire_points` helper is left alone (other consumers depend on it).
- **P2 magic number**: `CAP_INTERIORITY_MARGIN` is now a documented module-scope constant (module scope also satisfies `items_after_statements`), explaining why 5% — a tangent corner-cylinder circle sits exactly one radius out, so the slack only absorbs float noise.
- **P2 triple evaluation**: sections are partitioned in a **single pass**.
- **P2 unreachable `_ => s.start` arm**: gone — cap centres are captured during that same pass and handed to the carve step, so it never re-derives them.

Re-verified after the changes: repro raw GFA `bd=0` and operations analytic (no fallback); `algo` 166/0, `operations` 767/0 (incl. the regression, still fails-without/passes-with), `wasm gridfinity` **27/0**, `io` clean, clippy `-D warnings` clean.

PR #1123, opened 2026-07-19
fix(algo): anchor closed-rim splits at the edge's own start angle

andymai:
## Summary
Fixes a **double-cover bug in closed-rim edge splitting**: a closed circle's pieces came back sweeping **4π instead of 2π**. Found while digging the gridfinity lightweight family; the fix is general, not lightweight-specific.

## Root cause
A CLOSED circle has no span between its endpoints, so `domain_with_endpoints` returns the circle's **intrinsic** domain `(0, 2π)` — anchored at the circle's angular origin, not at the edge's start point. `find_splits_on_circle`'s consumer chains pieces from `start_3d` in ascending `t`, so whenever those anchors differ (a cylinder seam away from the origin) **the edge's own start point reads as an interior split**. Instrumented output for a pad rim seamed at u=4.71239:

```
t0=0.00000 span=6.28319
  t=0.250000 u=1.57080 ... t=0.750000 u=4.71239  <-- the edge's OWN start
```

The 2π ring came back as **7 arcs sweeping 4π**, with the wedge between origin and seam covered twice — once inside the leading arc, again as a trailing forward/reverse pair.

## Fix (all gated to closed circles on a cylinder/cone)
1. Anchor `t` at the edge's **start angle**, signed by traversal direction, so `t` is monotone along the walk.
2. Interpolate interior split UVs within the edge's own span (`start_u → start_u ± 2π`) instead of a raw principal-value projection that drops joints back into `[0, 2π)` and loses phase coherence with neighbouring boundary edges.
3. Derive the tail piece's `end_uv` from that span — `sample_edge_to_uv` ignores orientation, so a reverse-traversed ring otherwise closes its period on the wrong side of its own start.

## Verification
- Regression: two unit tests pin the invariant directly — a closed rim's pieces must sweep **exactly 2π**, forward and reversed. Verified failing without the fix (`got 12.566…` = 4π).
- Lightweight repro: raw-GFA free edges **15 → 8**, and a pad-wall sub-face is recovered (cylinder 48 → 49), analytic preserved.
- Foils: `algo` **169/0**, `operations` **769/0**, `wasm gridfinity` **27/0** (d4 canary), `io` clean, clippy `-D warnings` clean.

## Scope — what this does NOT do
It does **not** close the lightweight export family. A separate, independently-verified root remains: a **missing FF section** — a floor-plane arc at z=−3.8 over `u ∈ (3.7518, 4.10219)` that should exist (confirmed by material flips in point classification) but is never generated, so the fuse still falls back. That is tracked as the terminal root for that family; this PR lands the closed-rim invariant fix on its own merits.

---
## Summary by cubic
Fixes double-coverage when splitting closed circular rims on cylinders/cones. Pieces now tile the ring once (2π) from the edge’s own start angle, matching traversal.

- Anchor split parameter to the edge start angle (signed by traversal).
- Interpolate interior split UVs within `start_u → start_u ± 2π`; set the tail piece `end_uv` from that span so reversed rims close on the correct side.
- Gate both anchoring and UV-span logic to closed circle edges on `Cylinder`/`Cone` surfaces.
- Tests assert an exact 2π sweep (forward and reversed) and now check forward UV continuity (shared joints, monotone u).
- In the gridfinity lightweight model, reduces raw-GFA free edges and restores a pad-wall sub-face.

<sup>Written for commit 56585280c797fe3f8b212a1078be90e77515e3fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/algo/src/builder/face_splitter/edge_splitting.rs
+++ b/crates/algo/src/builder/face_splitter/edge_splitting.rs
@@ -50,6 +50,18 @@ pub(super) fn split_boundary_edges_at_3d_points(
 
         let circle_iso_v_rim = matches!(edge.curve_3d, EdgeCurve::Circle(_))
             && circle_edge_is_iso_v_rim(&edge, surface, tol);
+        // A CLOSED rim on a cylinder/cone spans a whole period in the edge's own
+        // unwrapped u (`start_u` -> `start_u ± 2π`, signed by the traversal). A
+        // raw principal-value projection of each split point would drop the
+        // pieces back into `[0, 2π)`, so the interior joints lose phase
+        // coherence with the neighbouring boundary edges and the ring stops
+        // reading as a monotone walk across the cut-open strip. Interpolate in
+        // the edge's own span instead — the closed-edge counterpart of the
+        // `circle_iso_v_rim` branch, which covers only OPEN rims.
+        let closed_ring_dir = (matches!(edge.curve_3d, EdgeCurve::Circle(_))
+            && matches!(surface, FaceSurface::Cylinder(_) | FaceSurface::Cone(_))
+            && (edge.start_3d - edge.end_3d).length() < tol)
+            .then(|| if edge.forward { 1.0_f64 } else { -1.0_f64 });
         let mut prev_uv = edge.start_uv;
         let mut prev_3d = edge.start_3d;
         for &(t, pt) in &splits {
@@ -69,6 +81,11 @@ pub(super) fn split_boundary_edges_at_3d_points(
             };
             let split_uv = if let Some(f) = frame {
                 f.project(split_3d)
+            } else if let Some(dir) = closed_ring_dir {
+                brepkit_math::vec::Point2::new(
+                    std::f64::consts::TAU.mul_add(dir * t, edge.start_uv.x()),
+                    edge.start_uv.y(),
+                )
             } else if circle_iso_v_rim {
                 // Interpolate within the edge's own UV span: an iso-v rim's
                 // pcurve is u-linear, and a raw principal-value projection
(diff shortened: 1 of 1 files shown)


andymai:
Addressed all three findings in 56585280:

- **Copilot + Greptile P2 (both flagged the same thing — `closed_anchor` scope broader than `closed_ring_dir`)**: correct, and my PR description was inaccurate. `closed_anchor` fired for closed circles on ANY surface while its UV-span counterpart is gated to cylinder/cone. Since only the periodic-rim case is exercised and verified here, I **narrowed the anchor to match** (cylinder/cone) rather than broaden the UV branch — shipping unverified breadth in the splitter isn't worth it. Added a comment noting the same origin-vs-start mismatch is latent for a closed circle on another surface (a bore rim on a plane) and is deliberately out of scope until it has a repro.
- **Greptile P2 (forward test missing UV-continuity assertions)**: added them, mirroring the reversed test — the period closes ABOVE the start (`start_u + 2π`), joints are shared, and u increases monotonically. The forward case previously only checked 3D chaining and the 2π sweep.

Re-verified after the change: `algo` **169/0** (both rim tests green), `operations` **769/0**, `wasm gridfinity` **27/0**, `io` clean, clippy `-D warnings` clean — and the lightweight repro still improves identically (raw GFA **bd 15 → 8**, cylinder 48 → 49), confirming the narrower gate still covers the case this fix targets.

PR #1125, opened 2026-07-19
feat(math): solve parallel-axis cone × cylinder in closed form

andymai:
## Summary
Cone × cylinder had **no algebraic arm for the parallel-axis case**, so it fell through to the grid-seeded marcher — which returned **49 overlapping partial traces of what is a single curve**. This replaces that with an exact closed-form solution (2 branches).

## Root cause
The marcher's seed threshold is half the partner's diagonal, and the march-result dedup only consumes seeds the resulting polyline passes near. On a near-tangent parallel-axis configuration, many seeds each spawn their own fragment and none subsumes the others — hence 49 fragments for one curve.

## Fix
With parallel axes the axis separation `d` is constant, so at cone radius `ρ` the intersection satisfies

```
u = φ₀ ± acos((d² + ρ² − R²) / (2·d·ρ))
```

giving two exact branches parameterised by the cone's own `v`. Emitted directly instead of marching — exact, and cheaper than seeding + marching.

## Verification
- Regression: three tests asserting exactly the two branches are returned. Verified failing without the fix: `expected exactly the two branches / left: 49 / right: 2`.
- Foils: `math` **459/0**, `algo` **169/0**, `operations` **769/0**, `wasm gridfinity` **27/0** (d4 canary), `io` clean, clippy `-D warnings` clean.

## Scope
Found while digging the gridfinity **lightweight** family, where a magnet pad straddles a tapered socket foot and the 49 fragments starve that fuse of a usable section chain. **This alone does not change that result** — later links in that chain (FF graze-refinement density, open-section trimming to curved faces, boundary-split gating, and an operations Euler gate) are still open and are deliberately not in this PR. It lands as a standalone correctness/robustness fix for the whole parallel-axis cone×cylinder class.

### DIFF
--- a/crates/math/src/analytic_intersection.rs
+++ b/crates/math/src/analytic_intersection.rs
@@ -1199,7 +1199,7 @@ pub fn intersect_analytic_analytic_bounded(
 ) -> Result<Vec<IntersectionCurve>, MathError> {
     // Try algebraic specialization for known surface pairs before falling
     // back to the general marching approach.
-    if let Some(result) = try_algebraic_intersection(&a, &b)? {
+    if let Some(result) = try_algebraic_intersection(&a, &b, v_range_hint_a, v_range_hint_b)? {
         return Ok(result);
     }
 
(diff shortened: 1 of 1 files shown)


andymai:
Addressed all three findings in cb6798f9:

- **Greptile P1 — silent interpolation failure drops a branch**: agreed, and this one mattered most: emitting only the branch that happened to fit is the *same silent half-answer* the marcher's 49 fragments produced, i.e. exactly the section starvation this path exists to prevent. Changed to defer the whole pair (`return Ok(None)`) on any interpolation error so the caller's other paths handle it — never a partial result. Kept the `pts.len() < 4` skip, which is a genuinely absent branch rather than a failure, and said so in a comment.
- **Copilot — apex dropped when `d == R`**: correct. `cos_alpha` has `rho` in its denominator, so at `rho → 0` it is only meaningful as a limit: it tends to 0 (`alpha → π/2`) exactly when the cylinder passes through the apex (`d == R`), and diverges otherwise — where the clamp would manufacture a spurious `alpha` of 0 or π. Now the apex is kept in the `d == R` case (a genuine point, and the shared endpoint where the two branches meet) and still skipped otherwise.
- **Copilot — misleading doc on the coaxial deferral**: right that `intersect_analytic_analytic_bounded` never calls `exact_cone_cylinder`. It *is* called, but from `phase_ff.rs:2304`. Doc now states that: phase FF gets the exact circles, while a coaxial pair arriving through the other caller falls through to the marcher.

Re-verified after the change: `math` **459/0**, `algo` **169/0**, `operations` **769/0**, `wasm gridfinity` **27/0** (d4 canary), clippy `-D warnings` clean.

PR #1131, opened 2026-07-20
fix(heal): revert a unify pass that would orphan edges

andymai:
## The defect

`unify_same_domain` could turn a watertight shell into a holed one. Merging is an **optimisation** — it must never leave the shell worse connected than it found it.

Two phases can orphan edges:

1. a group whose reassembly loses a boundary edge, and
2. **`merge_collinear_edges`**, which rewrites only the NEW faces' outer wires. A neighbouring face still referencing the pre-merge edges is left unpaired.

The second is a one-sided topology edit and is the larger defect. This PR **bounds** its damage rather than repairing it — a two-sided rewrite is still worth doing, and I did not want to attempt that reshaping on the back of a guard.

## Why it went unnoticed

It is invisible on analytic solids, where collinear runs are rare, and severe on mesh-fallback ones, where thousands of coplanar facets merge at once.

Found digging the gridfinity `4×4 lite export (stress)` failure (48 STL boundary edges). Its base mesh-falls-back to a 15648-face solid, and a **single** drill through `compound_cut` merged 3184 faces and 1630 collinear edges — orphaning 4574:

| | faces | free edges |
|---|---|---|
| `boolean(Cut)` 1 drill (no unify) | 15650 | 0 |
| `compound_cut` unify=**false** | 15650 | 0 |
| `compound_cut` unify=**true** — before | 12466 | **4574** |
| `compound_cut` unify=**true** — after | 15650 | **0** |

The guard logs its reason: `reverting (3184 faces, 1630 edges) — unpaired edges would rise 0 -> 4574`.

## Why the revert is complete

Removed faces are only dropped from the shell's face list, never deleted from the arena, and the collinear pass only mutates the **new** faces' wires — which the revert discards. So restoring the original face list fully restores the prior topology.

## Verification

- `compound_cut_unify_still_merges_normally` — new; pins that the guard does **not** over-trigger: a slotted bar's coplanar fragments must still merge back (≤20 faces). Without it, a future change could make the guard fire everywhere and silently disable unify.
- `compound_cut_unify_keeps_bore_opening` (from the closed-curve fix, #1129) continues to pass.
- Full workspace suite green; `brepkit-heal` 86/0; gridfinity canary 27/0.
- `brepkit-render` GPU tests SIGSEGV identically on clean `main` — pre-existing, unrelated.

## Honest gap

**No compact synthetic reproduces the collinear-edge path yet.** I tried six shapes (slotted bars, coplanar pocket grids, stacked fuses, through-slot grids, a mesh-fallback cylinder∪cone, box mesh booleans); none trigger it — adjacent facets on curved surfaces are not coplanar so nothing merges, and coplanar box facets get merged during mesh→B-Rep conversion. The reproducer is the captured 15648-face solid, too large to commit. The new test therefore pins the guard's *other* half (no over-triggering); the orphaning path itself is verified only against the captured operands, as recorded above.

This is a bound on damage, not a scenario fix — `4×4 lite export (stress)` also needs the pad fuse to stop falling back, which is separate (and previously measured as expensive).

---
## Summary by cubic
Add a connectivity guard to the unify pass and roll back each phase independently to prevent orphaned edges. This keeps watertight shells intact on mesh-fallback solids while still merging valid coplanar faces.

- **Bug Fixes**
  - Measure edge pairing before/after face merges and again after collinear-edge merges; revert only the phase that increases unpaired edges. Snapshot new-face wires to undo collinear merges while keeping valid face merges; warn with before/after counts.
  - Replaced the slotted-bar check with a differential L-shaped fuse test: unify reduces face count vs no-unify and the result stays watertight.

<sup>Written for commit 29af4141028d933524c04f92884326d7aae372a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1131?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/heal/src/upgrade/unify_same_domain.rs
+++ b/crates/heal/src/upgrade/unify_same_domain.rs
@@ -235,24 +235,20 @@ pub fn unify_same_domain(
     }
 
     // Merging is an OPTIMISATION — it must never leave the shell worse connected
-    // than it found it. Snapshot the pairing first, apply the merge and the
-    // collinear-edge pass, then put the original faces back if either degraded
-    // it. Both phases can orphan edges: a group whose reassembly loses a
-    // boundary edge, and `merge_collinear_edges`, which rewrites only the NEW
-    // faces' wires — a neighbour still referencing the pre-merge edges is left
+    // than it found it. Each phase is applied, measured, and rolled back on its
+    // own so a bad phase does not discard a good one.
+    //
+    // Both phases can orphan edges: a group whose reassembly loses a boundary
+    // edge, and `merge_collinear_edges`, which rewrites only the NEW faces'
+    // wires — a neighbour still referencing the pre-merge edges is left
     // unpaired. That is invisible on analytic solids (few collinear runs) and
     // catastrophic on mesh-fallback ones (thousands of coplanar facets: 1630
     // edge merges orphaned 4574 edges on a drilled 15648-face solid).
-    //
-    // Reverting is complete: removed faces are only dropped from the shell's
-    // list, never deleted from the arena, and the collinear pass only mutates
-    // the new faces' wires, which the revert discards.
-    let mut original_faces: Option<Vec<FaceId>> = None;
-    let mut bad_before = 0;
+    let mut bad_after_faces = 0;
     if total_merged > 0 {
         let shell = topo.shell(shell_id)?;
         let current_faces: Vec<FaceId> = shell.faces().to_vec();
-        bad_before = unpaired_edge_count(topo, &current_faces)?;
+        let bad_before = unpaired_edge_count(topo, &current_faces)?;
 
         let mut final_faces: Vec<FaceId> = current_faces
             .iter()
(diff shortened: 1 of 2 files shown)


andymai:
Both P2 findings addressed in `29af4141`.

**P2 — all-or-nothing revert discarded healthy merges.** Each phase is now measured and rolled back on its own. `merge_collinear_edges` rewrites each wire in place under the same `WireId`, so snapshotting those wires undoes just that phase and keeps the face merges. On the captured 4×4 operands:

| | faces | free edges |
|---|---|---|
| unguarded | 12466 | **4574** |
| all-or-nothing guard | 15650 | 0 — but **all merges given up** |
| **per-phase guard** | **12466** | **0** |

All 3184 face merges kept, watertight.

**P2 — test assertion was an upper bound only.** You were right, and it was worse than 'may go undetected': the bound was **vacuous**. Rewriting it differentially immediately failed — the slotted bar it used merges *nothing* (18 faces with unify and without), so it would have passed no matter what the guard did.

It now compares the same cut with unify on and off, on an L-shaped fuse with corner cuts that genuinely merges (22 → 16 faces). It **fails against the previous all-or-nothing commit**, so it pins the phase split rather than just describing it.

PR #1200, opened 2026-07-24
perf(algo): reject coplanar-but-disjoint SD pairs with an oriented box

andymai:
## What

The planar geometric-containment pass (Step 3b of `detect_same_domain`) grid-prunes candidate pairs by axis-aligned AABB, then runs the full `planar_faces_overlap` containment test on every survivor. An axis-aligned box is a poor bound for a thin **diagonal** face: a diagonal strut's AABB is a large square that overlaps every lattice member it crosses. On a dense coplanar lattice (the kumiko wall pattern) the broad-phase barely prunes, and the pass runs tens of thousands of full containment tests per boolean that find nothing.

This adds a per-planar-face **oriented bounding box** (thickness axis pinned to the plane normal, in-plane axes from PCA), built from the same outer-wire sampling already used for the AABB, and rejects a candidate pair when the two tol-expanded OBBs are disjoint — before the full test.

## Why it's safe

An OBB conservatively contains its face (both are tol-expanded), so OBB-disjoint proves the faces share no point, and `planar_faces_overlap` could only return `false`. The reject is **result-preserving**. Candidate iteration order and the union-find sequence are unchanged, so representative selection is identical.

## Measurements

Captured 180-strut kumiko fuse (replayed natively from the parity-capture cache):

- coplanar full-test count per boolean: **~40k -> ~10k (approx 4x)**
- sequential fuse chain: **40.9s -> 36.4s (~11%)**, **byte-identical result volume** (13776.89)

The AABB grid, candidate ordering, and all downstream SD logic are untouched.

## Verification

- `brepkit-algo` unit tests: pass
- Full `brepkit-io` + `brepkit-operations` foil: **1208 tests pass** (8 skipped are `#[ignore]` ready-repros), including the SD-load-bearing gridfinity lip/socket/honeycomb fuses
- clippy clean on the changed crate

---
## Summary by cubic
Adds an oriented-bounding-box early reject to the planar containment pass to skip coplanar‑but‑disjoint pairs. On the 180‑strut kumiko capture this yields ~4x fewer full overlap tests and ~11% faster runtime, with byte‑identical results.

- **Performance**
  - Build a tol‑expanded OBB per planar face from the same outer‑wire samples as the AABB (normal pinned to the plane; in‑plane axes via PCA).
  - After AABB pruning, reject pairs whose OBBs don’t intersect; OBBs are stored in a dense `Vec<Option<Obb3>>` by sub‑face index (no hashing in the hot loop), and candidates from the AABB set always have OBBs. Minor: preallocate outer‑wire sampling.
  - Verified: `brepkit-algo` unit tests and the `brepkit-io` + `brepkit-operations` suite (1208) pass; clippy clean.

<sup>Written for commit ea5acdfa146639cbfd23277fe219d8e983ffb692. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1200?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/algo/src/builder/same_domain.rs
+++ b/crates/algo/src/builder/same_domain.rs
@@ -246,9 +246,11 @@ pub fn detect_same_domain<S: BuildHasher>(
     {
         // Per planar face, from one outer-wire sampling: an AABB (drives the
         // grid broad-phase) and a tol-expanded OBB (tighter oriented reject in
-        // the pair loop below).
+        // the pair loop below). `planar_obbs` is indexed directly by sub-face
+        // index — dense and hasher-free in the hot loop; an entry is `Some`
+        // exactly for the faces that also entered `planar_aabbs`.
         let mut planar_aabbs: Vec<(usize, brepkit_math::aabb::Aabb3)> = Vec::new();
-        let mut planar_obbs: HashMap<usize, brepkit_math::obb::Obb3> = HashMap::new();
+        let mut planar_obbs: Vec<Option<brepkit_math::obb::Obb3>> = vec![None; n];
         for (idx, surf) in surfaces.iter().enumerate() {
             let Some(FaceSurface::Plane { normal, .. }) = surf else {
                 continue;
@@ -267,7 +269,7 @@ pub fn detect_same_domain<S: BuildHasher>(
                 continue;
             };
             planar_aabbs.push((idx, aabb));
-            planar_obbs.insert(idx, obb);
+            planar_obbs[idx] = Some(obb);
         }
         // Broad-phase: only test pairs whose AABBs overlap. Two planar faces
         // that don't share 3D space (expanded by tol for boundary-coincident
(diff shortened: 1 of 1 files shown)


andymai:
Thanks — all three addressed in ea5acdfa:
1. `planar_obbs` is now a `Vec<Option<Obb3>>` indexed by sub-face index (no hashing in the pair loop).
2. Fixed the misleading comment — both pair indices come from `planar_aabbs`, so their OBB entries are always `Some`; there is no degenerate-face fall-through.
3. `face_outer_wire_points` now preallocates `edges * SD_EDGE_SAMPLES`.

Full io + algo foil re-run green (221 + 191 tests).

PR #1211, opened 2026-07-24
docs(roadmap): custom-shape re-probe; add STEP solid health example

andymai:
Roadmap maintenance for the custom-shape family, per the skill's own rule that a session closing or discovering an item must update it.

## Verified: the T lip band cut reaches the tool

The entry said "Tool-side fuse re-probe pending release". Done, via a local-build overlay (md5-verified in both `node_modules` resolution paths):

| `3x3 T with lip` export | published 2.128.2 | local HEAD (#1209) |
|---|---|---|
| boundary edges | 173 | **0** |
| non-manifold | 6 | **0** |
| wall-clock | 10.4 s | **0.84 s** |

L and U with lip were already clean and are byte-identical across the two builds; exactly 1 of the 23 `customShape` triangle counts moved. The fix is surgical.

## Discovered and root-localized: the O-ring is a TESSELLATION bug

`3x3 O-shape (ring)` fails on **both** builds, so it is pre-existing rather than fallout. Single-variable isolation narrowed it well past "the O-ring export is broken":

- minimal broken config is `base=flat, lip=OFF` — `bnd=0 nm=88` at 1800 tris in **64 ms**, far too fast for a boolean fallback;
- the STEP capture replays faithfully (`tris=1800` matches the tool) and **the B-Rep is fine**: `F=47`, 24 cylinder + 23 plane, `brep_free=0 brep_over=0`;
- all 88 folded edges lie on **one** face — the `z=21` top plane carrying 3 inner wires — each used 3x;
- that face's mesh covers area **13944.0** against a true face area of **3207.97** (4.35x over-cover), while the no-hole control matches its own area to 0.007%.

So the planar mesher is not removing that face's hole interiors, and the `lip=on` variant's 24 s fallback is downstream consumption of the folded mesh rather than a second defect. Synthetic straight-edged multi-hole faces (1-4 holes) are all clean, so the trigger involves the arc-cornered wires — recorded as the starting point.

## Why the suite missed all of this

Every custom-shape case is `assert: 'snapshot'` — it pins `triangleCount` and never checks validity, so a leaky mesh and a correct one are indistinguishable to it. The STL edge-use probe that does discriminate is documented in `parity-benchmarking` (where it is reusable) rather than inline in the roadmap.

## Contents

- `.claude/skills/roadmap` — the two entries above (net +14 lines; kept to the skill's own one-entry-one-pointer rule).
- `.claude/skills/parity-benchmarking` — the STL edge-use probe recipe.
- `crates/io/examples/step_solid_report.rs` — new diagnostic that produced the numbers above: for any STEP file it reports face/surface mix, by-edge-id use counts, and position-welded mesh edge use, so "broken B-Rep" and "fine B-Rep, folded tessellation" can be told apart (they have identical STL symptoms). `DUMP_NM`, `MESH_AREA_AT`, `PLANES_AT` env vars for drill-down.

No production code paths touched.

### DIFF
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -573,14 +573,38 @@ stepped tip→centroid by fractions that overshoot a ~1.2mm annulus into the hol
 no valid probe → unstable ray-cast → Inside. Fix (`classifier/mod.rs`): deep centroid
 fractions first (honeycomb stacked-cap foil needs them), then a thin-band absolute-nudge
 fallback near the tip. Band now single-covered, translation-invariant, vol 6090.8 (was
-the doubled 20108.8, which `tship_lipcut` had enshrined as "watertight" — its by-edge-id
+the doubled 20108.8, which `tship_lipcut_inmem` had enshrined as "watertight" — its by-edge-id
 gate is blind to position-duplicate doubling). Benign residual: bottom tiles as ring + 2
 tiny T-armpit pieces (3 faces) from redundant FF-coplanar concave-corner sections — exact
-watertight tiling, not the defect. Tool-side fuse re-probe pending release.
-DURABLE detector reaffirmed: `solid_volume` is TRANSLATION-VARIANT on a doubled/malformed
+watertight tiling, not the defect. TOOL-SIDE RE-PROBE DONE (2026-07-24): 3x3 T export
+bnd=173 nm=6 @10.4s -> bnd=0 nm=0 @0.84s; L and U already clean and byte-identical, 1
+of 23 customShape triangle counts moved. DURABLE detector reaffirmed: `solid_volume` is TRANSLATION-VARIANT on a doubled/malformed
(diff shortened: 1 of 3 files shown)


.claude/skills/roadmap/SKILL.md
```diff
@@ -573,15 +573,39 @@ stepped tip→centroid by fractions that overshoot a ~1.2mm annulus into the hol
 no valid probe → unstable ray-cast → Inside. Fix (`classifier/mod.rs`): deep centroid
 fractions first (honeycomb stacked-cap foil needs them), then a thin-band absolute-nudge
 fallback near the tip. Band now single-covered, translation-invariant, vol 6090.8 (was
-the doubled 20108.8, which `tship_lipcut` had enshrined as "watertight" — its by-edge-id
+the doubled 20108.8, which `tship_lipcut_inmem` had enshrined as "watertight" — its by-edge-id
```
andymai, line 576: Fixed — the identifier now reads `tship_lipcut_inmem` to match the file on disk.

PR #1212, opened 2026-07-24
fix(operations): tessellate nested inner wires by even-odd nesting depth

andymai:
Closes the custom-shape **O-ring** export defect found while re-probing the T-lip fix (#1209). Despite the symptoms, this is **not** a boolean bug.

## Symptom and localization

The tool's `3×3 O-shape (ring)` exported closed-but-non-manifold and ~24x slower than its L/T/U siblings. Single-variable isolation found the minimal broken config is `base=flat, lip=OFF` — **1800 tris, 88 folded edges, 64 ms**, far too fast for a mesh fallback. The captured STEP replays faithfully (`tris=1800` matches the tool) and the B-Rep is **fine**: `F=47`, 24 cylinder + 23 plane, no free and no over-shared edges.

Every folded edge sat on one face: the `z=21` wall top, whose three inner wires **nest** — the cavity opening contains the island band around the central 1u hole, which contains that hole.

## Two stacked defects

1. **Colliding seeds.** Both solid tessellation paths seeded hole flood-removal at each inner wire's vertex **centroid**. A gridfinity bin is centred on the origin, so all three concentric wires share the centroid `(0,0)`: the first flood took the innermost cell and the other two found it already gone. The 13152-unit cavity was never removed. (The non-shared path already used `find_interior_seed`; only the two solid paths used a centroid.)
2. **Nesting ignored.** Removing one cell per inner wire is wrong regardless of seeding — nesting alternates material and void, so only **odd-depth** wires bound a hole. Fixing seeding alone erased the island band and swung to `boundary=100`.

Stored winding cannot classify them (a boolean can emit a hole wound like its outer — the documented hole-winding hazard), so `hole_removal_seeds` decides by geometric nesting depth.

## Cell arithmetic (outer ~15736)

| cell | area | |
|---|---|---|
| outer..I27 | 595.0 | material |
| I27..I28 | 13152.7 | void |
| I28..I29 | 198.1 | material |
| inside I29 | 1789.8 | void |

| | predicted | measured |
|---|---|---|
| before | 13945.8 | **13944.0** |
| seed fix only | 595.0 | **590.1** |
| even-odd (this PR) | 793.1 + chord | **799.9** |

Both failure modes match the model to chord error, so the mechanism is confirmed rather than inferred.

## Result

O-ring: `mesh_bnd 0`, `mesh_nm 88 -> 0`. The no-hole control is byte-identical.

## Verification

- 1413 tests green across `operations` + `io` + `algo`
- `cargo test -p brepkit-wasm --lib gridfinity` — 27 passed, including the d1-d5 lip canaries
- New fixture `crates/io/tests/oring_nested_holes.rs` **verified to fail on the parent commit** with exactly `88 folded edges`, and to pass here. It pins area as well as edge counts, because both failure modes pass an edge-use check on their own.

---
## Summary by cubic
Fix tessellation for planar faces with nested inner wires by removing holes based on even‑odd nesting depth and seeding floods per wire. Adds a safe fallback when a computed seed lies outside its wire, keeping O‑ring exports watertight and manifold while leaving non‑hole cases unchanged.

- **Bug Fixes**
  - Classify inner wires by geometric nesting; remove only odd‑depth loops (true holes).
  - Use per‑wire `find_interior_seed` to avoid centroid collisions on concentric wires; if a seed is outside its wire, flood from it unconditionally to avoid misclassification.
  - Add `hole_removal_seeds(...)` and use it across all planar tessellation paths; include a regression test and STEP fixture to pin area and edge counts.

- **Performance**
  - Bound containment checks with per‑wire bounding boxes and skip the scan when only one inner wire, keeping the pass near‑linear for many disjoint holes.

<sup>Written for commit f167793380ca71ee97444d8d258c05f06ae6045c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/operations/src/tessellate/planar.rs
```diff
@@ -703,6 +695,89 @@ fn find_interior_seed(polygon: &[brepkit_math::vec::Point2]) -> brepkit_math::ve
...
+                p.iter().fold(
+                    (f64::MAX, f64::MAX, f64::MIN, f64::MIN),
+                    |(x0, y0, x1, y1), q| {
+                        (x0.min(q.x()), y0.min(q.y()), x1.max(q.x()), y1.max(q.y()))
+                    },
+                )
+            })
+        })
+        .collect();
+
```
andymai, line 746: Good catch — the centroid fallback is unvalidated, and while the *old* code tolerated a bad centroid (it flooded from it regardless), the new nesting depth measured from such a point is meaningless.

Fixed in the follow-up commit: the seed is now validated against its own wire, and if it fails we drop back to the pre-nesting behaviour — flood from it unconditionally — rather than guess a depth. That keeps degenerate wires exactly as they behaved before this PR.

PR #1213, opened 2026-07-24
test(operations): root-map the tangent-section-circle boolean fallback

andymai:
The census's `cone ∪ box` is the **only remaining primitive-boolean mesh fallback**, and it was not root-mapped anywhere. Mapping it showed it is not a cone problem at all. Three `#[ignore]` repros plus a roadmap row; no production code changes.

## It is a general tangency defect, not cone-specific

`make_cone(6, 2, 12)` has radius exactly **4** at z=6, where the 8×8 box's bottom sits — the section circle is **inscribed in the box square, tangent to all four walls**.

A plain **cylinder** reproduces it identically:

| configuration | F | validate |
|---|---|---|
| cylinder r=4, tangent d=8 box | 14 | **4 non-manifold** |
| cylinder r=4, clear d=10 box | 8 | clean |

And the cone sweep isolates the same variable:

| configuration | F | z=0 disc | validate |
|---|---|---|---|
| tangent, d=8 zb=6 | 10 | **0** | 4 non-manifold |
| circle strictly inside, d=10 zb=6 | 8 | 1 | **clean** |
| tangent, d=6 zb=9 | 10 | **0** | 4 non-manifold |
| circle strictly inside, d=8 zb=9 | 8 | 1 | **clean** |
| circle crossing walls, d=6 zb=6 | 11 | 1 | 1 free edge |

Two independently chosen tangent configurations fail *identically*; every non-tangent one is clean. This is the tangential-contact class — the same family as the intwidth-tangency closure — and gridfinity geometry is full of it, so the blast radius is wider than one census row.

## What breaks on the cone case

Raw GFA nearly succeeds: `F=10`, 1 cone + 9 planes, **zero free edges**, rejected only for 4 non-manifold edges.

The **box side is correct** — the pinched square-minus-inscribed-circle region tiles into 4 curvilinear-triangle faces, each wall splitting at its tangency point.

The **cone side loses its z=0 base rim**, two symptoms of one cause:

1. Both cone plane faces vanish. The z=12 top disc correctly (it lies inside the box), but the **z=0 base disc (r=6) is nowhere near the box and must survive** — all 9 result planes are box-derived.
2. The cone lateral's inner wire repeats **its own outer wire** — the same four z=6 arc edge ids — instead of the base rim. Hence each arc used 3×.

The remainder is being closed by duplicating the wrong rim. First probe for whoever picks it up: why the z=0 rim/disc is dropped — classification, or never emitted.

## Discovered en route

A section circle **crossing** the walls (rather than tangent or interior) is a *separate* defect — 1 free edge — previously unmapped. Recorded alongside.

## Contents

- `cone_union_box_should_be_analytic` — ready-repro asserting the z=0 base disc survives and the result is manifold; currently fails on the first (`left: 0, right: 1`).
- `diag_cone_box_tangency_sweep`, `diag_cylinder_box_tangency` — the two tables above.
- One roadmap row under "DEFERRED but ready".

CI is unaffected: all three are `#[ignore]`d.


andymai:
Follow-up: measured the width of the failure band, which reshapes the fix.

`diag_tangency_epsilon_band` — cylinder r=4 against a box of half-width 4+eps:

| eps | result |
|---|---|
| −1e-3, −1e-5 (gross crossing) | **clean** |
| −1e-7 | 10 free edges |
| −1e-9 | 6 non-manifold + 7 free |
| **0 (exact tangency)** | 4 non-manifold |
| +1e-9 … +1e-3 (any clearance) | **clean** |

Gross crossing and gross clearance both work. Only **sub-tolerance penetration through exact tangency** breaks, over a band of ~1e-7 — the linear tolerance.

That is the intwidth-tangency signature, and its near-tangent root collapse in `Circle3D::intersect_segment` already ships. So the residual defect is **downstream of the collapse**: how a section circle split at pure tangency points (where the partner touches but does not cross) gets turned into regions.

It also corrects something I wrote earlier in this PR. I had called the "circle crossing the walls" case a separate unmapped defect based on the cone at d=6/zb=6 (1 free edge). The cylinder is clean under gross crossing, so that cone result is taper-specific rather than a general crossing defect — the roadmap row now says so.

crates/operations/src/boolean/tests.rs
```diff
@@ -6333,3 +6333,168 @@ fn compound_cut_coaxial_pair_clusters_match_sequential() {
...
+    let faces = brepkit_topology::explorer::solid_faces(&topo, result).unwrap();
+    let z0_discs = faces
+        .iter()
+        .filter(|&&fid| {
+            let f = topo.face(fid).unwrap();
+            f.surface().type_tag() == "plane"
+                && topo.wire(f.outer_wire()).unwrap().edges().iter().all(|oe| {
+                    let e = topo.edge(oe.edge()).unwrap();
+                    topo.vertex(e.start()).unwrap().point().z().abs() < 1e-9
+                        && topo.vertex(e.end()).unwrap().point().z().abs() < 1e-9
```
andymai, line 6381: Agreed — a vertical edge starting on z=0 would have counted. Both endpoints are now required, in the ready-repro and in the sweep. The repro still fails on the same assertion (`left: 0, right: 1`), so the check was not masking anything.

andymai:
Mechanism now measured rather than inferred, which pins where the fix must go.

`Circle3D::intersect_segment`, r=4 circle against a vertical segment at x:

| x | hits |
|---|---|
| 4.0 (exact tangency) | **1** — collapsed foot |
| 3.9999999 (sub-tolerance crossing) | **1** — collapse fires |
| 3.999 (gross crossing) | 2 |
| 4.001 (clear) | 0 |

The intwidth near-tangent collapse returns **one** hit for both true tangency and sub-tolerance penetration — correct in itself, and load-bearing for the intwidth closure. But it means each of the four walls contributes one hit, so `closed_circle_boundary_crossings` splits the section circle at four **touch** points, and the consumer treats them as crossings.

That is exactly why the failure band equals the collapse's range: `[-1e-7, 0]` is precisely where it fires.

**So the fix cannot be to stop splitting.** The box bottom genuinely needs those splits — its square-minus-inscribed-circle region is pinched at those four points and correctly tiles into four curvilinear triangles. The quadric side is what mis-assembles: it has to recognise the four arcs as one closed rim rather than an open chain. Roadmap row updated accordingly.

andymai:
Sharpest characterization yet — the trigger is the **number** of tangency points, not tangency itself.

`diag_tangency_count`, cylinder r=4 against boxes tangent on 0–4 walls:

| tangent walls | result |
|---|---|
| 0 | **clean** |
| **1** | **clean** |
| 2 | 4 **free** edges |
| 4 | 4 **non-manifold** edges |

One touch-split leaves the closed section circle a single closed edge (P→P), which the consumer handles. Two or more split it into a **chain** of arcs, and the quadric side fails to close the loop — which is why the same root surfaces as *open* at two walls and *over-shared* at four.

Combined with the earlier measurements this gives a complete picture for whoever implements the fix:

- `intersect_segment` correctly yields one hit per tangent (or sub-tolerance-crossing) wall — do not touch it, the intwidth closure depends on it;
- the box bottom genuinely needs those split points, since its square-minus-inscribed-circle region is pinched there;
- the defect is the quadric side treating ≥2 touch-splits as an open chain instead of one closed rim.

andymai:
Found why the existing rescue never fires, which turns this from "first probe" into a concrete fix shape.

`split_cylinder_band_by_arrangement` already exists and is gated on `greedy_broken` (`face_splitter/mod.rs` ~5007). It is never consulted here because the greedy result **looks structurally valid**: the cone/cylinder face's outer wire is `{16, 22, 23, 24}` and its inner wire is the *same* edge ids `{23, 22, 16, 24}`. A face with an outer wire and one inner wire trips no existing check, so nothing sets the broken flag.

Proposed trigger: **an inner wire whose edge set duplicates the outer wire's**. That is keyed to a demonstrated failure signature rather than a predicted one, so it cannot demote a working case — the same cheap route past the splitter's web of mutual calibrations that `loops_have_out_and_back` took for the label-sockets closure.

I have not implemented it: validating a `greedy_broken` change needs the full foil set (d4, honeycomb pcut3, divider-lip, the nub fixtures), and the tool is currently occupied by an export-integrity run. Recording it so the next session starts from the fix rather than the symptom.

andymai:
Pre-validated the proposed detector empirically, with a mixed result worth recording.

I added a temporary env-gated probe in `validate_boolean_result` that reports any face whose inner wire's edge set equals its outer wire's, then ran the foils. (Probe since reverted; nothing in this PR touches production code.)

**It is safe.** The signature fires on the 4-tangent-wall case and on **nothing else** — zero hits across the entire `brepkit-io` fixture suite (d4, honeycomb, divider-lip, the nub fixtures, all the gridfinity captures) and zero across the `brepkit-operations` lib tests. So adding it as a `greedy_broken` trigger cannot demote anything that currently passes.

**But it is incomplete.** The 2-tangent-wall variant — which fails with 4 *free* edges rather than 4 non-manifold — does **not** carry the duplicate-wire signature. So it needs its own trigger, and a fix that only adds this detector would close the 4-wall case while leaving the 2-wall case broken.

That is the kind of thing worth knowing before implementing rather than after: the obvious detector covers one of the two symptoms, not both. Roadmap row updated to say so.

andymai:
Correction to my earlier "same root, two symptoms" claim — I dumped the 2-wall case's structure and it is materially worse than I described.

**4 tangent walls:** the quadric face *survives* but is mis-wired — outer wire `{16,22,23,24}`, inner wire the same ids. Result is closed but non-manifold.

**2 tangent walls:** the quadric face is **dropped entirely**. Result is `F=8, all planar` — the whole lower cylinder is missing, and the surviving box bottom carries used-once circle and line edges (an open shell).

So the two share a trigger but not a demonstrated cause: one keeps a bad face, the other loses the face. I stated they were one root with two symptoms; that was an inference, not a measurement, and the structure dump does not support it. The roadmap row now says to treat them as two cases until shown otherwise.

This also reinforces the earlier finding that the duplicate-wire detector is incomplete — it cannot possibly catch the 2-wall case, because there is no surviving face to carry the signature.

PR #1215, opened 2026-07-24
docs(roadmap): record the export-integrity baseline and the kumiko root

andymai:
First full before/after over the tool's 408-test `export-integrity` matrix — its own version of the STL edge-use oracle (zero boundary edges, bounded non-manifold). Docs plus one diagnostic harness; no production code.

## Result

| | published 2.128.2 | local main |
|---|---|---|
| passed | 365 | **371** |
| failed | 43 | **37** |

Fixed: both `3×3 T with lip` cases (#1209), `3×3 O-shape (ring) with lip` and `O-shape + magnet base + lip` (#1212), plus pathfinder / permutation-matrix / scoop rows.

**No regressions.** One scenario looked like one — `wall patterns › slots carves 3×3×5 walls (scale 0.5)` passed in the baseline and failed after. Isolated A/B says otherwise: **63.3 s on published, 64.3 s on main**. It is a timing-borderline scenario whose full-suite verdict depends on cache warmth and machine load, not on the kernel. Recorded so nobody chases it as a correctness bug.

## Kumiko's 14 failures are ONE root

Families: kumiko 14, permutation matrix 7, custom-shape 6, solid cutouts 3, then singles.

`goma carves a 1×1×6 bin` runs ~48 s and then throws `recursive use of an object detected which would lead to unsafe aliasing in rust` — wasm borrow poisoning. The scenarios after it inherit the poisoned kernel; two surface as `Shape handle has been disposed`. Only `mitsukude bold` (bnd=571) has an independent assertion.

**This corrects an existing roadmap row** which records that poisoning class as "NOT REPRODUCIBLE on 2.124.13". It reproduces on 2.128.2. The panic-capture hardening already shipped, so `lastPanicMessage()` should name the underlying panic — that is the next probe.

**Refuted en route:** the captured goma `compound_cut` is *not* the culprit. All 180 tools replay natively in 11.8 s with `F=1146, free=0, over=0` — a clean result. Harness added as `crates/io/examples/replay_kumiko_goma.rs` with the refutation in its doc comment, so the next session doesn't burn an iteration replaying it expecting a repro.

## A counting caveat

The baseline's classified failure kinds (23 boundary-edge, 2 non-manifold, 4 poisoning, 1 timeout) sum to **30 of 43** — about a dozen failures carry no recognised error form, most likely cascade casualties but unconfirmed. Kumiko already shows how badly the raw number misleads: 14 failures, one root. The roadmap now says never to quote the failure count as a defect count.

---
## Summary by cubic
Recorded the 408‑test `export-integrity` baseline (2.128.2: 365 pass/43 fail; local `main`: 371 pass/37 fail) with no regressions. Corrected the kumiko root to a generation‑side perf overrun that triggers a test timeout and wasm borrow‑poisoning (not a kernel panic); added `crates/io/examples/replay_kumiko_goma.rs` and fixed capture path joins with `PathBuf`.

- **New Findings**
  - The cost is in `generateBin`: ~849 s total, ~847 s in generation; tessellation/STL are negligible.
  - Bottleneck is the boolean chain, not the mesher; `compound_cut` capture replays clean in 11.8 s (F=1146, free=0, over=0).
  - Structure: `kumikoWrapBuilder.ts` carves per family over an accumulating region, repeating `cutAll`; next step is to count/time those `cutAll` calls during one goma run.

<sup>Written for commit e936ac8e7f77e05d70fc1caae0471c69401ca9ad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/io/examples/replay_kumiko_goma.rs
+++ b/crates/io/examples/replay_kumiko_goma.rs
@@ -1,0 +1,92 @@
+#![allow(clippy::print_stdout, clippy::expect_used, missing_docs)]
+//! Replay the captured kumiko `goma` compound_cut operands natively.
+//!
+//! The tool's `kumiko › goma carves a 1×1×6 bin` export runs ~48s and then
+//! throws "recursive use of an object detected which would lead to unsafe
+//! aliasing in rust" — wasm borrow poisoning, i.e. a panic inside the kernel.
+//! Native replay turns that into a real panic with a backtrace.
+//!
+//! REFUTED 2026-07-24: this captured `compound_cut` is NOT where the poisoning
+//! happens. Full 180-tool replay completes in 11.8s with F=1146, free=0,
+//! over=0 — a clean result. The goma export failure is somewhere else in that
+//! scenario's chain, so do not replay this capture expecting a repro.
+//!
+//! Capture: `~/.cache/brepkit-parity-captures/2026-07-23/kumiko-goma/`
+//! Usage: `cargo run --release --example replay_kumiko_goma -p brepkit-io -- [N]`
+
+use std::collections::HashMap;
+use std::path::PathBuf;
+use std::time::Instant;
+
+use brepkit_io::arena_io::deserialize_solid;
+use brepkit_operations::boolean::compound_cut;
+use brepkit_topology::Topology;
+use brepkit_topology::edge::EdgeId;
+use brepkit_topology::explorer::solid_faces;
+
+fn main() {
+    let dir = PathBuf::from(std::env::var("CAPTURE_DIR").unwrap_or_else(|_| {
+        format!(
+            "{}/.cache/brepkit-parity-captures/2026-07-23/kumiko-goma",
+            std::env::var("HOME").expect("HOME")
+        )
+    }));
+    let limit: usize = std::env::args()
+        .nth(1)
+        .and_then(|a| a.parse().ok())
+        .unwrap_or(usize::MAX);
+
+    let mut topo = Topology::new();
+    let region = deserialize_solid(
+        &std::fs::read(dir.join("cut1-region.bin")).expect("region"),
+        &mut topo,
+    )
+    .expect("region parse");
+
+    let mut tools = Vec::new();
+    for i in 0.. {
+        let p = dir.join(format!("cut1-tool{i}.bin"));
+        if !p.exists() || tools.len() >= limit {
+            break;
+        }
+        tools.push(
+            deserialize_solid(&std::fs::read(&p).expect("tool"), &mut topo).expect("tool parse"),
+        );
+    }
+    println!("loaded region + {} tools", tools.len());
+
+    let t0 = Instant::now();
+    let result = compound_cut(
+        &mut topo,
+        region,
+        &tools,
+        brepkit_operations::boolean::BooleanOptions::default(),
+    );
+    let ms = t0.elapsed().as_millis();
+
+    match result {
+        Ok(sid) => {
+            let faces = solid_faces(&topo, sid).expect("faces");
+            let mut mix: HashMap<&str, usize> = HashMap::new();
+            let mut uses: HashMap<EdgeId, usize> = HashMap::new();
+            for &fid in &faces {
+                let f = topo.face(fid).expect("face");
+                *mix.entry(f.surface().type_tag()).or_default() += 1;
+                for wid in std::iter::once(f.outer_wire()).chain(f.inner_wires().iter().copied()) {
+                    for oe in topo.wire(wid).expect("wire").edges() {
+                        *uses.entry(oe.edge()).or_default() += 1;
+                    }
+                }
+            }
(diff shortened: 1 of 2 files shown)


crates/io/examples/replay_kumiko_goma.rs
```diff
@@ -0,0 +1,97 @@
...
+use std::time::Instant;
+
+use brepkit_io::arena_io::deserialize_solid;
+use brepkit_operations::boolean::compound_cut;
+use brepkit_topology::Topology;
+use brepkit_topology::edge::EdgeId;
+use brepkit_topology::explorer::solid_faces;
+
+fn main() {
+    let dir = std::env::var_os("CAPTURE_DIR").map_or_else(
```
andymai, line 28: Fixed — the fallback now builds with `var_os` + `PathBuf::push`, so it neither panics on an unset HOME nor bakes in separators. Verified the harness still loads and replays (20 tools: F=161, free=0, over=0).

andymai:
Followed up on the three fixes I listed without attributing, since the poisoning cascade could plausibly have explained them as ordering artifacts. It doesn't — they are genuine.

**Kumiko runs last.** Its block is the final one in the run (summary lines 34–47), so the `goma` poisoning happens at the end and cannot contaminate any earlier family.

**The extra fixes failed with real assertions on a healthy kernel:**

- `pathfinder › subtract group: top z-index member carves a hole in the base` — `non-manifold edges: expected 4 to be +0`
- `scoop two-variable › 2×2 shallow curved scoop (run > height)` — `non-manifold edges: expected 2 to be +0`

Both are **non-manifold** failures on hole-carving geometry, which is exactly the signature a folded tessellation produces and exactly what the nested-hole fix in #1212 addresses. So the causal story is coherent rather than coincidental.

Recording the check because "N failures became N−6" is the kind of number that is easy to over-claim when a cascade is present in the same run.

andymai:
**Correction — I had the kumiko root wrong, and the probe refuted it.**

I wrote that `goma` panics and poisons the kernel, and proposed `lastPanicMessage()` as the next probe. I ran that probe on published 2.128.2 in isolation. Result:

```
export OK in 849351ms, 2715284 bytes
lastPanicMessage: (none)
```

The export **succeeds** — 2.7 MB of valid STL, no panic, nothing for the panic hook to report. But it takes **849 seconds (14 minutes)**.

So the root is **perf, not a crash**. The `recursive use of an object detected which would lead to unsafe aliasing in rust` seen in the suite is a *consequence*: the export blows through vitest's per-test timeout, the abandoned async generation chain stays pending, and it re-enters the kernel concurrently with the next test. The later kumiko scenarios then inherit the poisoned object (two surfacing as `Shape handle has been disposed`).

Two things this changes:

1. **Do not chase this as a panic.** `catch_unwind` and `lastPanicMessage()` have nothing to report — I would have sent the next session down that path.
2. **The earlier "poisoning class reproduces" correction was itself misleading.** The *symptom string* reproduces, but not the panic-driven mechanism the original roadmap row describes. The row I added has been rewritten to say so.

Where the time goes is still open: native `compound_cut` on the captured operands is only 11.8 s, so the other ~14 minutes are elsewhere in the tool's goma chain and need a per-stage capture. That is the real next probe.

The roadmap entry in this PR now leads with the 849 s measurement rather than the panic story.

andymai:
Located the 849 s. **It is generation, not export.**

```
generateBin(forExport=false): 847047ms  tris=60412
```

847 of the 849 seconds are in `generateBin` alone, so tessellation and STL writing are noise. That rules out the mesher and points at the boolean chain.

**Structure** (`kumikoWrapBuilder.ts` ~line 411) — the carve is not one `cutAll`:

```js
for (const family of [...families, fillings]) {
  const carved = unwrap(cutAll(region, family, { trackEvolution: false }));
  region = carved;
}
```

One `cutAll` per family over an **accumulating** region, and that whole function runs **per perimeter chunk**. So there are many such calls, each starting from a more complex region than the one I captured. My native replay (180 tools, 11.8 s, F=1146 clean) was a single one of them.

**What is measured vs. what is not**, since I have already been wrong twice on this scenario:

- Measured: the 847 s split; the single captured `cutAll` at 11.8 s; the loop structure in the source.
- **Not** measured: whether the total is simply the sum of those calls, or whether brepjs's bisecting `cutAll` (try the whole batch, on failure split the tool list and recurse) is thrashing on a batch that fails.

Those imply different fixes, so the next probe is to count and time the actual `cutAll` invocations and their tool counts during one goma generation — rather than assume the single capture is representative. Recorded that way in the roadmap.

PR #1216, opened 2026-07-24
test(io): measure region-complexity scaling in the kumiko carve

andymai:
Follow-on to #1215's kumiko root. Adds a `CHUNK=n` mode to the goma replay harness and records what the measurements actually show — including a hypothesis of mine they refute. No production code.

## cutAll is only 23% of the time

Wrapping `cutAll` across a whole goma generation (`vi.mock` + `importOriginal`):

```
generateBin total: 911238ms
cutAll calls: 32   summed: 205715ms (23% of total)   failed: 0
tool counts: 4,6,7,180,6,3,14,80  ×4
```

- **Zero failures**, so brepjs's bisecting `cutAll` is **not** thrashing — that hypothesis is dead.
- Structure is 4 chunks × 8 families, matching the source.
- **~705 s (77%) of generation is not in `cutAll` at all**, and is still unlocated.

## Batching is real but secondary

`CHUNK=n` replays the captured 180 tools in n sequential batches:

| batching | total | per-batch |
|---|---|---|
| 1 × 180 | **11.7 s** | — |
| 3 × 60 | 15.8 s | 1.7 → 5.3 → 8.8 |
| 6 × 30 | 18.6 s | 0.4 → 0.8 → 2.4 → 4.5 → 6.1 → 4.4 |

All three end at the identical `F=1146`, so this is pure batching cost. The same 30 tools cost **393 ms at F=168 but 6103 ms at F=1131** — cost grows steeply with region complexity, which makes the tool's per-family loop over an accumulating region the pessimal shape.

**But that cannot recover the missing 77%.** I previously suggested fusing the families into one `cutAll` as the fix; on these numbers that addresses part of 23%, not the bulk. Corrected here and in the roadmap.

## Also unexplained

The 180-tool call takes **33.5 s in-tool vs 11.7 s natively** — 3× on identical work. Not chased yet.

Next probe is to wrap the rest of the brepjs ops rather than just `cutAll` — candidates for the 705 s are the per-strut sketch/extrude/translate prism construction and `fuseAllOrNull` in `wallPatternClips`.

---
## Summary by cubic
Adds a chunked replay mode (`CHUNK=n`) to the kumiko goma harness to measure how `compound_cut` cost scales with region complexity; also corrects the roadmap by removing a stale claim and recording measured results (cutAll ~23% of total; batching adds overhead). Testing-only; no production code changes.

- **New Features**
  - `crates/io/examples/replay_kumiko_goma.rs`: support `CHUNK=n` to run `compound_cut` in N sequential batches, printing per-batch timing and face count; guard `CHUNK=0`/empty tools and surface face-enumeration errors.

<sup>Written for commit b0b98da5c579741c7c786290bf0df42b4ae02d51. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Both fixed.

**Zero chunk** — `CHUNK=0` (or empty tools) now reports and returns instead of panicking on a zero batch size. Verified: `CHUNK=0 with 180 tools: nothing to do`.

**Silent `F=0`** — this one mattered more than a normal nit, because the harness exists to produce a complexity/timing series and a fabricated zero would have quietly corrupted it. Face enumeration errors are now printed and abort the run rather than being folded into the data. Re-ran `CHUNK=3` to confirm the numbers are unchanged (1631ms→F=309, 5269ms→F=880, 8792ms→F=1146).

PR #1217, opened 2026-07-24
docs(roadmap): attribute goma cost to one cutAllBisect call

andymai:
Continues the kumiko/goma dig from #1215 and #1216. Full op-level attribution, two refuted candidates, and a correction to a claim I recorded as settled. Docs only.

## Where the time actually goes

Wrapping every brepjs op across one goma generation at h=4 — 99.9% accounted, total 292.7 s:

| op | calls | ms | % |
|---|---|---|---|
| **`cutAllBisect`** | **1** | **203552** | **69.5** |
| `cutAll` | 32 | 88543 | 30.2 |
| everything else | 7 | ~318 | 0.1 |

The bulk is **one call** — the step applying the pattern solids to the bin — not the per-family carve.

The phase split agrees: `buildKumikoWallPatterns` is only 32.8% of generation, and within it prism construction is **322 ms (0.1%)**.

## Two candidates refuted

- **Prism construction** ("thousands of small sketch/extrude/translate kernel ops") — 322 ms. Dead. I had written this into the roadmap as a leading candidate.
- **Batching as the main cause** — the per-family loop is genuinely the pessimal shape (#1216 measured 1×180 = 11.7 s vs 6×30 = 18.6 s), but at 30.2% it cannot explain the total.

## A correction

An earlier probe wrapped `cutAll`, saw 32 calls with zero failures, and I concluded there was **no bisect thrash** — and recorded that in the roadmap.

That was invalid. `cutAllBisect` is a **separate export**, and its internal retries go through an internal ops reference my wrapper never saw. The hypothesis was **untested**, not refuted. The measurement was real; the claim simply covered a code path the instrument couldn't observe.

## Next probe

Read `cutAllBisect`'s own `BatchBisectTelemetry`:

- `batchAttempts` >> `batchSucceeded`, or a large pairwise count → the batch is **failing** and the 203 s is retry cost — a correctness bug wearing a perf costume.
- one attempt → honest N-way work on a hard batch.

Different fixes, so it is left open rather than guessed.

## Tooling note

V8 `--cpu-prof` does **not** work in this harness: vitest's fork pool drops it via both `NODE_OPTIONS` and `poolOptions.forks.execArgv`, and `vite-node` is not installed. Two attempts (~20 min) produced only idle parent-process profiles. Recorded so the next session doesn't repeat it — use `vi.mock` wrapping, and check the wrapper covers the path you intend to claim about.

---
## Summary by cubic
Pins kumiko/goma cost to one 8‑tool `cutAllBisect` (~203.5s), not bisect thrash. RAW GFA: cut 0 leaves 30 free edges and cut 1 aborts, forcing mesh fallback (all‑planar, open); all free edges sit on a 0.05 mm slab at x=17.00–17.05, now identified as plane‑vs‑cylinder; prism build is 322 ms.

- **New Features**
  - Added `crates/io/examples/replay_cut_capture.rs` to replay `compound_cut` captures; supports `RAW=1` to call analytic GFA, `DUMP_FREE=1` to print free‑edge locations, and `XSCAN=<x>` to list X‑normal planes near a target.

<sup>Written for commit d35773fe13db5a154f43bc09b27e3bb9b701dbb7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -663,10 +663,12 @@ confirms it: `buildKumikoWallPatterns` is only 32.8%, of which prism constructio
 the "thousands of small sketch/extrude ops" candidate is REFUTED. CORRECTION: an earlier probe wrapped
 `cutAll`, saw 32 calls with zero failures, and concluded there was no bisect thrash — INVALID, because
 `cutAllBisect` is a separate export whose internal retries use an internal ops reference the wrapper never
-saw. That hypothesis was untested, not refuted. NEXT PROBE: read `cutAllBisect`'s own
-`BatchBisectTelemetry` (batchAttempts vs batchSucceeded, pairwise count) — attempts >> successes means a
-FAILING batch whose retry cost is the 203s (a correctness bug wearing a perf costume), one attempt means
-honest N-way work. TOOLING NOTE: V8 `--cpu-prof` does NOT work here — vitest's fork pool drops it via both
+saw. That hypothesis was untested, not refuted. (5) BISECT TELEMETRY (the right instrument this time): `{totalInputs:8, batchAttempts:1,
+batchSucceeded:1, singletonFallbacks:0, failedInputs:[]}` — **NO thrash**, one attempt, one success. So
+**ONE `cutAll` of just EIGHT tools takes 203.5s**. Those 8 tools are the carved kumiko lattice bands
+(each ~F=1146 per the captured replay), so this is a single honest N-way boolean of the bin body against
+8 very complex solids — squarely a BREPKIT perf defect, not tool-side structure. THE TARGET IS NOW ONE
+CALL WITH NINE OPERANDS: capture body + the 8 pattern solids via serializeSolid and replay natively. TOOLING NOTE: V8 `--cpu-prof` does NOT work here — vitest's fork pool drops it via both
 NODE_OPTIONS and poolOptions.forks.execArgv, and vite-node is not installed; two attempts produced only
 idle parent-process profiles. Use `vi.mock` wrapping instead, and make sure the wrapper actually covers
 the call path you intend to claim about. REFUTED: the captured goma `compound_cut`
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -684,10 +684,16 @@ dev-dependency of `brepkit-io`, so io examples can reach it): **RAW cut 0 = 231m
 for the SAME cut — the analytic engine is 12x faster and keeps every cone and cylinder, but leaves
 **30 free edges**, so `validate_boolean_result` rejects it and the mesh fallback runs. **RAW cut 1
 fails outright: "assembly failed: open growth shell with 20 faces would be dropped; aborting
-analytic assembly".** So the 203s is the CONSEQUENCE of those 30 free edges. THE TARGET IS NOW:
-why does cut 0 leave 30 free edges (F=494), and why does cut 1 drop an open 20-face growth shell?
-Fixing that makes goma analytic AND ~12x faster per cut, and removes the broken-fallback
-consumption at the same time. TOOLING NOTE: V8 `--cpu-prof` does NOT work here — vitest's fork pool drops it via both
+analytic assembly".** So the 203s is the CONSEQUENCE of those 30 free edges. (8) THE 30 FREE EDGES ARE LOCALIZED (`DUMP_FREE=1`): every one lies at
+x = 17.00 or x = 17.05 — a single **0.05mm-thin slab** — with y in [-21.95, -17.15] and z in
+[2.70, 12.37]. Curve mix is mostly Line plus a few Ellipse/Circle, on plane and cylinder faces;
+the ellipses each BRIDGE the two planes (e.g. (17.00,-20.75,12.34)->(17.05,-20.75,12.37)), so this
+is a sliver between two near-parallel faces 0.05mm apart, where the lattice band's edge meets the
+bin's corner cylinders. 0.05mm is FAR above the 1e-7 linear tolerance, so this is a genuine
+thin-feature failure, not numerical noise. THE TARGET IS NOW that one sliver region: why the
+analytic assembly cannot close a 0.05mm slab there, and whether cut 1's "open growth shell with 20
+faces" is the same region. Fixing it makes goma analytic AND ~12x faster per cut, and removes the
+broken-fallback consumption at the same time. TOOLING NOTE: V8 `--cpu-prof` does NOT work here — vitest's fork pool drops it via both
 NODE_OPTIONS and poolOptions.forks.execArgv, and vite-node is not installed; two attempts produced only
 idle parent-process profiles. Use `vi.mock` wrapping instead, and make sure the wrapper actually covers
 the call path you intend to claim about. REFUTED: the captured goma `compound_cut`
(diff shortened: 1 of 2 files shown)


andymai:
Telemetry landed, and it **refutes the thrash hypothesis properly this time** — with an instrument that actually covers the path.

```
cutAllBisect[0] 203520ms ok=true
telemetry: {"totalInputs":8,"batchAttempts":1,"batchSucceeded":1,
            "singletonFallbacks":0,"failedInputs":[]}
```

One attempt, one success, zero singleton fallbacks, no failed inputs. There is no retry cost hiding in there.

**So the finding is: a single `cutAll` of just EIGHT tools takes 203.5 seconds.**

Those 8 tools are the carved kumiko lattice bands — each around `F=1146` per the captured replay in #1216 — so this is one honest N-way boolean of the bin body against 8 very complex solids. That makes it **a brepkit perf defect**, not tool-side structure. The per-family batching shape (30.2%) is a real but secondary contributor; this single call is the headline.

The dig has narrowed from "a 14-minute scenario" to **one call with nine operands**, which is about as well-scoped as a perf target gets. Next step is to capture body + the 8 pattern solids with `serializeSolid` and replay natively, giving a fixture that runs without the tool at all.

Roadmap updated with the telemetry and the target.

andymai:
Captured the nine operands and replayed them natively. The result reframes goma again — and turns up a correctness bug alongside the perf one.

Capture: `~/.cache/brepkit-parity-captures/2026-07-24/goma-bisect/` (4.1 MB, kept out of the repo like the kumiko-goma capture). Harness: `crates/io/examples/replay_cut_capture.rs`.

```
base:  F=78   {cone:12, cylinder:24, plane:42}    ← bin body
tool0: F=663  all plane                           ← carved lattice band
tool1: F=726  all plane
```

| tools | ms | F | mix | free | over |
|---|---|---|---|---|---|
| 1 | 2824 | 1003 | all plane | 0 | 0 |
| 2 | 6187 | 1519 | all plane | **756** | **320** |
| 3 | 12659 | 1920 | all plane | 723 | 333 |

**(a) It is the mesh fallback, not N-way boolean work.** Every result is all-planar even though the base carries 12 cones and 24 cylinders — the analytic types are destroyed, which is the fallback signature. So the 203 s is fallback meshing growing superlinearly with tool count, and the real fix is to stop falling back at all: find why GFA rejects bin-body × lattice-band.

**(b) From two tools on, `compound_cut` returns `Ok` with a broken solid** — 756 free edges, 320 over-shared. That is a native repro of the already-open *"mesh-boolean fallback emits OPEN meshes that get CONSUMED"* row, which until now had no small reproducer.

Worth noting (b) means goma is not only slow — the geometry it produces is unsound, and downstream ops consume it. That is likely why the kumiko family's failures looked so varied.

Next probe is the GFA rejection reason for `base × tool0`, which needs an algo-level harness: `brepkit-io` cannot reach `brepkit_algo` under the layer rules, so the probe has to live in `brepkit-operations`.

andymai:
**Root found — and it is a correctness bug, not a perf bug.**

Added a `RAW=1` mode that calls `gfa::boolean` directly, bypassing the ops-level gate and its mesh fallback. (`brepkit-algo` is already a dev-dependency of `brepkit-io`, so no layer change was needed.)

```
RAW cut 0: 231ms  F=494  {cone:12, cylinder:24, plane:458}  free=30  over=0
RAW cut 1: 676ms  ERR "assembly failed: open growth shell with 20 faces
                       would be dropped; aborting analytic assembly"
```

Compare the ops path on that same first cut: **2824 ms, F=1003, all-planar.**

So the analytic engine is **12× faster and preserves every cone and cylinder** — it simply leaves **30 free edges** out of 494 faces, which `validate_boolean_result` correctly rejects, sending the whole thing to the mesh fallback. The second cut does not even get that far.

That means the 203 s is downstream of a small topological defect. Fixing those 30 free edges would, in one change:

- make goma analytic instead of all-planar,
- cut ~12× off each of these booleans,
- and remove the broken-fallback consumption from #1215's kumiko family (the `Ok`-but-756-free-edges result), since the fallback would no longer run.

**The target is now:** why does cut 0 leave 30 free edges at F=494, and why does cut 1 drop an open 20-face growth shell?

That is where this dig stops and a fix begins — it needs the usual foil set, so I am not starting it inside this docs PR. The chain from "14-minute scenario" to here is recorded in the roadmap entry.

andymai:
Localized the 30 free edges with a `DUMP_FREE=1` mode. They are not scattered — they are all in **one tiny region**.

Every single one lies at **x = 17.00 or x = 17.05**, a **0.05 mm-thin slab**, with y ∈ [−21.95, −17.15] and z ∈ [2.70, 12.37].

```
free line    on plane    (17.05,-18.41, 3.16)->(17.05,-19.55, 3.16)
free ellipse on plane    (17.00,-20.75,12.34)->(17.05,-20.75,12.37)
free line    on cylinder (17.00,-19.55, 2.70)->(17.00,-19.55, 3.19)
free circle  on plane    (17.05,-20.75, 2.70)->(17.00,-20.75, 2.70)
…
```

The ellipse and circle edges are the tell: each one **bridges the two planes** (17.00 ↔ 17.05), so this is a sliver between two near-parallel faces 0.05 mm apart, where the lattice band's edge meets the bin's corner cylinders. Curve mix is mostly `Line` with a few `Ellipse`/`Circle`, on both plane and cylinder faces.

**0.05 mm is four orders of magnitude above the 1e-7 linear tolerance**, so this is a genuine thin-feature failure rather than numerical noise — which also means it is unlikely to be fixed by any tolerance widening.

So the target has narrowed from "goma is slow" to **one 0.05 mm sliver region in one boolean**. Open questions for the fix:

1. why the analytic assembly cannot close a 0.05 mm slab there;
2. whether cut 1's `open growth shell with 20 faces` is the same region failing harder.

Roadmap updated. Stopping the dig here — the fix is a GFA change and needs the foil set, which does not belong in a docs PR.

andymai:
One more step with an `XSCAN` mode — where do the two sides of the 0.05 mm slab come from?

```
base:  X-normal planes near 17: []
tool0: X-normal planes near 17: [17.05]
tool1: X-normal planes near 17: [16.05 … 17.00 … 17.90]
```

The base has **no** X-normal plane near x=17 at all. tool0 contributes exactly one, its cut plane at **x=17.05**.

The 30 free edges come from cut 0, which uses tool0 only — so the **x=17.00 side is not a plane**. The dump confirms it: `free line on cylinder (17.00,−19.55,2.70)`. That is the base's corner **cylinder** surface.

So the sliver is between **tool0's cut plane at x=17.05 and curved base geometry reaching x≈17.00** — a plane-versus-cylinder thin sliver. That is plausibly the same family as the tangent-section-circle row in #1213, but I want to be explicit that **the mechanism is not proven** — the geometric coincidence is suggestive, nothing more, and today has repeatedly punished that kind of inference.

Handoff target, unchanged in shape but sharper: why the analytic assembly cannot close a 0.05 mm plane-vs-cylinder slab in that one region, and whether cut 1's `open growth shell with 20 faces` is the same region failing harder.

PR #1218, opened 2026-07-25
test(io): show four closed outlines are missing in the goma sliver

andymai:
Continues the goma dig from #1217, which ended at "30 free edges in a 0.05 mm plane-vs-cylinder sliver." Adds `FREE_LOOPS=1` and `XSCAN=<v>` modes to the replay harness. Diagnostic only — no production code.

## Exactly four faces are missing

```
free components=4   odd-degree vertices=0   (every chain closed)
  component 1: 7 vertices
  component 2: 7 vertices
  component 3: 7 vertices
  component 4: 9 vertices
```

Chaining the 30 free edges by shared vertex gives **4 components, all closed** — zero odd-degree vertices. So these are not ragged partial boundaries but four well-formed polygon outlines (7–9 vertices each) where a face should be and is not, all inside the 0.05 mm slab.

That is the same class as cut 1's `assembly failed: open growth shell with 20 faces would be dropped` — the assembler dropping faces in this region, harder on the second cut.

## Where the slab comes from

`XSCAN=17.0` on each operand:

```
base:  X-normal planes near 17: []
tool0: X-normal planes near 17: [17.05]
tool1: X-normal planes near 17: [16.05 … 17.00 … 17.90]
```

The base has none. tool0 contributes its cut plane at x=17.05. Since the free edges come from cut 0 (tool0 only), the x=17.00 side is **not a plane** — the dump shows `free line on cylinder (17.00,−19.55,2.70)`, the base's corner cylinder. So it is a plane-vs-cylinder sliver, not two parallel planes.

I want to be explicit that this **resembles** #1213's tangent-section-circle family but the mechanism is **not proven**; the resemblance is geometric coincidence only.

## Why this matters

The chain from #1215/#1217: the kumiko family's 14 export-integrity failures reduce to one 203 s `cutAll`, which is 203 s only because the analytic path is rejected for these 30 free edges and the mesh fallback runs instead. The analytic result is **12× faster and keeps all 12 cones and 24 cylinders**.

So recovering these 4 faces would make goma analytic, cut ~12× per boolean, and remove the broken-fallback consumption (`Ok` with 756 free edges) at the same time.

**Target for the fix:** why those 4 sliver faces are dropped from the analytic assembly.

---
## Summary by cubic
Proves four simple closed loops bound the missing faces in the goma sliver via a degree‑2 check, dumps each loop’s geometry, and shows the defect is op‑independent and present across all pattern bands. Adds debug-only toggles and a ready‑repro test to reproduce the 30‑free‑edge failure; diagnostic only.

- Debug flags: `FREE_LOOPS=1` (chain and count closed components), `LOOP_GEOM=1` (dump per‑loop curves), `OP` (`cut|fuse|intersect`), and `TOOL=<i>` (single‑tool cut). Even tools: F=494 free=30 over=0; odd tools: assembler aborts (“open growth shell with N faces…”).
- New test: ignored ready‑repro at `crates/io/tests/goma_wall_band_cut_inmem.rs` with fixtures `goma_wall_base.bin` and `goma_wall_band.bin` that cuts one band and demonstrates the 30 free edges while preserving curved analytic faces.

<sup>Written for commit 9fcc058747dd060f675ff5f417f9888c996d0457. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
You are right, and this was a real logical error on my part — thank you for catching it twice.

Zero odd-degree vertices only rules out open chain endpoints. A degree-4 junction (figure-eight, or two outlines meeting at a shared vertex) is even-degree and would have passed my check while not being a simple closed outline — and in that case "four components" would not have meant "four missing faces" either. The conclusion did not follow from the evidence I gave for it.

Replaced the parity test with the condition that actually proves it — **every vertex at degree exactly 2** — and re-ran:

```
free components=4  degree histogram(deg:count)=[(2, 30)]  all_degree_2=true
  components of 9, 7, 7, 7 vertices
```

All 30 vertices are degree 2, so the four components really are simple closed outlines and "four missing faces" holds. The conclusion survives; the justification did not, and is now stated correctly in both the probe output and the roadmap entry.

The probe now also prints the full degree histogram rather than a single boolean, so a future junction case is visible rather than silently passing.

andymai:
Added `LOOP_GEOM=1` to dump each missing outline's edges. All four share one signature — component 1:

```
5 lines    at x=17.050   (y −21.95…−20.75, z 6.765…7.595)
1 line     at x=17.000   (17.000,−20.750,7.624)→(17.000,−20.750,6.794)
2 ellipses bridging      17.000 ↔ 17.050 at y=−20.750
```

So each loop **straddles the 0.05 mm gap**: most edges lie in tool0's cut plane at x=17.05, one edge sits on the far side at x=17.00, and two short ellipse arcs bridge between them.

That makes the loops **non-planar** — the missing faces span from the cut plane across to the base cylinder, rather than being flat slivers in either surface. Each covers roughly 1.2 × 0.83 mm with a 0.05 mm depth excursion.

Useful for the fix because it rules out the simplest reading: these are not degenerate flat slivers that a planar-face guard dropped. They are small curved patches with well-formed boundaries, which the assembler declines to build.

Sharpened target: **why four non-planar 0.05 mm patches, each with a valid closed boundary, are dropped from the analytic assembly** — and whether cut 1's `open growth shell with 20 faces` is the same patches failing harder.

andymai:
Added an `OP=` switch to test a discriminator: if the faces survive under a different boolean op, the splitter created them and classification dropped them; if every op loses them, the splitter never made them.

```
OP=cut        F=494   free=30  over=0
OP=fuse       F=1155  free=29  over=1
OP=intersect  ERR "open growth shell with 6 faces would be dropped"
```

**The defect is op-independent** — all three fail in the same region, and Intersect's error is the same family as cut 1's `open growth shell with 20 faces`.

That points **away from classification**. Classification selects different subsets per op, so a classifier bug would not produce near-identical free-edge counts for Cut and Fuse. It points instead at the shared upstream: the splitter/section machinery emitting a partition that cannot be assembled under any op.

So these are **not** faces that were built and then wrongly classified out — the partition is defective before classification runs.

**Target narrowed again:** the FF/section/split stage for this 0.05 mm plane-vs-cylinder sliver, not the classifier.

(Process note: my first attempt at this measurement silently did nothing — `cargo fmt` had reshaped the code so my patch anchor missed, and all three runs were still `Cut`. They returned byte-identical numbers, which is exactly what a working op-independent result would also look like. Caught it only because identical-to-the-digit output across three different ops was too clean. Worth flagging: a patch that fails to apply and a real null result can be indistinguishable in the output.)

andymai:
Added `TOOL=<i>` to cut the base by a single tool. The result is systematic — an even/odd split:

```
tool0  F=494 free=30 over=0      tool1  ERR open growth shell with  9 faces
tool2  F=494 free=30 over=0      tool3  ERR open growth shell with 22 faces
tool4  F=494 free=30 over=0      tool5  ERR open growth shell with 23 faces
tool6  F=494 free=30 over=0      tool7  ERR open growth shell with 36 faces
```

**Every one of the eight tools fails**, in exactly two modes. The even tools produce byte-identical results — they are congruent bands on the bin's four walls, and the capture file sizes agree (evens ~650 KB, odds ~380 KB), so the split is the two kumiko diagonal families.

So this is not one awkward tool. It hits **100% of the pattern bands**, which is why the mesh fallback runs for all eight and compounds to the 203 s.

Two consequences for whoever takes the fix:

- The repro is cheap. A single tool reproduces in **~230 ms** — no need to run the 203 s eight-tool call, or the 292 s generation, or the 850 s scenario.
- One fix plausibly clears both modes if they share a root. The identical even-tool signatures **suggest** that, but do not prove it, and the odd-tool face counts (9/22/23/36) differ enough to leave room for two distinct causes.

andymai:
Added a committed ready-repro so the defect is discoverable without the capture cache.

`crates/io/tests/goma_wall_band_cut_inmem.rs` — `#[ignore]`d, so CI is unaffected — plus 691 KB of operands (`goma_wall_base.bin` 47 KB, `goma_wall_band.bin` 660 KB). That is in line with the existing largest fixtures (`halfsockets_export_deferred.bin` is 587 KB), and it is one tool rather than all eight, which would have been 4.1 MB. The remaining seven bands stay in `~/.cache/brepkit-parity-captures/2026-07-24/goma-bisect/` and replay through `replay_cut_capture.rs`.

Current behaviour:

```
assertion `left == right` failed: cut must be closed;
30 free edges send the whole kumiko family to the mesh fallback
  left: 30   right: 0
```

The test also asserts `curved >= 30` and **that passes** — the analytic surfaces are preserved. That assertion is deliberate: it documents why this is worth recovering rather than conceding to the fallback, and it will fail loudly if a future change "fixes" the free edges by degrading to planes.

Full `brepkit-io` suite: 223 passed, 1 skipped (the new ignored repro).

The doc comment carries the whole measured chain — 850 s scenario → timeout poisoning → 14 kumiko failures → one 203 s `cutAll` → mesh fallback → 30 free edges → 4 non-planar patches in a 0.05 mm plane-vs-cylinder sliver — so the next session starts from the fix rather than re-deriving it.

PR #1219, opened 2026-07-25
test(io): show tool0 drops nothing — its faces are never created

andymai:
Follow-on to #1218. Adds a `SHELL_LOG=1` mode that surfaces `BuilderSolid`'s shell summary and its sliver-drop debug lines, and uses it to **refute a hypothesis I formed from reading the assembler**. Diagnostic only.

## The hypothesis

`assemble()` in `builder_solid.rs` handles a non-outer growth shell three ways: join it if closed, **abort** if open with ≥4 faces, **silently drop** it otherwise. That looked like it unified both failure modes — even tools losing small open shells (→ free edges), odd tools tripping the ≥4 abort.

## The measurement

```
tool0:  [algo] BuilderSolid: 1 growth shells, 0 hole shells   → F=494 free=30
tool1:  [algo] BuilderSolid: 20 growth shells, 17 hole shells → ERR (open 9-face shell)
```

tool0 has **one** growth shell, **zero** hole shells, and produces **no** `dropping open N-face growth sliver` message. Nothing is dropped. That single shell is the outer shell — which `assemble()` deliberately never tests for closedness — and it is itself open.

**So tool0's four faces were never created.** The assembler is not discarding them.

tool1 is a genuinely different situation: 20 growth shells, 17 hole shells, fragmentation plus the ≥4 guard.

## Consequence

The two modes are **not** one mechanism, and the unification would have sent the fix at the assembly guard — which is a safety net working as designed, not the bug. The even mode is a splitter/section gap; the odd mode is fragmentation.

This is consistent with the op-independence result in #1218 (Cut/Fuse/Intersect all fail in the same region), which also pointed upstream of classification.

Target, unchanged in location but now better evidenced: **the FF/section/split stage for the 0.05 mm plane-vs-cylinder sliver**, with the odd mode's fragmentation treated as possibly separate.

`log` added as a **dev**-dependency of `brepkit-io` for the example's logger — no change to the published crate.

---
## Summary by cubic
Adds `SHELL_LOG` to the IO replay example to print `BuilderSolid` shell/sliver lines scoped to `brepkit_algo`. Confirms tool0 drops nothing (faces were never created); tool1 hits the ≥4 open-shell abort; roadmap trimmed to note the deliberate 0.05 mm slab overlap, summarize the tangency row, and remove a duplicated tangency tail.

- **New Features**
  - `SHELL_LOG=1` enables a lightweight logger in `replay_cut_capture` that emits "growth shell" and "growth sliver" lines.
  - Logger is scoped to the `brepkit_algo` target to reduce noise.

- **Dependencies**
  - Add `log` as a dev-dependency of `brepkit-io` for the example logger. No changes to the published crate.

<sup>Written for commit f6146ea7a22d943946dd208f013626efe2caac38. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1219?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Fixed the logger scope — `enabled()` now filters to the `brepkit_algo` target at Debug or finer, and `log()` re-checks it rather than trusting the caller to have honoured `enabled()`.

Verified the probe still shows what it exists to show, rather than assuming the filter was harmless: tool0 prints `1 growth shells, 0 hole shells` and tool1 prints `20 growth shells, 17 hole shells`, both unchanged. Worth checking, since a too-narrow target filter would have silently produced the same "no sliver drops" observation for the wrong reason — which is the exact failure mode that bit me three times earlier in this session.

andymai:
Traced where the 0.05 mm actually comes from, and it changes how this should be read.

`kumikoWrapBuilder.ts`:

```js
/** Overlap of flat slabs past the corner tangent planes (boolean robustness). */
const SLAB_OVERLAP = 0.05;
```

The tool extends each flat slab **0.05 mm past the corner tangent planes, explicitly for boolean robustness** — a deliberate workaround to avoid exact tangential contact. And that overlap is precisely where brepkit fails.

Three things follow:

1. **A tangency workaround manufactures a thin-sliver problem instead.** The consumer did the conventional thing to dodge exact tangency and landed in a different failure.
2. **It explains the uniformity.** All 8 tools fail with the same signature because the offset is a constant, not a geometric accident of one band.
3. **It is real evidence for the link to the tangent-section-circle family** (#1213), which I had flagged as unproven geometric coincidence. The sliver sits against the *corner tangent planes* by construction.

A tool-side mitigation (a different overlap value) may well exist and would be worth knowing, but it would be papering over the kernel defect — and 0.05 mm is a perfectly reasonable thing for a consumer to ask of a B-Rep kernel.

Recorded in the roadmap alongside the rest of the chain.

andymai:
Housekeeping that was overdue: the goma roadmap entry had grown to **1561 words** across thirteen appended findings — I wrote every one of them, and it had become the thing the roadmap explicitly warns against ("keep every entry to ONE line with a pointer... never duplicate the detailed truth here").

Compressed to **235 words**: the summary, the target, the refuted list, the tooling gotcha, and pointers. The full measured chain already lives in the doc comment of `crates/io/tests/goma_wall_band_cut_inmem.rs`, which is the right home for it — it sits next to the failing assertion.

What I deliberately kept, because it is what saves the next session time rather than describing what I did:

- **the target** — FF/section/split, faces never created;
- **the refuted list** — panic, bisect thrash, prism construction, batching, classification, the assembly guard. Six dead ends, each of which cost measurable time here;
- **the tooling gotcha** — `--cpu-prof` does not work through this harness, plus the "verify your wrapper covers the path" lesson;
- **the repro cost** — ~230 ms per tool, versus the 850 s scenario it came from.

Net effect on the file: 722 lines, down from where the appends had pushed it.

PR #1220, opened 2026-07-25
test(io): show cylinders stop 0.05mm short of the cut plane

andymai:
Follow-on to #1219. Adds `FACES_NEAR_X=<v>` to list result faces spanning a slab, which produces the sharpest datum yet on the goma root. Diagnostic only.

## The gap, stated exactly

```
face Id(5676) cylinder x[17.000, 20.750]     ← every corner cylinder stops at 17.000
face Id(5678) cylinder x[17.000, 20.750]
face Id(5688) plane    x[17.050, 17.050]     ← tool0's cut plane
face Id(5687) plane    x[16.756, 17.050]     ← planes do reach it
```

Every corner **cylinder** face in the result is trimmed at **x = 17.000**, while tool0's cut plane sits at **x = 17.050**. The 0.05 mm of cylinder between those values has **no face at all** — and that band is precisely what the 30 free edges bound.

The planes reach 17.050 correctly. The cylinders stop 0.05 mm short.

So the four missing patches are **cylinder slivers that should run from x=17.000 out to the cut plane**, and the free-edge loops are the holes they leave behind. That matches everything upstream: the loops are non-planar (#1218), the defect is op-independent (#1218), and nothing is dropped at assembly (#1219) — the splitter simply never extends those cylinder faces to the cut plane.

## Where this leaves the fix

The question is now specific enough to act on: **why the cylinder faces are trimmed at 17.000 rather than at tool0's cut plane 0.05 mm further out.** Candidates worth checking first are the section emitted for the cylinder × cut-plane pair, and whether the cylinder's own trim boundary is being taken from the wrong partner.

Repro is unchanged and cheap: `goma_wall_band_cut_inmem.rs` (~230 ms), or `TOOL=0 FACES_NEAR_X=17.025` through `replay_cut_capture`.

---
## Summary by cubic
Add `BASE_FACES_NEAR_X=<v>` and `FACES_NEAR_X=<v>` to `replay_cut_capture` to show the cut plane at x=17.050 does not split the corner cylinders, which stop at x=17.000. Confirms a 0.05 mm band with no cylinder face and narrows the defect to the FF/section/split stage; investigation notes updated.

- **New Features**
  - `BASE_FACES_NEAR_X=<v>` scans input operands to confirm 17.000 is the base cylinder tangent (not a trim).
  - `FACES_NEAR_X=<v>` slab scan prints face type and edge count to show cylinders span [17.000, 20.750] while the cut plane is at 17.050.

- **Refactors**
  - Renamed a loop vertex variable to avoid shadowing; no behavior change.

<sup>Written for commit 0020b3146a16cf3ad96ba6db12f52f5c7a817099. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -653,8 +653,12 @@ bands, slow only because the analytic path is rejected for **30 free edges** and
 runs; the analytic path is ~12x faster and keeps all 12 cones + 24 cylinders. Those 30 edges are
 **4 missing faces** in one **0.05mm plane-vs-cylinder sliver** — the tool's deliberate
 `SLAB_OVERLAP = 0.05` past the corner tangent planes (a tangency workaround). Repro is ~230ms per
-tool; all 8 bands fail, evens with free=30, odds aborting on an open growth shell. TARGET: the
-FF/section/split stage — the faces are never created.
+tool; all 8 bands fail, evens with free=30, odds aborting on an open growth shell. TARGET: the FF/section/split stage — the faces are never created. SHARPEST DATUM
+(`FACES_NEAR_X=17.025`): in the result every corner CYLINDER face is trimmed at x=17.000 while
+tool0's cut plane is at x=17.050 — the 0.05mm of cylinder between them has NO face at all, and
+that band is exactly what the 30 free edges bound. Planes do reach 17.050; the cylinders stop
+0.05mm short. So the missing patches are cylinder slivers that should run from 17.000 out to the
+cut plane.
 REFUTED, do not re-attempt: a panic (none; `lastPanicMessage()` is empty), bisect thrash
 (telemetry: 1 attempt, 1 success), prism construction (322ms, 0.1%), boolean batching as the main
 cost (30%), classification (defect is op-independent), and the assembly sliver-drop guard (tool0
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -657,8 +657,15 @@ tool; all 8 bands fail, evens with free=30, odds aborting on an open growth shel
 (`FACES_NEAR_X=17.025`): in the result every corner CYLINDER face is trimmed at x=17.000 while
 tool0's cut plane is at x=17.050 — the 0.05mm of cylinder between them has NO face at all, and
 that band is exactly what the 30 free edges bound. Planes do reach 17.050; the cylinders stop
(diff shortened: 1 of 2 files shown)


andymai:
Added `BASE_FACES_NEAR_X` to scan the **inputs**, which settles whether 17.000 is a trim or pre-existing geometry.

```
base Id(5)  cylinder x[17.000, 20.750]
base Id(4)  plane    x[-17.000, 17.000]
tool0       X-normal plane at 17.049999999999997
```

**17.000 is the base's own geometry** — the tangent point where each corner cylinder meets the flat wall. Not a trim.

So tool0's cut plane at **x = 17.050 lies 0.05 mm inside the cylinder's range** (17.000 → 20.750), exactly `SLAB_OVERLAP` past the tangent, and it should split those cylinders.

The result's cylinders still span **[17.000, 20.750]** — identical to the base. Had they been split with the outer piece kept, they would read `[17.050, 20.750]`.

**So the cut plane does not appear to split the corner cylinders at all.** That is the defect, and the four missing patches are the cylinder pieces it should have produced.

This is where black-box probing is exhausted. Everything visible from outside the kernel now agrees — non-planar loops, op-independence, nothing dropped at assembly, cylinders unsplit — and the next question is one level down: **is a section emitted for the cylinder × cut-plane pair at all?** That needs algo-level instrumentation rather than another replay flag, so I am stopping the probe series here rather than adding modes that cannot see it.

Ten narrowings, all measured, from a 14-minute scenario to one unsplit cylinder face, with a 230 ms reproducer.

andymai:
Fixed — the inner vertex loop is now `vid`, so it no longer shadows the `v` bound from `env::var`, and its `expect` message says "vertex" rather than "v".

Re-ran both scans to confirm the rename did not disturb what they report: `FACES_NEAR_X` still lists 65 result faces in the slab, and `BASE_FACES_NEAR_X` still lists the same base and tool0 faces. Worth the extra 30 seconds — a rename inside a filter predicate is exactly the kind of edit that can silently change a bound without failing to compile.

PR #1221, opened 2026-07-25
test(algo): add an env-gated FF pair trace

andymai:
Continues the goma dig past the point where black-box probing ran out (#1220). Adds `BK_FF_TRACE=<x>`, an env-gated diagnostic in `phase_ff::perform`, and uses it to **rule out phase FF as the gap**.

## What it showed

At the sliver's x=17.05, on the `TOOL=0` repro:

```
 64 FF_TRACE pair          (all cylinder × plane, raw_curves=1 each)
736 FF_TRACE reject-aabb   (416 cylinder × plane)
```

**Sections are emitted.** 64 cylinder × plane pairs survive the AABB test and each returns a curve. The rejections are mostly correct — most face pairs genuinely do not meet.

So phase FF is not where the four missing cylinder patches are lost. The defect is **downstream of section computation**: restrict/clip, pave-block splitting, or face building. That is the next target, traceable with the same ~230 ms repro.

I want to flag a wrong turn honestly: my first read of this output was "zero pair lines, everything rejected — found it." That came from a frequency-sorted top-10 that happened to contain only rejects. The `pair` lines were there, just less frequent. The corrected counts above are what the trace actually says.

## On the diagnostic itself

The first version resolved `std::env::var` **inside the face-pair loop** — 800 iterations on this small case and far more on real geometry, so an env lookup plus allocation per pair in the boolean hot path. That is a perf regression introduced by a diagnostic, which would have been a poor trade.

Now resolved once per process behind a module-level `OnceLock` (`ff_trace_x()`). Trace output is unchanged, and the untraced path measures 229–260 ms as before.

Verified: `cargo clippy --all-targets` clean, 425 tests green across `brepkit-algo` and `brepkit-io`.

---
## Summary by cubic
Adds an env-gated FF pair trace to `phase_ff::perform` and extends the replay tool to inspect faces near a target X, helping rule out phase FF as the source of the missing cylinder patches.

- **New Features**
  - `BK_FF_TRACE=<x>`: logs each face pair that straddles x and AABB rejections.
  - `FACES_NEAR_X` and `BASE_FACES_NEAR_X` in `replay_cut_capture`: print face x-spans for result and input solids.
  - Logger prints `FF_TRACE` lines to stdout for easy scanning.

- **Refactors**
  - Resolve `BK_FF_TRACE` once via `OnceLock` instead of per face pair to avoid hot-path overhead.

<sup>Written for commit c5cebd92592699309dc79c95dd4ec12bdcb365e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1221?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -653,8 +653,12 @@ bands, slow only because the analytic path is rejected for **30 free edges** and
 runs; the analytic path is ~12x faster and keeps all 12 cones + 24 cylinders. Those 30 edges are
 **4 missing faces** in one **0.05mm plane-vs-cylinder sliver** — the tool's deliberate
 `SLAB_OVERLAP = 0.05` past the corner tangent planes (a tangency workaround). Repro is ~230ms per
-tool; all 8 bands fail, evens with free=30, odds aborting on an open growth shell. TARGET: the
-FF/section/split stage — the faces are never created.
+tool; all 8 bands fail, evens with free=30, odds aborting on an open growth shell. TARGET: the FF/section/split stage — the faces are never created. SHARPEST DATUM
+(`FACES_NEAR_X=17.025`): in the result every corner CYLINDER face is trimmed at x=17.000 while
+tool0's cut plane is at x=17.050 — the 0.05mm of cylinder between them has NO face at all, and
+that band is exactly what the 30 free edges bound. Planes do reach 17.050; the cylinders stop
+0.05mm short. So the missing patches are cylinder slivers that should run from 17.000 out to the
+cut plane.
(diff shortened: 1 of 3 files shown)


andymai:
Superseded by a clean branch off main — this one picked up #1220's commits and went CONFLICTING, and it also carried the misattached-attribute bug Copilot caught. Reopened as a single-commit PR with both fixed.

PR #1222, opened 2026-07-25
test(algo): add an env-gated FF pair trace

andymai:
Replaces #1221, which went CONFLICTING after picking up #1220's commits. Single commit off current main, with both review findings from there fixed.

Continues the goma dig past where black-box probing ran out (#1220). Adds `BK_FF_TRACE=<x>`, an env-gated diagnostic in `phase_ff`, and uses it to **rule out phase FF as the gap**.

## What it showed

At the sliver's x=17.05 on the `TOOL=0` repro:

```
 64 FF_TRACE pair          (all cylinder × plane, raw_curves=1 each)
736 FF_TRACE reject-aabb   (416 cylinder × plane)
```

**Sections are emitted.** 64 cylinder × plane pairs survive the AABB test and each returns a curve; the rejections are mostly correct, since most face pairs genuinely do not meet.

So phase FF is not where the four missing cylinder patches are lost. The defect is **downstream of section computation** — restrict/clip, pave-block splitting, or face building — traceable with the same ~230 ms repro.

## Fixed from #1221

- **Misattached attribute** (Copilot). My splice landed between `perform`'s doc block and its `#[allow(clippy::too_many_lines)]`, so both attached to the new helper and `perform` silently lost its allow. The helper now sits above the doc block; verified `#[allow(clippy::too_many_lines)]` is immediately followed by `pub fn perform`.
- **Formatting churn** (Copilot) — `cargo fmt` applied.

## Diagnostic hygiene

The first version resolved `std::env::var` **inside the face-pair loop** — 800 iterations on this small case, far more on real geometry, so an env lookup and allocation per pair in the boolean hot path. Now resolved once per process behind a module-level `OnceLock`. Trace output unchanged; untraced path measures 229–260 ms as before.

Verified: `clippy --all-targets` clean, 425 tests green across `brepkit-algo` and `brepkit-io`.

One correction carried over: my first read of this trace was "zero pair lines, everything rejected." That came from a frequency-sorted top-10 that happened to contain only rejects. The counts above are what the trace actually says.

---
## Summary by cubic
Adds an env-gated FF pair trace in `phase_ff` via `BK_FF_TRACE=<x>` to log face-pair behavior at a chosen x. Roadmap updated: sections survive FF/restrict/emission; only two cylinder faces report sections, and the x-only filter is non-localizing — next step is to check y/z-local faces.

- **New Features**
  - `BK_FF_TRACE=<x>` logs: "reject-aabb", `raw_curves` count, "restrict N -> M", and "emit curve#k type".
  - Resolves the env var once via a module `OnceLock`; untraced path performance unchanged.
  - `replay_cut_capture` prints `FF_TRACE` lines for easier inspection.

- **Bug Fixes**
  - Corrected placement of `#[allow(clippy::too_many_lines)]` on `perform`.
  - Formatting cleanup.

<sup>Written for commit 7573772b2b9c44c401e39815e0e6edfbb3918c75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1222?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -663,9 +663,11 @@ tangent point where each corner cylinder (x[17.000,20.750]) meets the flat wall 
 range (exactly SLAB_OVERLAP past the tangent), and should split it. The result's cylinders still
 span [17.000,20.750], identical to the base: had they been split with the outer piece kept they
 would read [17.050,20.750]. **So the cut plane appears not to split the corner cylinders at all** —
-that is the defect, and the missing patches are the cylinder pieces it should have produced. NEXT
-STEP needs algo-level instrumentation (black-box probing is exhausted): whether a section is
-emitted for the cylinder x cut-plane pair at all.
+that is the defect, and the missing patches are the cylinder pieces it should have produced. ALGO TRACE (`BK_FF_TRACE=<x>` in phase_ff.rs, env-gated): sections ARE
+emitted and DO survive clipping — at x=17.05, 64 cylinder x plane pairs pass the AABB test (736
+rejected, mostly correctly) and **12 sections survive `restrict_curves_to_faces` intact**. So
+phase FF and restrict both do their job; curves reach the splitter. NEXT STEP: pave-block splitting
+or face building, same 230ms repro.
 REFUTED, do not re-attempt: a panic (none; `lastPanicMessage()` is empty), bisect thrash
 (telemetry: 1 attempt, 1 success), prism construction (322ms, 0.1%), boolean batching as the main
 cost (30%), classification (defect is op-independent), and the assembly sliver-drop guard (tool0
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -673,8 +673,14 @@ BETWEEN ARENA EMISSION AND THE PER-FACE SECTION LISTS (`BK_SPLIT_TRACE=1`, eprin
 `fill_images_faces` face loop): that loop DOES run, over 741 faces, and reports **only 2 of 24
 CYLINDER faces with has_sections=true** (22 without; planes are 420 true / 285 false) — although FF
 emitted 12 cylinder x plane curves. So ~10 emitted sections never reach their cylinder face's
-section list, which is why those faces are never split. NEXT STEP: `build_section_edges` / how
-arena curves are attached to faces. WARNING: `log::debug!` markers added inside
+section list, which is why those faces are never split. NEXT STEP: `build_section_map` in fill_images_faces.rs, which SKIPS any
+`arena.curves` entry whose `pave_blocks` is empty (`continue`). HYPOTHESIS, EXPLICITLY UNVERIFIED:
+those 10 cylinder sections carry no pave blocks and are skipped there — which would make this the
+SAME root as the snapClip deepened-notch row ("marched FF sections on curved faces carry
+pave_block_id=None, bypassing the pave machinery", canonical fix at phase-FF/make_blocks altitude).
+Could NOT be confirmed: `log::debug!` inside fill_images_faces.rs does not emit (see warning below),
+and `eprintln!` is denied there by `clippy::print_stderr`, so both probes read as a false zero.
+Confirm it by counting empty-`pave_blocks` curves some other way before acting on it. WARNING: `log::debug!` markers added inside
 `fill_images_faces` did NOT reach a custom logger that WAS receiving `builder_solid`'s debug lines,
 so log-based probes there read as a false ZERO — use `eprintln!` gated on the env var instead.
 REFUTED, do not re-attempt: a panic (none; `lastPanicMessage()` is empty), bisect thrash
--- a/.claude/skills/roadmap/SKILL.md
+++ b/.claude/skills/roadmap/SKILL.md
@@ -674,15 +674,15 @@ BETWEEN ARENA EMISSION AND THE PER-FACE SECTION LISTS (`BK_SPLIT_TRACE=1`, eprin
 CYLINDER faces with has_sections=true** (22 without; planes are 420 true / 285 false) — although FF
 emitted 12 cylinder x plane curves. So ~10 emitted sections never reach their cylinder face's
 section list, which is why those faces are never split. NEXT STEP: `build_section_map` in fill_images_faces.rs, which SKIPS any
-`arena.curves` entry whose `pave_blocks` is empty (`continue`). HYPOTHESIS, EXPLICITLY UNVERIFIED:
-those 10 cylinder sections carry no pave blocks and are skipped there — which would make this the
-SAME root as the snapClip deepened-notch row ("marched FF sections on curved faces carry
-pave_block_id=None, bypassing the pave machinery", canonical fix at phase-FF/make_blocks altitude).
-Could NOT be confirmed: `log::debug!` inside fill_images_faces.rs does not emit (see warning below),
-and `eprintln!` is denied there by `clippy::print_stderr`, so both probes read as a false zero.
-Confirm it by counting empty-`pave_blocks` curves some other way before acting on it. WARNING: `log::debug!` markers added inside
-`fill_images_faces` did NOT reach a custom logger that WAS receiving `builder_solid`'s debug lines,
-so log-based probes there read as a false ZERO — use `eprintln!` gated on the env var instead.
+`arena.curves` entry whose `pave_blocks` is empty (`continue`). REFUTED: that skip never fires — all
+**890 curves carry exactly ONE pave block, none are empty**, so this is NOT the snapClip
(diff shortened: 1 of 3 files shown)


andymai:
Extended the trace one stage further, to `restrict_curves_to_faces`:

```
52  restrict cylinder×plane  0 -> 0    (no curve by this point)
12  restrict cylinder×plane  1 -> 1    (curve computed AND survives)
```

**12 sections survive restrict intact.** So phase FF emits them and restrict keeps them — curves reach the splitter. Neither stage is the gap, which eliminates the two most obvious suspects given the geometry.

The next stage down is pave-block splitting / face building, traceable with the same ~230 ms repro.

One correction to my own earlier reading in this PR: I had sampled the pair output with `sort -u`, saw `raw_curves=1`, and described all 64 pairs that way. The distribution is actually 12 with a curve and 52 without by restrict time — the 52 are dropped by the intermediate plane/quadric trim. Unique-ing collapsed the distribution and I read the sample as the population. Same shape as the frequency-sorted top-10 mistake earlier: the numbers were real, the aggregation hid what mattered.

andymai:
Traced one more stage — section emission into the arena:

```
12 FF_TRACE emit a=cylinder b=plane curve#N line
```

All 12 surviving sections are emitted, as **line** curves, and that type is geometrically correct: the cut plane's normal is along X while the corner cylinders' axes are along Z, so the plane is *parallel to the axis* and meets them in generator lines rather than ellipses. One line per pair is right as well — only one of the two generators lies on each quarter-arc face.

So the pipeline through FF is healthy end to end:

| stage | result |
|---|---|
| AABB pairing | 64 cylinder × plane pairs survive |
| section computation | curves produced |
| `restrict_curves_to_faces` | 12 survive intact |
| arena emission | 12 emitted, correct `line` type |

**The splitter receives correct sections and still does not split the cylinder faces.** That is a much stronger statement than "something upstream is missing" — it localises the defect to pave-block splitting or face building, with everything they consume verified good.

Next stage down, same ~230 ms repro. 425 tests green.

andymai:
Traced through to the per-face section lists, and found the gap — plus a near-miss I want on the record.

```
420 plane    has_sections=true
285 plane    has_sections=false
 22 cylinder has_sections=false
  2 cylinder has_sections=true
 12 cone     has_sections=false
```

FF emits **12** cylinder × plane section curves, but only **2 of 24 cylinder faces** have sections attached when `fill_images_faces` reaches them. **Roughly ten emitted sections never arrive at their cylinder face's section list**, which is why those faces are never split and the four patches are never created.

So the gap is between arena emission and `build_section_edges` / per-face attachment. That is the next place to look.

## The near-miss

My first attempt used `log::debug!` markers inside `fill_images_faces`. They produced **zero** output — while `builder_solid`'s `log::debug!` lines were reaching the same custom logger in the same run. I was one step from recording "the face-split loop is never entered."

It runs 741 times. An `eprintln!` gated on the env var proved it immediately. I have not chased why that module's `debug!` records do not reach the logger, but the roadmap now warns that log-based probes there read as a false zero.

That is the sixth instrument-didn't-fire incident today, and the first that would have put an actively false claim into the roadmap rather than merely a premature one.

## Not committed

The `eprintln!` instrumentation is reverted — `clippy::print_stderr` is denied in kernel crates, and silencing that lint to keep a diagnostic is a bad trade. The roadmap records the numbers and how to re-derive them. The `BK_FF_TRACE` addition in `phase_ff` stays, since it uses `log` and demonstrably works (876 records).

425 tests green.

andymai:
Traced to the most likely mechanism, but **could not verify it**, so it is recorded as a hypothesis rather than a finding.

`build_section_map` in `fill_images_faces.rs` skips any `arena.curves` entry whose `pave_blocks` is empty:

```rust
for (idx, curve) in arena.curves.iter().enumerate() {
    if curve.pave_blocks.is_empty() {
        continue;      // ← curve never reaches any face's section list
    }
```

That would explain the ~10 lost cylinder sections exactly. It would also make this **the same root as the snapClip deepened-notch row** already in the roadmap — *"marched FF sections on curved faces carry `pave_block_id=None`, bypassing the pave machinery"*, whose canonical fix is named at phase-FF/`make_blocks` altitude. Two open families collapsing into one would be a significant result.

**Why I am not claiming it.** Both available probes read as a false zero in that file:

- `log::debug!` there does not emit — proven, since the face loop logs nothing yet demonstrably runs 741 times, and removing my logger's target filter changed nothing;
- `eprintln!` is denied by `clippy::print_stderr` in kernel crates.

So "no output" is uninformative: I cannot distinguish "no curve has empty `pave_blocks`" from "the probe never fired." Given that this session has produced six instrument-didn't-fire incidents, treating this particular zero as evidence would be indefensible.

Recorded in the roadmap with the hypothesis, the reason it is unconfirmed, and the instruction to count empty-`pave_blocks` curves by some other route before acting on it.

Kernel diagnostics reverted; only the working `BK_FF_TRACE` in `phase_ff` remains.

andymai:
Broke through the tooling wall, and the answer **refutes my own hypothesis**.

First, the wall itself is now proven rather than suspected. An adjacent `log::debug!` and `eprintln!` on the *same line* in `fill_images_faces.rs`:

```
  0 PROBE_LOG   (log::debug!)
890 PROBE_ERR   (eprintln!)
```

So logging in that file genuinely does not emit, while `builder_solid`'s `log::debug!` reaches the same logger fine. Cause not diagnosed, but every log-based probe there is a false zero.

With a probe that actually fires:

```
890 curves, all pb=1     (zero curves with empty pave_blocks)
```

**Every curve carries exactly one pave block.** The `if curve.pave_blocks.is_empty() { continue }` never fires, so no section is lost there — and this is **not** the snapClip deepened-notch pave-bypass root. Had I claimed that collapse of two families on the strength of an unfired probe, it would have been wrong, and it would have pointed the fix at `make_blocks` for no reason.

The loss is instead in the `section_map` → face lookup: curves are keyed by `curve.face_a`/`face_b`, yet 22 of 24 cylinder faces find nothing. A face-id mismatch between the arena's recorded faces and the faces the split loop iterates is the natural next suspect — stated as a suspect, not a finding.

Roadmap updated with the refutation and the proven logging warning. Kernel probes reverted; only `BK_FF_TRACE` remains.

andymai:
Tested the face-id-mismatch suspect. **Also not it** — and the result sends the search back upstream.

```
PROBE map keys total=422  cylinder=2  plane=420
22 loop cylinder  in_map=false  has_sections=false
 2 loop cylinder  in_map=true   has_sections=true
```

The `section_map` holds only **2 cylinder faces**. The other 22 are simply absent (`in_map=false`), so nothing is lost in the lookup and there is no id mismatch. Only 2 of 24 cylinder faces ever had a section emitted for them at all.

That forces a re-reading of the restrict numbers I reported earlier. The `52 restrict 0 -> 0` pairs had **zero curves before restrict as well** — I had described them as "no curve by this point", which was accurate but I did not follow through on what it implied. So:

**`compute_raw_curves` returns nothing for 52 of the 64 cylinder × plane pairs.** That is where the 22 cylinders lose their sections.

The 12 pairs that do produce a curve prove the code path works for this configuration — these are generator-line cases (plane normal ⟂ cylinder axis). So the question is what distinguishes the 52 from the 12.

Three suspects eliminated in this PR, each by measurement: phase FF pairing, `restrict_curves_to_faces`, and the section-map lookup. Two hypotheses of mine refuted outright: the pave-blocks skip and the face-id mismatch.

Roadmap updated with the new target.

andymai:
Correcting my own framing, and stopping the dig here.

I have been treating "only 2 of 24 cylinder faces get sections" as the anomaly. That reading does not hold up. The `FACES_NEAR_X` filter I built tests **only the X range**, and every corner cylinder spans `[17.000, 20.750]` — so all four corners pass it regardless of where they sit in y/z. It never localised to tool0's wall.

tool0's slab covers **one wall**. Most of those 22 cylinders are on other corners and *should* be untouched. So 2 sectioned may be exactly right, and the last few narrowings were chasing a number that was never anomalous.

**Where to start next**, recorded in the roadmap: identify which cylinder faces tool0 actually reaches, using y and z rather than x alone, and only then decide whether 2 is right or short.

What remains measured and solid from this PR is still substantial — FF pairing, `restrict_curves_to_faces`, section emission, and the section-map lookup are all confirmed working for this case, and two hypotheses of mine (the empty-`pave_blocks` skip, and a face-id mismatch) are refuted outright.

I am stopping rather than pushing further because my error rate over the last several steps has been climbing — three consecutive readings needed correcting, this one included. The measured results are worth more than another speculative narrowing on top of them, and the handoff is better served by an honest boundary than by a deeper but shakier claim.

PR #1415, opened 2026-08-07
build(release): retry rate limits, add a semver check and a stability policy

andymai:
Follow-ups from the crates.io bootstrap, which hit crates.io's new-crate rate limit (a burst of 5, then one per 10 minutes) partway through publishing and needed an out-of-repo wrapper script to finish.

## Retry on 429

`publish-crates.sh` now handles the rate limit itself. The retry is bounded (`PUBLISH_MAX_RETRIES`, default 3): a run that adds more crate names than the cap should be finished by re-running the script rather than holding a CI runner idle for an hour, and the skip-if-published check makes resuming free. A non-429 failure still aborts immediately with no retries.

Verified against a stubbed `cargo` for all three paths:

| Path | Result |
|---|---|
| Two 429s then success | exits 0, publishes everything |
| Retries exhausted | exits 1 with a resume message |
| Non-429 error | exits immediately, zero retries |

One subtlety worth noting for future edits: the `out=$(cargo publish ...) \|\| status=$?` form is load bearing. A bare assignment from a failing command substitution trips `set -e` before the status can be inspected, which silently defeats the retry.

## SemVer check (advisory)

`cargo semver-checks` runs on every PR but is **deliberately absent from `CI Pass`**.

The crates publish at 2.x while the kernel's internals still move, and adding a `FaceSurface` or `EdgeCurve` variant is both semver-major and on the roadmap (surface-of-revolution and surface-of-linear-extrusion are needed for STEP files that currently fail to import). A blocking gate would force a major bump across all crates on routine work, before there is a deprecation process to justify it. The check makes breakage visible and deliberate; the policy for acting on it is in STABILITY.md.

## STABILITY.md

Says what the shared 2.x version does and does not promise. The number came from the npm release line, not from the Rust APIs having settled, and nothing said so until now.

Written specifically to resist going stale:

- Stability is stated **per layer**, deferring to the boundary table that `scripts/check-boundaries.sh` already enforces in CI, rather than as a hand-maintained per-crate list that nobody would update.
- Variant counts and delegate-method names are **referenced, not copied**. An earlier draft hardcoded "six variants and four", which CLAUDE.md already tells you to derive with `rg`.
- Feature maturity defers to the README status table rather than repeating it.
- No version numbers; `cargo add` resolves them.

For the same reason this drops the README's per-crate dependency table, which duplicated the Architecture table a few sections above it.

## Verification

Scripts shellcheck clean. Boundaries, doc-paths, version check, and taplo all green. Both new CLAUDE.md anchors resolve.

---
## Summary by cubic
Hardens the release process: retries crates.io 429 rate limits during publish, adds an advisory `cargo semver-checks` job in CI, and documents our API stability policy. This reduces failed publishes and clarifies what `2.x` guarantees.

- New Features
  - Release: `scripts/publish-crates.sh` now retries 429s with bounded attempts (`PUBLISH_MAX_RETRIES`, `PUBLISH_RETRY_WAIT`), resumes safely on re-run, and still fails fast on non-429 errors.
  - CI: Adds an advisory `SemVer Check` job using `cargo semver-checks` against the latest crates.io baselines; it runs on PRs but is not part of `CI Pass`.
  - Docs: Adds `STABILITY.md` describing per-layer stability, the shared `2.x` policy, and the upcoming `FaceSurface`/`EdgeCurve` variant addition risk; updates README to link it and removes a duplicate per-crate table; notes policy in `CLAUDE.md`.

<sup>Written for commit ed7f0a70ad8117d4186b6b0af839a2037398d764. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1415?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,6 +259,27 @@ jobs:
       # tarball. Publishes nothing.
       - run: cargo publish --workspace --dry-run
 
+  # Advisory, not a gate: deliberately absent from `ci-pass`. brepkit's crates
+  # are published at 2.x but the kernel's internals still move, and adding a
+  # `FaceSurface` or `EdgeCurve` variant is a semver-major change that is also
+  # on the roadmap. Making this blocking would force a major bump across all
+  # 13 crates on a routine change. It runs to make breakage visible; the
+  # policy for acting on it lives in STABILITY.md.
+  semver:
+    name: SemVer Check (advisory)
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+      - uses: taiki-e/install-action@41049aa56687c35e0afa74eed4f09cec4f9afabf # v2.62.30
+        with:
+          tool: cargo-semver-checks
+      # Baselines against the newest version of each crate on crates.io.
+      - run: cargo semver-checks --workspace
+
   machete:
     name: Unused Dependencies
     runs-on: ubuntu-latest
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -605,6 +605,9 @@ workspace-inherited relative readme path does not resolve), an entry in
 `[workspace.dependencies]` **with a version**, a matching jsonpath in
 `release-please-config.json`, and its name in `scripts/publish-crates.sh`.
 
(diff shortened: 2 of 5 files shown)


andymai:
Re the grep-vs-bash-substring finding: acknowledged but leaving as is for this PR. The script's only supported environments are the CI ubuntu runner and developer machines that already satisfy the cargo/jq/curl preflight, and grep is present in both. Swapping to substring matching is a fine cleanup for a future pass over the script; it does not change behavior on any environment the script actually runs in.

Issue #1517, opened 2026-08-10
parity: brepkit leads on perf but the tool's generator suite has 152 failures, dominated by non-watertight exports


andymai:
Full head-to-head run of the gridfinity tool's generator suite under brepkit 3.2.18 vs occt-wasm 4.2.0, same session and machine, kernel resolution verified by path. This is the first end-to-end parity measurement I can find; previous green numbers for that suite were all taken on the default (reference) kernel.

## Verdict

**Performance: brepkit leads.** **Geometry: not yet at parity.**

| | |
| --- | --- |
| Generator suite under brepkit | **137 failed / 2550 passed / 7 skipped** (272 files) — updated after brepjs#2012 landed; originally 152/2535 |
| Same suite under occt-wasm | green |
| Perf, 26-scenario matrix | brepkit **0.69x aggregate**, faster on 24 of 26 |
| Structural cleanliness | brepkit **0 non-manifold scenarios**, occt-wasm 5 |

An important caveat on reading the first number: an earlier count of 261 was mostly **stale brepkit snapshot baselines**, not defects. The tool routes each kernel to its own `.brepkit.snap`, and those baselines predate the 3.2.9 density campaign that cut triangle counts. Refreshing them (only `.brepkit.snap`; reference baselines untouched) removed ~110 false failures. The 152 above is after that refresh.

## Root-cause taxonomy

130 distinct error lines. Categories overlap slightly because assertion text matches more than one bucket.

| Root | Count | Status |
| --- | --- | --- |
| `cannot mesh shape of type 'compound'` | 17 | **FIXED** in brepjs#2012 |
| Mesh not closed (`boundary edges != 0`) | 45 | open, dominant geometric gap |
| Timeouts (30s / 180s) | 13 | open, perf tail |
| Non-manifold edges | 11 | open |
| `recursive use of an object … unsafe aliasing in rust` | 4 | open, wasm reentrancy |
| `fuse requires a solid, got compound` | 2 | open, same adapter class as the mesh gap |
| `unreachable` (panic) | 1 | open |

### The dominant gap: export meshes are not watertight

The `boundary edges != 0` failures concentrate in the **export** path and span many unrelated features, which suggests a shared root rather than per-feature bugs:

```
8  binGenerator.export.permutationMatrix
4  binGenerator.export.kumikoWrapping
4  binGenerator.export.kumikoPatterns
4  baseplateGenerator.scenario.snapClip
3  binGenerator.export.wallThickness
[...]

PR #1521, opened 2026-08-10
test(io): pin the #1517 lid magnet-post fuse, the root of the crash and the timeouts

andymai:
Diagnoses #1517 root (b). No kernel change here: this pins the repro and records the chain, because the fix is a face-splitter change and the roadmap says not to guess at one without a foil set.

## The crash and the 14 timeouts are one root

I traced every kernel call in the `lidRetentionMagnets` scenario and serialised each boolean's operands. The result is not a loop with a missing exit:

```
call ~180  fuse(lid, post)  A: F=44  [(cone,4),(cylinder,16),(plane,24)]
                            B: F=3   [(cylinder,1),(plane,2)]
           raw GFA          F=47 [(cone,4),(cylinder,17),(plane,26)] free=6
           -> gate rejects (euler=-1, 13 boundary edges) -> mesh fallback
           -> F=305 [(plane,305)]
call ~190+ each further post fuses into the all-planar blob
           -> 8852 faces, all flat, by call 278
call  248  fuse(blob, post) never returns; ~12 min later the kernel aborts
```

One boolean loses its analytic surfaces, and two dozen later booleans pay compound interest on the debris. Whether a given test reports a timeout or the abort just depends on which limit arrives first.

## The geometry

The post is a cylinder of radius 4 centred at (-118.500, 76.500), spanning z=-2.800..-0.700. It sits at a lid corner and pokes out through **both** perpendicular walls, x=-122.000 and y=80.000. The lid has a horizontal ledge at z=-0.800, so the post's cross-section there is a full circle that the walls cut into arcs:

```
circle x=-122.000 at y=78.436 and y=74.564
circle y= 80.000  at x=-116.564 and x=-120.436
```

The six free edges are exactly those arcs plus the post's z=-0.700 rim. The telling part: the ledge plane keeps three arcs and the post cylinder keeps two, and they are **different** arcs of the same circle. So this is not one surface failing to split. The two are split against each other inconsistently.

## Red herring, recorded so the next pass does not chase it

`plane_internal_line_loops` logs three rejections here:

```
endpoint ... not strictly interior (dist 0.000000000000028424365601609406)
endpoint ... not strictly interior (dist 0.0000000000000008881784197001252)
endpoint ... not strictly interior (dist 0)
```

Distances of 2.8e-14, 8.9e-16 and exactly zero against a 1e-5 margin look like a strictness bug. They are not: those sections genuinely touch the face boundary, so declining to treat them as an *internal* loop is correct.

## What this PR adds

- `crates/io/tests/lidpost_fuse_inmem.rs` with the captured operands. **5 ms**, against ~12 minutes in the tool.
  - `lidpost_operands_are_well_formed` (active) guards both inputs so a future failure cannot be blamed on a poisoned capture
  - `lidpost_fuse_is_closed_and_analytic` (ignored) is the acceptance bar
- Roadmap: root (b) rewritten with the real mechanism, and the `#[ignore]` inventory updated, since this adds the first deferred-defect pin since it was last declared clean

## Reproducing

```sh
cargo test -p brepkit-io --test lidpost_fuse_inmem
A=crates/io/tests/data/lidpost_fuse_lid.bin B=crates/io/tests/data/lidpost_fuse_post.bin \
  OP=fuse FREE_EDGES=1 RUST_LOG=debug cargo run --release -p brepkit-io --example replay_pair
```

## Method note

Local wasm builds were recorded as blocked (`wasm-pack` failing on a missing `libbz2.so.1.0`). Only the SONAME alias is missing; `/usr/lib64/libbz2.so.1.0.8` is present, so a symlink on `LD_LIBRARY_PATH` plus the pinned `wasm-bindgen-cli` restores the loop. That is what made this tractable: kernel change to tool-side answer in ~10 minutes instead of a release cycle.

---
## Summary by cubic
Pins a fast in-memory repro for the lid magnet‑post corner fuse in #1517, the single root behind the crash and the 14 timeouts. Adds fixtures, an ignored acceptance test, a passing single‑wall control, and hardens the operand guard; no kernel changes.

- **New Features**
  - Added `crates/io/tests/lidpost_fuse_inmem.rs` with an active well‑formedness check that resolves wires (prevents false “clean” counts on broken topology), an ignored acceptance test (`lidpost_fuse_is_closed_and_analytic`), and a control (`lidpost_single_wall_control_fuses_closed_and_analytic`).
  - Added fixtures: `crates/io/tests/data/lidpost_fuse_lid.bin`, `crates/io/tests/data/lidpost_fuse_post.bin`, `crates/io/tests/data/lidpost_fuse_post_control.bin`.
  - Updated `.claude/skills/roadmap/SKILL.md` to mark the deferred‑defect pin and document #1517 root as the lid magnet‑post corner fuse.

<sup>Written for commit 3bc995614e43e68992c0e534c5f0fd180d8f3abf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1521?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Added a control that measures the discriminant rather than assuming it.

The same lid, fused with an earlier post from the same scenario that protrudes through **one** wall instead of two:

```
RAW fuse: F=48 mix=[(cone,4),(cylinder,18),(plane,26)] free=0  vol=44098.604
```

Clean. So posts, protrusion and this lid are all ruled out, and the two-wall corner is the variable that matters. Pinned as `lidpost_single_wall_control_fuses_closed_and_analytic` (active, not ignored), so a future fix that repairs the corner by breaking the ordinary case fails loudly.

One more thing the control surfaced: an even earlier post is fully **inside** the lid, and raw GFA gives `free=6` there too. The operations layer only rescues it because the pre-GFA containment shortcut returns A unchanged. So the same defect is present in the contained case and merely masked, which is worth knowing before anyone concludes the corner geometry is what triggers it in the engine.

PR #1524, opened 2026-08-10
docs: correct the #1517 control claim, it is a two-wall corner post too

andymai:
Corrects a wrong claim I published in #1521 and in the #1517 thread, before anyone builds on it.

## What I got wrong

I described `lidpost_fuse_post_control.bin` as a post "protruding through ONE wall instead of two" and concluded the two-wall corner was the discriminant. `BK_FF_DUMP` says otherwise.

The control is the **mirror-image corner post** at (+122, -80). It crosses two walls exactly as the failing case does, and both produce the same seven sections:

| | failing | control |
| --- | --- | --- |
| ceiling circle at z=-0.800 | 3 (duplicated) | 3 (duplicated) |
| wall lines | 4 | 4 |

So neither the two-wall corner nor the duplicated ceiling sections separates the two cases. Both were plausible roots and both are now ruled out by measurement rather than argument. The test name `lidpost_single_wall_control_fuses_closed_and_analytic` is a misnomer; I left the name alone so the merged reference in #1521 still resolves, and said so in the doc comment.

## What actually differs

Crossing angles measured about each post's own centre, seam at 0 degrees:

```
failing  crossings  61.05  118.95  151.05  208.95   seam 0 is OUTSIDE every crossing sector
control  crossings  28.95  331.05  241.05  298.95   seam 0 is INSIDE the x-wall sector (331.05..28.95)
```

The spur in the failing result anchors exactly at that seam, so "the seam lands in a sector the split has to keep" is the live hypothesis. Recorded as a hypothesis, not a result.

## Also worth knowing

The ceiling-cylinder section is stored **three times** in both cases. That is not this bug, but three identical circles on one face pair is odd enough to be worth a look on its own.

---
## Summary by cubic
Correct the #1517 control-case: the control post is the opposite-corner two-wall case at (+122, -80), not single-wall or a mirror twin. Update docs/tests to note both cases share the same seven sections, seam placement is the working hypothesis, and `POINT_IN` asymmetry means the control is a guard (not a one-variable comparison); no code or test behavior changes.

<sup>Written for commit d8260cfff3863cf7145edbfca56a1fc7385351ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1524?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Second correction, to my own correction. I called the control the "mirror-image" corner post. It is not a mirror twin, and the lid is not symmetric between those corners.

Sweeping across the wall with `POINT_IN` at z=-1.800:

```
y=+76.5   x=-121.5 Outside   x=-122.2 Inside    x=-122.54 Inside   x=-123.5 Inside
y=-76.5   x= 121.5 Outside   x= 122.2 Outside   x= 122.54 Outside  x= 123.5 Inside
```

Material starts at x=-122.000 on one side and only near x=+123 on the other, even though a wall plane at x=+122.000 exists and does cut sections against the post. So the two posts sit at geometrically different places on the lid.

That demotes the control: it is a **guard**, not a one-variable comparison. It still rules out posts, protrusion and this lid in general, and it will still fail loudly if a fix breaks the working case. It does not identify the discriminant, and my seam-placement hypothesis rests on treating the two as comparable, so that is weaker than I stated too.

The unexplained asymmetry is now the most concrete open thread: a wall plane exists at x=+122.000 and cuts sections, but the solid does not begin until roughly x=+123 there. Either the lid has structure I have not accounted for, or the classifier is wrong on that side, and those two have very different consequences for #1517.

PR #1631, opened 2026-08-30
feat(blend): complete corrected fillet cutover

ceeray:
BK9: ship the corrected `FilletBuilder` engine as the production and WASM fillet engine.

- `blend_ops::fillet_v2` becomes the first production/WASM engine; publication is atomic (unchanged, partial, open, non-manifold, inward results rejected)
- `fillet_rolling_ball` removed from the dispatch chain, callable only as reference/diagnostic; flat-bevel `fillet` remains a last-resort fallback
- Junction corners use the terminal end of each cross-section pair; convex trihedral setbacks apply only to near-equal-length stripes
- Four-stripe (mixed-radius) junctions close with an apex-fan of ruled corner faces
- Analytic wire boundaries sampled per edge; NURBS tessellation preserves face handedness
- Single constant-radius planar stripes propagate mid-edge contact splits into untouched caps
- Trimmed torus faces use wire-bounded AABBs instead of the full analytic torus extent
- Gridfinity D2 lip-ring regression now selects the peak rim edge as the tool does

---
## Summary by cubic
Switches production and WASM fillets to the corrected `FilletBuilder` engine (`blend_ops::fillet_v2`), which fixes box-corner over-removal, open junctions at mixed-radius corners, and inverted face tessellation that the rolling-ball engine produced.

**Refactors**
- `fillet_rolling_ball` is removed from the dispatch chain and callable only as a reference/diagnostic implementation; flat-bevel `fillet` remains a last-resort WASM fallback.
- WASM publication is now atomic — unchanged, partial, open, non-manifold, or inward results are rejected.
- Adds `boundary_registry`, `fillet_plan`, `signed_volume`, and `orientation` modules for staged planning, canonical boundary ownership, and shell orientation repair.
- Adds a frozen cross-fillet regression fixture with a shell audit in `crates/io/tests`.

**Bug Fixes**
- Junction corners use the terminal end of each cross-section pair instead of over-removing corner material.
- Four-stripe mixed-radius junctions close with an apex-fan of ruled corner faces.
- Convex trihedral setbacks apply only to near-equal-length incident stripes.
- Mid-edge contact splits on planar stripes propagate into untouched caps.
- NURBS tessellation mirrors triangle winding on left-handed parameterizations so faces keep their material side.
- Trimmed torus faces use wire-bounded AABBs instead of the full analytic extent.
- The Gridfinity D2 lip-ring regression selects the peak rim edge.
- `signed_volume` samples spheres with the `SphericalSurface` UV convention, so off-origin spherical faces no longer flip the shell inward.
- Selected edges absent from the source shell fail planning instead of panicking.
- WASM `filletV2`/`chamferV2` are wrapped in `catch_unwind` so a blend panic can't abort the WASM instance.

<sup>Written for commit b4d86dbc7b6d1e568223a658863bb3e48776e8c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1631?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thanks for this, @ceeray, and welcome! This is a substantial and careful first contribution, and it holds up well under review. I ran a correctness pass and a dedicated security pass over the cutover.

A few things that stood out as genuinely solid:

- The atomic-publication guard in `try_fillet` really does enforce what the description claims. It rejects unchanged (`r.solid != solid_id`), partial (`!is_partial && failed.is_empty() && succeeded.len() == edges.len()`), open/non-manifold (`validate_shell_closed`), and globally-inverted shells (`oriented_solid_volume > 1e-12`, which edge-sense checks alone cannot catch). It also removes the old silent no-op success on the `fillet` binding.
- The `BoundaryRegistry` shared edge/pcurve ownership and `propagate_orientation` are correct and respect the arena's snapshot-then-allocate rule.
- Analytic curve preservation in the trimmer paths is intact.

I pushed three fixes directly to the branch (commit b4d86dbc):

1. **Sphere volume sign** (`signed_volume.rs`): the sphere UV grid had latitude and azimuth in swapped slots relative to `SphericalSurface::evaluate`, so `signed_shell_volume` returned a wrong, translation-variant value for any `Sphere` face. Filleting a solid with an off-origin spherical face then flipped the shell inward (WASM caught it via the guard and errored; the direct `fillet_v2` API returned an inverted solid). Bounds corrected to `(0..2pi, -pi/2..pi/2)`, plus a regression test asserting the flux equals `(4/3)*pi*r^3` and is translation-invariant.

2. **Empty-contour panic** (`g1_chain.rs`): `group_g1_contours` sorted contours by `contour[0]`, which panics on the empty contour produced when a selected edge is not on the source shell (a foreign, cavity, or stale edge). It is reachable from the unguarded `filletV2` binding and aborts the whole WASM instance. It now sorts empties last so the existing `PlanningFailure` path handles them. The regression test mixes a valid and a foreign edge, because a single foreign edge hides the bug (a one-item sort skips the key closure).

3. **Panic-safety parity** (`bindings/operations.rs`): wrapped `filletV2` and `chamferV2` in `catch_unwind`, matching the v1 `fillet` binding, so a blend panic cannot cross the FFI boundary.

I also dropped the manual `CHANGELOG.md` entry, since release-please generates it from the conventional commit on merge, and taplo-sorted the new `crates/io` dev-dependency.

One non-blocking follow-up for a later PR: the recursive `boundary_cycle::search` in `corner.rs` is a backtracking DFS with no work budget, so a high-valence junction could in theory make it slow. A `None`-past-a-cap guard would bound it, and the caller already treats `None` as recoverable.

Local checks pass (clippy `-D warnings` on blend/operations/wasm, and the blend/operations/wasm fillet suites). CI is running now; I will merge once it is green. Thanks again for the excellent work.

andymai:
I went through cubic's re-review (24 findings, confidence 2/5). Triage below.

The two genuinely severe issues were caught in the review pass and are already fixed in b4d86dbc (the sphere volume sign bug and the empty-contour panic); cubic surfaced neither. The rest do not block the constant-radius production path (the `fillet` binding, which routes through the atomic `try_fillet` guard) and do not regress anything the cutover changed, so I'm merging. Worth a follow-up PR, roughly in priority order:

- **Direct `filletV2`/`chamferV2` atomicity** (`bindings/operations.rs`): unlike the `fillet` binding, the direct v2 bindings return `result.solid` without rejecting partial/failed/unchanged results. Pre-existing, but worth aligning with the atomic contract (with care for consumers that may rely on partial returns).
- **Full-circle spine closure** (`spine.rs`): a full circular/elliptical edge is flagged periodic with zero-length coincident endpoints, so `locate` samples only the seam. Compute arc-length / param-span before marking periodic.
- **Custom radius law flattened** (`fillet_plan.rs`): `RadiusLaw::Custom` is reduced to linear endpoint interpolation, so variable fillets with a custom law use the wrong interior radius. Constant-radius production is unaffected.
- **Full-torus AABB** (`measure/bounding_box.rs`): the two-degenerate-seam torus form is not recognized, collapsing the AABB toward the seam point.
- **Closed-rim torus seam** (`builder_utils.rs`): the periodic seam is built as a straight chord off the torus surface.
- **Registry double-bind** (`boundary_registry.rs`): two logical keys binding the same pre-existing edge overwrites the first mapping in `by_edge`.
- **`signed_volume` magnitude** (`signed_volume.rs`): the grid integrates the full analytic surface and the planar-cap polygon undercounts circular boundaries. This is fine for the current use (its one call site is a sign-only orientation check in `fillet_builder`), but it would need the trimmed region plus circular-cap area if the helper is ever reused for true volume.

The remainder are P2/P3 test-strength notes and a deprecated-path (`rolling_ball`) issue. None block the cutover. Merging now; these are tracked for a follow-up.

PR #1636, opened 2026-09-06
fix: repair isolated adjoining rolling-ball fillets and rational STEP surfaces

ceeray:
## Reduced defect

On upstream v3.4.0 (`f3b0eb113db42dfc997996c7f4fc6523661c6923`), extrude a 40 × 25 mm XY rectangle by 30 mm. With the explicit legacy rolling-ball API, fillet `(0,0,0)–(40,0,0)` at 1 mm, then resolve `(0,0,1)–(0,0,30)` on that result and fillet at 1 mm. The second operation returns an open shell with four free edges: the inherited cylinder is passed through although a new strip terminates at its tangent vertex.

OttoCAD's production adapter deliberately still uses this explicit engine; this PR does **not** reroute brepkit production/WASM dispatch or claim to fix fillet_v2. Its v2 sequential probe still fails at the incidence audit.

## Bounded repair

- Recover an isolated equal-radius convex quarter-cylinder's sharp supports on copied topology, then rebuild the inherited and requested strips with a G1-constrained exact rational setback corner.
- Require perpendicular rectangular planar supports, no cavity shells, and more than twice the radius of three-edge corner clearance. Other inputs retain existing behavior, not new support claims.
- Validate the result strictly and require positive finite volume. Preserve original/accepted source geometry on success and refusal.
- Fix a necessary STEP defect: rational surface weights were omitted by the writer, and complex surface components were not assembled correctly by the reader. Preserve the explicit grid; reject malformed/nonpositive/nonfinite weights rather than silently substituting ones. Rational curves are unchanged.

A positional Coons candidate and a horn-torus candidate were rejected despite closed topology because they failed seam geometry checks. No tolerance, closure policy, other engine, or manifest change is included.

## Evidence

- Native workspace tests/build, all-target strict Clippy, formatting and boundary/doc-path/version checks pass. Existing ignored cases and the explicitly ignored failing v2/comparison probes remain visible.
- New regressions cover the unchanged reduced radii, 36 radius/end/support/rigid-pose combinations, sampled geometric covariance, four tangent joins, independent grouped/sequential edges, insufficient clearance and excessive-radius source preservation.
- The second solid is strictly valid, 9 faces/20 edges (6 planes, 2 cylinders, 1 NURBS). Volume at 0.01 mm deflection: 29984.72052947068 mm³.
[...]

PR #1679, opened 2026-09-15
fix(blend): adjoining, mixed-radius and sharp-mitered fillet corners

ceeray:
Updated 2026-09-16: this branch now carries the **complete** line — 15 commits on upstream `v3.4.14` (`188ae66c`, current `main`), replacing the 10 commits it previously carried on `v3.4.13`. Five commits are new since the last review state, and **the regression I disclosed here is now fixed** rather than merely reported (details below). Every commit is authored under my own name and email; patch content is unchanged from the previously reviewed work, re-landed on the newer base.

Supersedes #1636, which is closed.

This is the fillet-corner work from my CAD app's kernel fork, offered back upstream. The line was developed as a bounded contingency, always with the intent of returning to an equivalently qualified upstream release; it is offered as one reviewable series rather than kept indefinitely out of tree.

## What it fixes

Each defect is an adjoining/corner fillet that upstream builds into an open shell, a wrong corner face, a false refusal, or (in one case) silently returns unfilleted geometry as `Ok`.

| # | Commit | Defect on upstream | Repair |
|---|---|---|---|
| 1 | N339 | Second fillet of an adjoining equal-radius pair returns an open shell with 4 free edges: the inherited cylinder is passed through though a new strip ends at its tangent vertex | Recognize the isolated equal-radius convex corner, rebuild both strips with a G1-constrained exact rational setback patch. Also fixes rational surface weights being dropped by the STEP writer and silently rounded to polynomial by the reader |
| 2 | N340 | Third blend on the retained edge at a smaller radius | Bounded third mixed-radius runout |
| 3-4 | N341 | The *same* two-edge corner with `R1 != R2` leaves `V=13 E=20 F=8`, 4 boundary edges — one face short of the closed result an equal-radius pair produces | Generalize the setback patch to two independent radii (`surface_mixed(r1, r2)`); `surface_mixed(r, r)` reproduces `surface()` to ~1e-14 as the correctness oracle |
| 5-6 | N342 / N342b | Three mutually perpendicular edges at one corner. N342b is the interesting one: **three of the six fillet orders** reproduced the original open-shell defect (`V=16 E=25 F=10`, 5 boundary edges) even after N342, because recognition walked topology storage order — which tracks *which strip was filleted first*, not the corner's geometry — and compared against only one of the two possible role assignments. No existing test exercised those three orders | Order-independent recognition |
| 7-8 | N392 / N392b | Spherical corner radius unvalidated; trihedral setback guard too aggressive | Validation guard; guard correction |
| 9 | N421 | Equal-radius n=2 corner (two adjacent top edges, or all four) builds a torus/NURBS blob, flap or wedge where the correct result is a single sharp miter crease | Rebuilt construction: recognizer, three admissibility rules, exact crease-ellipse geometry, shared-face splice. Terminal ownership moved entirely onto #1654's `terminal_cap_is_notchable` + pre-solver notch rather than racing it |
| 10 | N422 | A trimmed fillet stripe's display/export tessellation can't see its own curvature — raw cylindrical UV mixes radians with model units, so boundary-only CDT joins opposite angular contacts straight across a long axial stripe. Worst measured sag **7.44 mm against a requested 0.25 mm** | Scale the CDT's angular coordinate into model units, then refine interior triangles that exceed the requested chord deflection (32 passes, fail-closed). No tolerance constant changes — the fix makes the triangulation satisfy what the caller already asked for |
| **11** | **N424** | **The regression previously disclosed in this PR** (below) | Gate the spherical-corner radius check on the setback classification. `cross_one_row_fillet_inmem`'s two `full_capture_*` tests pass again — 186/186 edges, valid, closed, watertight |
| **12** | **N430** | The degenerate-face guard refuses valid fillets at the exact geometric limit. Its own doc comment claims it "is keyed on actual face area … so a valid fillet (every face has real area) is never rejected" — that claim is false: at `r = S` a legitimate limit fillet leaves a zero-area planar remnant and is refused with the closed-circular-rim message | Give the guard a real discriminator on the collapsed face's surface kind, so the rim case it was written for still refuses and the limit case builds. No tolerance was loosened |
| **13** | **N431** | The adjacent-edge-length refusal measured the wrong thing and said so unclearly | Corrected message and measurement basis (the input snapshot, not post-mutation topology) |
| **14** | **N432** | Two adjacent edges at the exact limit `r = S` cannot be stitched: the corner must close without the third edge, which vanishes at the limit | Built by construction. Measured volume `10923.770237` against the closed form `(2/3)S³ = 10924.709` (−0.0086 %); 157-point sweep across `(0, S]` with zero refusals and zero `Ok`-but-open results |
| **15** | **N433** | Four edges at the exact limit `r = S/2` (the "pillow") leave an open shell — the four stripes do not meet at the vanished top face | Built by construction. Measured `13655.10` against `(5/6)S³ = 13655.887` (−0.0058 %); 184-point sweep across `(0, S/2]` |

On 14 and 15, the **sweeps** are the point, not the endpoints: an earlier fix had left an `Ok`-but-open band from ~12.446 mm up to the pillow limit, so a two-point boundary test would have passed while the interior stayed broken.

Three findings surfaced during N421 that are worth calling out on their own:

- **`trimmer::rebuild_mapped_parametric_face`'s split guard was dimensionally inconsistent.** It compared the coarse scan's minimum distance (a length) against the carrier's parametric span (1.0 for every line, an angle for arcs). On a 254 mm edge the coarse samples are 3.97 mm apart, so a point exactly on the carrier could not pass the filter: terminal spokes were never split and #1654's notch silently never fired. Now bounded against `max(span, chord)`. `BoundarySplitPolicy::Selective`'s semantics are untouched and #1654's own tests stay green. 

*This finding has been addressed in https://github.com/andymai/brepkit/pull/1680

- **`collect_terminal_boundaries` registered support edges against every stripe's cross arcs**, so a boundary whose owner slot no built patch claims was registered as a runout. Narrowed to the current stripe's own arc.
- **`FilletPlan` drops tangent-continuous selected edges** (#1650), so a *requested* collinear continuation can be absent from `plan.selected_edges`; measuring the radius bound from the plan alone refused a whole valid fixture. The radius rule now also recognizes a requested collinear continuation over the source topology.

## The regression I previously disclosed here is fixed

Earlier revisions of this PR carried this warning:

> **This series introduces one regression I have not fixed.** `crates/io/tests/cross_one_row_fillet_inmem.rs`'s `full_capture_fillet_is_complete_and_closed` and `full_capture_step_round_trip_stays_watertight` fail with `Blend(CornerFailure { vertex: Id(0) })`. Bisected to N392 (commit 7).

**N424 (commit 11) fixes it at the root cause**: N392's spherical-corner radius check was firing on corners it should not have, and is now gated on the setback classification. Both tests pass, the fixture builds 186/186 edges valid/closed/watertight, and the full native workspace went to **zero failures** for the first time since N392 landed.

The offer I made earlier — to drop commit 7 (N392) and re-land the rest without it — is therefore **withdrawn as unnecessary**. The guard it added is worth keeping; it was the gating that was wrong.

## Still disclosed

**One deliberate test-tolerance change.** `crates/wasm`'s `prism_vertical_corner_fillets_are_watertight` compares `solid_volume` (exact analytic) against `oriented_solid_volume` (inscribed mesh at 0.01 mm) with a 1.0 mm³ tolerance. The tessellation repair removes ~207 mm³ of inscribed-mesh undercount and leaves a 1.385 mm³ residue (4e-5 relative) — deflection-level mesh error, not the inverted face the check exists to catch. Tolerance widened 1.0 → 5.0 with that measurement in the comment. Flagging it explicitly since widening an existing assertion deserves a second opinion.

**Not claimed:** no reroute of production/WASM dispatch, no `fillet_v2` rewrite, no tolerance/closure-policy change, no sewing fallback or healing, no acceptance of open output. Unrecognized geometry still falls through to the ordinary builder. A known pre-existing upstream behavior this series does *not* fix: near the degenerate limit, a two-edge `fillet_v2` returns a vacuous `Ok` (`succeeded=0`, `partial=true`, the unfilleted solid) rather than an error — this line refuses it with a typed `RadiusTooLarge` instead, which I measured as strictly better than both baselines, but the vacuous path elsewhere is untouched.

## Qualification

On the tip, base `v3.4.14`:

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `scripts/check-boundaries.sh`, `git diff --check` — all exit 0.
- `cargo test --workspace --no-fail-fast`: **2949 passed / 0 failed / 26 ignored**. The two `cross_one_row` failures this PR previously carried are gone (N424).
- Tessellation regression (`regress_fillet_stripe_tessellation_deflection.rs`, 24 cases: box/extrusion × identity/rigid × single/adjacent/four edges): fails 24/24 before the fix on the actual measurement, passes 24/24 after.
- N413/N414 geometry matrix: 96/96 rows, `crossings=0` (both scales, half radii, order variants, single-edge control, rigid transform). STEP roundtrips 4/4, torus-incidence 2/2, seam-stub strong oracle 6/6, scoop fixtures 4/4 and 20/20 each.
- Upstream coverage held: #1643 grouped scoop, #1654 prism corner pin, #1653 assembly replays, kumiko and lship fixtures all pass.
- The exact-limit constructions are additionally exercised end-to-end through a real application (my CAD app, pinned to this line with no local patch): two adjacent edges at `r = S` and all four at `r = S/2` build every target, and the radii just above refuse cleanly with the source solid preserved.

The pre-N424 commits were re-landed onto `v3.4.14` patch-id-identically — every `git patch-id --stable` value equals the original, zero conflicts, every carried file byte-identical.

## Base

Branched at `v3.4.14` (`188ae66c`), which is current `main`, so this should apply without a rebase. The whole series was qualified on that tree.

## Review guidance

The volume is mostly tests. The geometry to actually review is:

- `crates/blend/src/corner.rs` (the recognizers, the miter junction, the exact-limit corner closure)
- `crates/operations/src/fillet/setback_patch.rs` and `reblend.rs` (the rational setback corner)
- `crates/operations/src/fillet/rolling_ball.rs` (the degenerate-face discriminator and the radius-admissibility guards)
- `crates/blend/src/fillet_builder.rs` (the explicit ordering contract)
- `crates/operations/src/tessellate/nonplanar.rs` (the CDT fix)
- `crates/blend/src/trimmer.rs` (one line — the split-guard bound)
- `crates/io/src/step/{reader,writer}.rs` (rational weights)

Design notes for the first commit are in `docs/n339-adjoining-fillet.md`.

Implemented with AI assistance and verified by the test suites above; the sharp-miter corner and both exact-limit constructions were additionally checked visually against reference geometry before I accepted them. No auto-merge requested — review the qualification envelope and tell me what you want changed, split, or dropped.


andymai (changes requested):
Thanks for offering this back, and for the level of qualification in the description. I read the whole diff rather than sampling it. The core geometry is good in places: I re-derived the miter crease ellipse independently and `miter_crease_edge` is exactly right, `trihedral_ball_centre` is a real de-duplication, and swapping the approximate NURBS corner for a true `Sphere` is the kind of change this kernel wants. But I can't merge this as one 13k-line squash across two engines, a tessellation change and a STEP change, and there are a few correctness gates I need fixed before any of it lands. Details below, ordered by what blocks.

## Blocking

**1. The miter recognizer never checks convexity, and the path it commits to has no fallback.**
`detect_equal_two_edge_miter` (`crates/blend/src/corner.rs:2717`) and `find_equal_two_edge_miter_junctions` (`:417`) document a "convex" trihedral corner but test only: `Junction` classification, two incident stripes, equal radius, three mutually perpendicular planes, one shared face. Nothing in that path, in `check_equal_two_edge_admissibility`, or in `CornerClassification::Junction` (`fillet_plan.rs:331`) distinguishes convex from concave. Two concave floor-to-wall edges of a rectangular pocket at one radius pass every test; `apply_equal_two_edge_miter_corner` then computes `p00 = vertex - n*r` on the wrong side of the material, `adopt_side_face_vertical_edge` finds no edge, and the whole `fillet_v2` selection errors where `main` builds it through `build_junction_fan`. Please gate on convexity (the retained unselected edge must leave the vertex along `-n`) and return `Ok(None)` rather than committing when any precondition fails. Same request for `reblend::try_adjoining_strip` (`crates/operations/src/fillet/reblend.rs:87`), which runs on every single-edge `fillet_rolling_ball` call and turns "recognized but unbuildable" (`:106`, `:118`, `:129`, `:138`) into a hard failure of the fillet.

**2. N392b deletes a guard the production cutover shipped on purpose, with an empty commit body.**
`corner.rs:1068` removed the end-decile check ("corners whose center projects onto a stripe mid-section are short-spine junctions, e.g. rib-top corners: trimming there removes valid material", from #1631). The replacement comment says the caller's trim ranges reject setbacks that consume an entire stripe, but the caller (`:949`) only rejects total consumption, not mid-section. Every convex trihedral junction with `0.1 < r/L < 0.5` now takes the setback path it was excluded from. I need either the guard restored, or a stated justification plus a fixture for the rib-top class it named.

**3. The tessellation refinement over-produces triangles on the hottest path in the kernel, and "fail-closed" is really a silent fallback.**
`crates/operations/src/tessellate/nonplanar.rs:2094` inserts triangle centroids into a CDT whose u axis is scaled to model units, which yields isotropic triangles on a surface that is developable along v. Your own measurement in `assembly_probe_tests.rs:334` is 252 to 8450 triangles for four r=2.5 stripes at 0.01 mm; a strip of two triangles per angular division needs roughly a tenth of that. Every export, preview, render mesh and the wasm `try_fillet` validity gate pays it. Please subdivide only the angular span (pre-seed the u grid to the chord requirement, or insert on the u-long edge). Separately, the non-converged branch (`:2098`) returns an error that every caller in `tessellate/solid.rs` (`:1113`, `:1153`, `:1242`) swallows by falling back to `tessellate_nonplanar_snap`, so a face that needs more than 32 passes pays the full cost and then ships the old sag with no log line. State that contract in the doc comment and log it.

**4. Rule 2 at `corner.rs:835` refuses a whole selection for an edge the construction never consumes.**
The doc concedes it checks every unselected sharp edge at the terminal and is "never more permissive" than checking the one the runout consumes. A prism whose filleted edge ends at a valence-4 vertex with a short unselected notch lip at `r > run` now returns `RadiusTooLarge` for the entire 20-edge selection, and the wasm `try_fillet` then silently degrades to the flat bevel. Restrict it to the consumed edge, or scope the refusal to the offending contour. Also `RadiusTooLarge` has no consumer today, so the "caller can recover `max_radius`" promise isn't realised anywhere.

**5. Scale-dependent absolute tolerances in new code.**
This kernel has to behave identically at 1 mm and 1000 mm. New absolute lengths: `spherical_triangle.rs:115` (`TOL*100` on a radius), `corner.rs:1245` (`TOL*100` as the set-back classification switch), `builder_utils.rs:488` (`1e-7` on a radius test), `fillet_builder.rs:1413` (`1e-7` raw literal), `reblend.rs:1225` (`linear*100` on control points of a radius-scaled template), `reblend.rs:169` (`0.01` mesh deflection inside an accept gate; at 1e-3 scale a valid result gets rejected as `NonManifoldResult`), `nonplanar.rs:1955` (`.max(0.01)` floor on the u scale), and the `< 5.0` mm³ absolute in `assembly_probe_tests.rs:342`. Make them relative to radius or derive them from `Tolerance`. For the probe assertion, since an inscribed mesh can only under-count, `residue >= 0.0 && residue < 1e-4 * volume` keeps its teeth without a hand-tuned constant.

**6. Mesh-derived decisions inside modeling ops.**
`reblend.rs:169` gates accept/reject on a tessellated volume, and N430 turns the pre-existing `face_area` mesh measurement into a face-dropping trigger (`rolling_ball.rs:2517`). This repo does not let tessellation decide B-Rep topology. `validate_solid` immediately above the volume gate is the right check; the vanished-remnant test should be analytic (planar polygon area from the wire).

**7. New `_ =>` wildcards on `EdgeCurve` and `FaceSurface`.**
CLAUDE.md states no production wildcard arms remain on these enums so the compiler flags every site when a variant is added. This adds six: `nonplanar.rs:1955`, `corner.rs:446`, `corner.rs:2756`, `reblend.rs:516`, `:709`, `:1598`. The two in `corner.rs` can just call `Face::effective_plane_normal()`.

**8. Comments cite documents that do not exist here, and carry ticket numbers nobody upstream can resolve.**
21 references to ten `docs/N4xx...` paths that are not in this repo (`docs/N414-evidence/miter-construction-design.md`, `docs/N413-twoedge-construction-diagnosis.md`, `docs/N342-*`, and so on), and about 220 `Nxxx` prefixes in source and tests. Roughly two thirds of those comments are load-bearing once the prefix and the citation are stripped (`rolling_ball.rs:438`, `corner.rs:2788`, `:2926`, `builder_utils.rs:474`, `boundary_registry.rs:395` are good examples: keep the prose, drop the ref). The rest is before/after narration that should go: `reblend.rs:37-48` and `:1551-1566` (including the dead `_requested_radius` parameter), `corner.rs:2869-2881` (cites a fork commit hash), the five `a7ba558d` mentions, and `nonplanar.rs:1899-1925`. `docs/n339-adjoining-fillet.md` is a fork-internal note (it names your app, its private test counts, and a fork base commit) and this repo has no `docs/` directory; fold its two useful paragraphs (the STEP prerequisite and the rejected candidates) into the PR body or a comment on the construction. `regress_fillet_mixed_radius_adjoining.rs:127` still says "this is the RED baseline assertion: it must currently FAIL" above an assertion this PR makes pass.

## Smaller correctness items

- `corner.rs:535` `requested_collinear_continuation` scans the entire arena, which is the bug `incident_face_count` was scoped against 500 lines earlier in this same PR. Pass the same `original_faces` scope.
- `corner.rs:2788` resolves the shared top face from stripe `a` only; one line checking stripe `b` agrees makes the stated invariant enforced instead of assumed.
- `corner.rs:2452` `planned_side` looks unreachable: the only site registering that key is `:2520`, after the test, in a function called once per junction. Dead or a second path I could not find?
- `corner.rs:3136` `fraction = radius / length` assumes the contact NURBS is arc-length parameterised; only the variant is checked.
- `fillet_builder.rs:866` never sets `stripe_faces[si]` on the non-registry branch, and `apply_equal_two_edge_miter_corner` hard-errors on the missing entry. Probably unreachable for the miter class, but nothing enforces it.
- The STEP weights fix is surfaces-only. `write_nurbs_curve` (`writer.rs:343`) still drops curve weights, so a `circle_to_nurbs` 90° arc re-imports with a 6% radial error, and the composite-curve reader path (`reader.rs:470`, `:736`) keeps the lenient uniform-weights fallback you removed for surfaces. Either fix both halves or say explicitly that curves are out of scope. Also validate weights in `write_nurbs_surface`: `fmt_f64` flushes `|w| < 1e-15` to `0.` and the reader rejects `w <= 0`, so brepkit can now emit a file it refuses to read back. `reader.rs:1187` compares weights with `assert_eq!` after a decimal round-trip; use a tolerance.
- `reblend.rs:736` picks the sphere's pole axis to dodge a tessellator longitude-unwrap defect. Fix the mesher rather than letting a display concern steer the parameterisation.
- `is_fillet_stripe_boundary` is a topology fingerprint (2 NURBS + 2 arcs, or 1 NURBS + 2 transverse), so boolean-trimmed cylinders and cones change density too, and every other cylinder trim keeps the radians-vs-mm CDT metric. Worth stating in the doc comment.
- `reblend` and the blend miter give two different answers for the same configuration (equal-radius orthogonal convex trihedral, two edges filleted): an exact two-cylinder miter with an analytic crease in production, and a G1 Hermite-Coons rational patch with a setback in the deprecated engine, whose deviation from the true blend is not bounded anywhere. Nine of the fifteen test files exercise `fillet_rolling_ball`, which has been deprecated since 2.44.0. I would rather the deprecated engine mitre too (which deletes `setback_patch.rs`) than carry an approximation for a shape that is exactly representable.

## Tests

The tessellation pin (`regress_fillet_stripe_tessellation_deflection.rs`) is the best-constructed file here, keep it as is. Otherwise this adds ~6200 test lines (65% of the current `operations/tests` directory) and roughly 600 extra fillet builds at opt-level 0 in CI:

- Move `regress_exact_limit_fillets.rs`, `regress_two_edge_exact_limit.rs`, `regress_four_edge_exact_limit.rs` out of `crates/io/tests/`; none of them touch io.
- Drop `regress_n341_naive_control.rs` (byte-identical fixture to `regress_fillet_mixed_radius_adjoining.rs`; its one unique assertion is two lines there) and the two `#[ignore = "diagnostic engine comparison"]` probes; `#[ignore]` here is reserved for ready-repros with an acceptance target.
- Shrink the sweeps: the 83- and 31-radius ladders to the transition point ±1 ulp, the endpoints and a few interior samples; the 96-case `n414_native_matrix.rs` to one size and one fraction; the two 48-case order sweeps to one order plus one permutation each. Order-independence is one property, not 48 measurements.
- Rename `n414_native_matrix.rs` and `regress_n341_naive_control.rs` after the geometry, like the repo's other regression files. `edge_at` is defined 13 times across the new files; a `tests/common/mod.rs` or fewer files.
- The diff changes exactly one pre-existing assertion (the probe tolerance, loosened 5x) while changing the corner construction for every equal-radius orthogonal two-contour junction, the commonest box fillet in the suite. Name the existing fixtures that now take the miter path and show their face counts and volumes moved intentionally, or explain why none do.

## How I'd like this to land

Split it, in dependency order, each with a conventional-commit title and a real body (N392 and N392b currently have none):

1. The trimmer split guard is already open as #1680 with a regression test. I'm landing that first with a tightened bound (the coarse scan's step length rather than the chord, which is zero on a closed circle). Rebase this line onto it and drop the duplicate hunk.
2. The STEP rational weights and the ELLIPSE placement fix as `fix(io)`; both are good and independent.
3. The tessellation deflection fix as `fix(operations)`, once the refinement is anisotropic and the fallback is logged.
4. The blend miter corner (N421/N432/N433) with the convexity gate and fallbacks from item 1, and the exact-limit sweeps trimmed.
5. The trihedral setback classification (N392/N392b/N424) with the decile-guard justification from item 2.
6. The deprecated rolling-ball engine work (N339 through N342, N430, N431), or a decision to mitre there instead.

I know that's a lot. The work is worth landing, which is why I'd rather review it in pieces I can actually verify than wave through a squash I can't.

andymai:
Rather than send this back for a rework, I split it and landed it myself, with you as co-author on every commit. Where the slices went:

- #1680 (merged): the trimmer split guard, bounded by the scan's own step length.
- #1682 (merged): the trihedral setback classification (N392, N392b, N424). The decile guard's job is already done by the near-equal-stripe guard, which is the justification the removal needed.
- #1683 (merged): STEP rational weights for surfaces and curves, export-time weight validation, and the ELLIPSE placement fix.
- #1685 (merged): the stripe deflection fix (N422), with the refinement made anisotropic so triangle counts match main and the fallback logged.
- #1687: the sharp-mitered corner and both exact-limit constructions (N421, N432, N433). Added a convexity gate and precondition fallbacks in the recognizer, restricted the runout rule to the support-face edges, scoped the collinear-continuation scan to the requested set, removed a dead branch, and trimmed the sweeps and the matrix.

Not landed: N339 through N342b, N430 and N431. Those extend `fillet_rolling_ball`, which is deprecated and explicitly kept only as a reference implementation that production and wasm never call, and they carry a rational approximation of a corner the production engine now builds exactly. If your app needs that route upstream, the path I would take is to have it mitre the same way and delete the setback patch, and I would review that as its own PR.

Thanks for the work and for the qualification detail; it made the split tractable.

andymai:
Closing as superseded: every slice is on main (#1680, #1682, #1683, #1685, #1687), see the summary above.

ceeray:
Thanks Andy, much appreciated! :)

PR #1680, opened 2026-09-16
fix(blend): bound the trimmer's split guard by chord length, not parametric span

ceeray:
## Summary

`rebuild_mapped_parametric_face` decides whether a boundary vertex lies on a candidate edge's carrier with a 65-sample coarse scan across the carrier's parameter domain, then refines with a ternary search. The scan's minimum (`best_d`) is a model-space distance, but the rejection bound it was compared against, `(d1 - d0).abs()`, is the carrier's *parametric* span — 1.0 for every line and an angle for circles/ellipses. The two are not comparable: on a line the bound stays pinned at 1.0 model unit however long the edge is, while the coarse samples spread out as `edge_length / 64`. An on-carrier point can sit up to half a step (`edge_length / 128`) from the nearest sample, so past roughly 128 model units of edge length the filter reads "not this edge" for a point that lies exactly on it. This bounds the guard by the larger of the parametric span and the endpoint chord — a lower bound on the edge's true extent — which keeps every on-carrier point inside the filter at any scale.

`BoundarySplitPolicy::Selective`'s semantics are unchanged: the guard only decides whether a candidate carrier is worth refining, and the exact acceptance test after refinement (`dist_at(tt) > 1e-6`) still rejects points that are not actually on the carrier.

## Consequence

A 254 mm prism with its four vertical corners filleted at r = 25.4 mm — the same relative geometry as a 25.4 mm prism at r = 2.54 — samples the carriers every 3.97 mm, and the terminal contact, which lands r along the cap edge, sits 1.59 mm from the nearest sample against the 1.0 bound. The terminal spokes were therefore never split, so the pre-solver terminal notch added in #1654 silently never fired and each support face kept its doubled-back runout tail: `fillet_v2` reported 4/4 corners succeeded and the solid passed every id-level check, yet it tessellated to 8 planar faces overlaying the caps and 32 open mesh edges. At 25.4 mm the samples are 0.40 mm apart and the same point sits 0.16 mm from its nearest sample, which is why the existing fixtures never caught this.

## Verification

`crates/operations/tests/regress_blend_split_guard_scale.rs` fillets the four vertical corners of a prism at both scales and requires four successful corners, a valid shell and a watertight tessellation (0 boundary mesh edges).

Reviewer reproduction (no fork needed, both directions):

```sh
git checkout -b repro v3.4.14
git checkout fix/trimmer-split-guard-scale -- crates/operations/tests/regress_blend_split_guard_scale.rs
cargo test -p brepkit-operations --test regress_blend_split_guard_scale
# → 254 mm case FAILS: "corner fillets left open mesh edges ... left: 32 right: 0" (exit 101),
#   one-inch control passes
git checkout fix/trimmer-split-guard-scale -- crates/blend/src/trimmer.rs
cargo test -p brepkit-operations --test regress_blend_split_guard_scale
# → 2 passed, 0 failed (exit 0)
```

Full gates on the branch tip: `cargo fmt --all -- --check`,
`cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`
(2869 passed, 0 failed, including #1654's `prism_vertical_corner_fillets_are_watertight`),
`./scripts/check-boundaries.sh`, `./scripts/check-doc-paths.sh` — all exit 0.

*Found downstream against this tag while qualifying fillet-corner work in a fork; the reproduction above is native to this tree and has no fork dependency.*

---
## Summary by cubic
Fixes the trimmer's split guard, which compared a model-space sample distance against a carrier's parametric span. Large scaled fillets could skip valid terminal splits and produce open mesh edges; the guard now bounds the filter by the longest coarse-scan step's model-space length, which also covers closed circular carriers where the endpoint chord is zero.

- Adds regression coverage for similarly shaped 25.4 mm and 254 mm prisms with vertical corner fillets.
- Verifies both cases produce valid, watertight tessellations, including the previously failing 254 mm case.

<sup>Written for commit 0b0cbe7514a35646d04f64c64ce9399b410374e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/brepkit/pull/1680?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thanks, the diagnosis is right and both fixtures reproduce here. I pushed one commit on top (0b0cbe75) that tightens the bound: the endpoint chord is zero on a closed circular carrier, so a full-circle boundary edge on a large solid still rejected on-carrier points. The scan can miss a point by at most one sample step, so the guard now uses the longest step's model-space length, which the scan already computes. Clippy clean, blend + operations + wasm suites green locally. Merging once CI runs on the new sha; please drop the duplicate trimmer hunk from #1679 when you rebase it.

